PR #14, opened 2018-12-06
adding age histogram into header

konradjk:


variant_qc/prepare_data_release.py
```diff
@@ -577,17 +578,17 @@ def make_info_dict(prefix, label_groups=None, bin_edges=None, faf=False, popmax=
         if prefix == 'gnomad':
             age_hist_dict = {
                 f"{prefix}_age_hist_het_bin_freq": {"Number": "A",
-                                                            "Description": "Histogram of ages of heterozygous individuals; bin edges are: {}".format(bin_edges[f'{prefix}_het'])},
+                                                    "Description": f"Histogram of ages of heterozygous individuals; bin edges are: {bin_edges[f'{prefix}_het']}; number of individuals in each bin: {age_hist_data}"},
```
gtiao: This description makes it sound like you're listing the number of heterozygous individuals in each bin (i.e., the content of the annotation), so it's a bit confusing. How about "total number of individuals of any genotype per bin" instead?

PR #17, opened 2019-01-25
Hail syntax fix

konradjk:


konradjk:
Whoops, consanguineous calculations came with this one, but that's ok. @lfrancioli might you have a moment to review?

annotations/generate_frequency_data.py
```diff
@@ -145,6 +146,24 @@ def generate_frequency_data(mt: hl.MatrixTable, calculate_downsampling: bool = F
     return mt.rows(), sample_data
 
 
+def generate_consanguineous_frequency_data(ht: hl.Table, f_ht: hl.Table, mt: hl.MatrixTable) -> hl.Table:
+    mt = mt.annotate_cols(meta=mt.meta.annotate(f=f_ht[mt.col_key].pop_f))
+    projects_to_use = hl.literal({'PROMIS', 'T2D_28K_PROMIS_exomes', 'T2D_28K_Singapore_S.Asian_exomes', 'T2DGENES'})
+    consang_call_stats = hl.agg.filter((mt.meta.f.f_stat >= F_CUTOFF) & mt.adj &
+                                       (mt.meta.pop == 'sas') & projects_to_use.contains(mt.meta.project_description),
+                                       hl.agg.call_stats(mt.GT, mt.alleles))
+    consang_call_stats = hl.bind(lambda cs: cs.annotate(
```
lfrancioli: I generally don't like reusing the same var for different types. Here, I think the  following would work well:
```
consang_call_stats = hl.bind(
lambda cs: cs.annotate(AC=cs.AC[1], AF=cs.AF[1], homozygote_count=cs.homozygote_count[1]),
hl.agg.filter((mt.meta.f.f_stat >= F_CUTOFF) & mt.adj &
	                                       (mt.meta.pop == 'sas') & projects_to_use.contains(mt.meta.project_description),
	                                       hl.agg.call_stats(mt.GT, mt.alleles))
)
```

konradjk: It just keeps consistency with what we did above.

lfrancioli: So above is similarly bad?

PR #20, opened 2019-03-04
adding basic liftover functionality

konradjk:


load_data/import_resources.py
```diff
@@ -25,6 +25,15 @@ def import_methylation(overwrite: bool = False):
     kt = kt.transmute(locus=hl.locus(kt.CHROM, kt.POS))
     kt.key_by('locus').write(methylation_sites_mt_path(), overwrite)
 
+    ht = hl.read_table(methylation_sites_mt_path())
+    ref_37 = hl.get_reference('GRCh37')
+    ref_38 = hl.get_reference('GRCh38')
+    ref_37.add_liftover('gs://hail-common/references/grch37_to_grch38.over.chain.gz', ref_38)
+    ht = ht.annotate(new_locus=hl.liftover(ht.locus, 'GRCh38', include_strand=True),
+                     old_locus=ht.locus)
+    ht = ht.key_by(locus=ht.new_locus.result)
```
ch-kr: This is a naive question, but do you need to add the `.result`?

konradjk: Yes, if you use `include_strand=True` - take a look at the result of `hl.liftover` with True and False - they look different

PR #23, opened 2019-06-11
Aggregate coverage by platform / sex

lfrancioli:


load_data/load_coverage.py
```diff
@@ -2,6 +2,25 @@
...
+    return hl.bind(
+        lambda grp_expr: hl.struct(
+            **{
+                f'over_{x}': hl.int32(grp_expr.get(x, 0)) for x in [1, 5, 10, 15, 20, 25, 30, 50, 100]
+            }
+        ),
+        hl.agg.counter(
+            hl.case()
+                .when(mt.coverage >= 100, 100)
+                .when(mt.coverage >= 50, 100)
```
konradjk: 50? and 30 on the next line

load_data/load_coverage.py
```diff
@@ -2,6 +2,25 @@
 from gnomad_hail import *
 
 
+def get_coverage_expr(mt):
```
konradjk: So I like the spirit behind this, but I think it needs some documentation. Most notably, now over_50 does not include those samples that are over_100, right?

load_data/load_coverage.py
```diff
@@ -83,42 +102,38 @@ def main(args):
...
+        if data_type == 'genomes':
+            meta_ht = meta_ht.annotate(qc_platform=hl.cond(meta_ht.pcr_free, 'pcr_free', 'pcr+'))
+
+        mt = hl.read_matrix_table(coverage_mt_path(data_type))
+        mt = mt.annotate_cols(meta=meta_ht.select('release', 'qc_platform', 'sex')[mt.s])
         mt = mt.filter_cols(mt.meta.release)
-        agg_ds = (mt
-                  .group_cols_by(mt.meta.pop, mt.meta.qc_platform)
+
+        n = hl.literal(mt.aggregate_cols(hl.agg.counter((mt.meta.qc_platform, mt.meta.sex))))
```
konradjk: skip the `.literal` and just do `_localize=False` inside the aggregate

load_data/load_coverage.py
lfrancioli: that's a great point -- I'll edit so it has the same behavior as before!

lfrancioli:
@konradjk , this is ready for a new round of review

load_data/load_coverage.py
```diff
@@ -3,19 +3,26 @@
 
 
 def get_coverage_expr(mt):
+
+    cov_arrays = hl.literal({
```
konradjk: Any reason not to have 100 in here?

lfrancioli: since we stop binning at 100  I don't think it makes too much sense. Would you want something like
```
.when(mt.coverage >= 100, cov_arrays[-1])
```
?

konradjk: Fair enough. I guess it doesn't matter too much

PR #24, opened 2019-08-21
Coverage import and analysis

lfrancioli:
- Optimized coverage loading expressions (much faster!)
- Added exomes/genomes coverage stats computation script

### DIFF
--- a/variant_qc/exomes_genomes_coverage.py
+++ b/variant_qc/exomes_genomes_coverage.py
@@ -1,0 +1,145 @@
+from gnomad_hail import *
+from gnomad_hail.resources.variant_qc import get_ucsc_mappability
+
+def import_cds_from_gtf() -> hl.Table:
+    """
+    Creates a HT with a row for each base / gene that is in a CDS in gencode v19
+    :return: HT
+    :rtype: Table
+    """
+    gtf = hl.experimental.import_gtf(
+        'gs://hail-common/references/gencode/gencode.v19.annotation.gtf.bgz',
+        reference_genome='GRCh37',
+        skip_invalid_contigs=True, min_partitions=200
+    )
+    gtf = gtf.filter((gtf.feature == 'CDS') & (gtf.transcript_type == 'protein_coding') & (gtf.tag == 'basic'))
+    gtf = gtf.annotate(locus=hl.range(gtf.interval.start.position, gtf.interval.end.position).map(lambda x: hl.locus(gtf.interval.start.contig, x, 'GRCh37')))
+    gtf = gtf.key_by().select('gene_id', 'locus', 'gene_name').explode('locus')
+    gtf = gtf.key_by('locus', 'gene_id').distinct()
+    return gtf.checkpoint('gs://gnomad-tmp/gencode_grch37.gene_by_base.ht', overwrite=True)
+
+def compute_per_base_cds_coverage():
+    """
+    Creates a HT with a row for each base and gene that is in a CDS in gencode v19 with the following information:
+    1) gene id
+    2) gnomAD exomes coverage by platform
+    3) gnomAD genomes coverage by PCR status
+    """
+
+    def get_haploid_coverage_struct(mt):
+        return hl.cond(
+            mt.locus.in_autosome_or_par(),
+            hl.struct(
+                mean=0.5*hl.agg.sum(mt.mean  * mt.n)/hl.agg.sum(mt.n),
+                median=0.5*hl.median(hl.agg.collect(mt.median)),
+                **{
+                    f'over_{i/2}': hl.agg.sum(mt[f'over_{float(i)}']) / hl.agg.sum(mt.n)
+                    for i in [1] + list(range(5,31,5)) +  [50, 100]
+                }
+            ),
+            hl.struct(
+                mean=hl.agg.sum(hl.cond(mt.sex == 'male', mt.mean * mt.n, 0.5*mt.mean * mt.n))/hl.agg.sum(mt.n),
+                median=hl.median(hl.agg.collect(hl.cond(mt.sex == 'male', mt.median, 0.5*mt.median))),
+                **{
+                    f'over_{i/2}':
+                    hl.null(hl.tfloat32) if not f'over_{i/2}' in mt.entry else
+                    hl.agg.sum(
+                        hl.cond(
+                            mt.sex == 'male',
+                            mt[f'over_{i/2}'],
+                            mt[f'over_{float(i)}']
+                        )
+                    ) / hl.agg.sum(mt.n)
+                    for i in [1] + list(range(5,31,5)) +  [50, 100]
+                }
+            )
+        )
+
+    def get_cov_ht(data_type: str) -> hl.Table:
+        cov_mt = hl.read_matrix_table(coverage_mt_path(data_type, grouped=True))
+        qc_platforms = cov_mt.aggregate_cols(hl.agg.collect_as_set(cov_mt.qc_platform))
+        return cov_mt.select_rows(
+            **{
+                f'{data_type}_{platform}': hl.agg.filter(
+                    cov_mt.qc_platform == platform,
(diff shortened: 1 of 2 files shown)


variant_qc/exomes_genomes_coverage.py
```diff
@@ -0,0 +1,145 @@
...
+    gtf = hl.experimental.import_gtf(
+        'gs://hail-common/references/gencode/gencode.v19.annotation.gtf.bgz',
+        reference_genome='GRCh37',
+        skip_invalid_contigs=True, min_partitions=200
+    )
+    gtf = gtf.filter((gtf.feature == 'CDS') & (gtf.transcript_type == 'protein_coding') & (gtf.tag == 'basic'))
+    gtf = gtf.annotate(locus=hl.range(gtf.interval.start.position, gtf.interval.end.position).map(lambda x: hl.locus(gtf.interval.start.contig, x, 'GRCh37')))
+    gtf = gtf.key_by().select('gene_id', 'locus', 'gene_name').explode('locus')
+    gtf = gtf.key_by('locus', 'gene_id').distinct()
+    return gtf.checkpoint('gs://gnomad-tmp/gencode_grch37.gene_by_base.ht', overwrite=True)
```
konradjk: consider parameterizing `overwrite`, and doing `_read_if_exists = not overwrite`

variant_qc/exomes_genomes_coverage.py
```diff
@@ -0,0 +1,145 @@
...
+
+    def get_haploid_coverage_struct(mt):
+        return hl.cond(
+            mt.locus.in_autosome_or_par(),
+            hl.struct(
+                mean=0.5*hl.agg.sum(mt.mean  * mt.n)/hl.agg.sum(mt.n),
+                median=0.5*hl.median(hl.agg.collect(mt.median)),
+                **{
+                    f'over_{i/2}': hl.agg.sum(mt[f'over_{float(i)}']) / hl.agg.sum(mt.n)
+                    for i in [1] + list(range(5,31,5)) +  [50, 100]
```
konradjk: `[1] + list(range(5,31,5)) +  [50, 100]` is used a few times - maybe make a global variable, or even in resources?

variant_qc/exomes_genomes_coverage.py
```diff
@@ -0,0 +1,145 @@
...
+    """
+
+    def get_haploid_coverage_struct(mt):
+        return hl.cond(
+            mt.locus.in_autosome_or_par(),
+            hl.struct(
+                mean=0.5*hl.agg.sum(mt.mean  * mt.n)/hl.agg.sum(mt.n),
+                median=0.5*hl.median(hl.agg.collect(mt.median)),
+                **{
+                    f'over_{i/2}': hl.agg.sum(mt[f'over_{float(i)}']) / hl.agg.sum(mt.n)
```
konradjk: Seems mildly dangerous to be using a fractional float in the string (might have been easier, if more awkward, to multiply the haploid ones by 2 rather than divide diploid), but I guess if it works it works

variant_qc/exomes_genomes_coverage.py
```diff
@@ -0,0 +1,145 @@
...
+    1) gene id
+    2) gnomAD exomes coverage by platform
+    3) gnomAD genomes coverage by PCR status
+    """
+
+    def get_haploid_coverage_struct(mt):
+        return hl.cond(
+            mt.locus.in_autosome_or_par(),
+            hl.struct(
+                mean=0.5*hl.agg.sum(mt.mean  * mt.n)/hl.agg.sum(mt.n),
```
konradjk: Spacing around `*` here and elsewhere

variant_qc/exomes_genomes_coverage.py
```diff
@@ -0,0 +1,145 @@
...
+    exomes_cov = get_cov_ht('exomes')
+    ucsc_mappability = get_ucsc_mappability()
+    gtf = import_cds_from_gtf()
+
+    gtf = gtf.annotate(
+        **exomes_cov[gtf.locus],
+        **genomes_cov[gtf.locus],
+        mappability=ucsc_mappability[gtf.locus].duke_35_map
+    )
+    gtf.write('gs://gnomad-lfran/exomes_genomes_coverage/gencode_grch37.gene_by_base.cov.ht', overwrite=True)
```
konradjk: move to gnomad-resources maybe?

konradjk: also parameterize overwrite here

variant_qc/exomes_genomes_coverage.py
```diff
@@ -0,0 +1,145 @@
...
+    cov = cov.explode('cov_stats')
+    cov = cov.select(
+        'gene_name',
+        'gene_interval',
+        'mean_mappability',
+        'median_mappability',
+        **cov.cov_stats
+
+    )
+    cov = cov.annotate_globals(
```
konradjk: These aren't going anywhere. A write might not be a bad idea here anyway though

variant_qc/exomes_genomes_coverage.py
```diff
@@ -0,0 +1,145 @@
...
+    2) gnomAD exomes coverage by platform
+    3) gnomAD genomes coverage by PCR status
+    """
+
+    def get_haploid_coverage_struct(mt):
+        return hl.cond(
+            mt.locus.in_autosome_or_par(),
+            hl.struct(
+                mean=0.5*hl.agg.sum(mt.mean  * mt.n)/hl.agg.sum(mt.n),
+                median=0.5*hl.median(hl.agg.collect(mt.median)),
```
konradjk: You could just use approximate quantile here too since you do in the creation, but doesn't really matter

variant_qc/exomes_genomes_coverage.py
lfrancioli: Yeah, the problem here is that we didn't compute the right numbers from the BAMs.

variant_qc/exomes_genomes_coverage.py
lfrancioli: This seems aggregated enough that it could go in 
`gs://gnomad-public/papers/2019-flagship-lof/v1.1/summary_gene_coverage/`
along with the tsv?

konradjk: This is just per-base, per-platform/sex coverage right (which you later aggregate over gene to get the per-gene file that is publicly released)? If so, then yes.

variant_qc/exomes_genomes_coverage.py
konradjk: Yeah agreed, I just mean the practical purpose of a float in a string (in case you get something like `over_10.00000` but I think python doesn't do this, so it's fine)

variant_qc/exomes_genomes_coverage.py
lfrancioli: Agreed. Since I'm not going to re-run it, I'll leave it as it is so it matches the results.

variant_qc/exomes_genomes_coverage.py
```diff
@@ -16,41 +20,44 @@ def import_cds_from_gtf() -> hl.Table:
     gtf = gtf.annotate(locus=hl.range(gtf.interval.start.position, gtf.interval.end.position).map(lambda x: hl.locus(gtf.interval.start.contig, x, 'GRCh37')))
     gtf = gtf.key_by().select('gene_id', 'locus', 'gene_name').explode('locus')
     gtf = gtf.key_by('locus', 'gene_id').distinct()
-    return gtf.checkpoint('gs://gnomad-tmp/gencode_grch37.gene_by_base.ht', overwrite=True)
+    return gtf.checkpoint('gs://gnomad-tmp/gencode_grch37.gene_by_base.ht', _read_if_exists=not overwrite)
```
konradjk: also keep `overwrite=overwrite`

konradjk:
looks great! just one minor thing

PR #26, opened 2019-08-23
gnomad SV import

lfrancioli:


load_data/import_gnomad_sv.py
```diff
@@ -0,0 +1,184 @@
...
+    cols = cols.join(mt2.add_col_index(name='idx2').cols().select('idx2'))
+    return (
+        mt1.choose_cols(cols.idx1.collect()),
+        mt2.choose_cols(cols.idx2.collect())
+    )
+
+
+def merge_with_short_variants(overwrite: bool, min_af: float):
+    # Load gnomAD genomes and filter to PASS / release sites and samples
+    gnomad_genomes = get_gnomad_data('genomes', adj=True, release_samples=True)
```
konradjk: We also have `release_annotations=True` so that you can skip the join in the next few lines

PR #33, opened 2020-02-07
Moved v3 resources currently used to new resources framework

lfrancioli:
*  Note that since family inference and variant QC haven't been moved yet resources for these modules may need additional work


lfrancioli:
@mike-w-wilson I think this is down your alley given all the resource work you've done lately. It's mostly just transferring v3 resources to the new framework. I also removed the `split` arg from `get_gnomad_v3`. I debated whether to trigger `hl.experimental.sparse_split_multi` when using `split=True` but ultimately decided that because of interactions between monomorphic sites, densify and split it may be best to let the user do it afterwards to avoid unintended computing.

gnomad_qc/v3/load_data/load_v2_qc_sites.py
```diff
@@ -1,30 +1,25 @@
...
+    hl.init(log='/load_resources.hail.log', default_reference='GRCh38')
+    hl.get_reference('GRCh37').add_liftover('gs://hail-common/references/grch37_to_grch38.over.chain.gz', hl.get_reference('GRCh38'))
+
     v2_qc_sites = hl.read_matrix_table(qc_mt_path('joint', ld_pruned=True)).rows().select()
     v2_qc_sites = v2_qc_sites.key_by()
     v2_qc_sites = v2_qc_sites.transmute(
         locus=hl.liftover(v2_qc_sites.locus, 'GRCh38'),
         locus37=v2_qc_sites.locus
     )
     v2_qc_sites = v2_qc_sites.filter(v2_qc_sites.locus.contig == 'chr' + v2_qc_sites.locus37.contig).key_by('locus', 'alleles')
```
mike-w-wilson: For my knowledge, why do you only filter to where the contig matches and not the position? Isn't it possible the position does not lift over?

gnomad_qc/v3/load_data/split_multi.py
```diff
@@ -1,23 +1,29 @@
...
 
 
 if __name__ == '__main__':
     parser = argparse.ArgumentParser()
+    parser.add_argument('--create_split_for_real', help='Needed to create the split MT. IMPORTANT: The cost of storing the split MT is likely larger than recomputing it!', action='store_true')
     parser.add_argument('--overwrite', help='Overwrites existing files', action='store_true')
-    main(parser.parse_args())
+    args = parser.parse_args()
+    if args.create_split_for_real:
+        main(args)
```
mike-w-wilson: Since the script is deprecated should this arg `create_split_for_real` even be an option? I think it is a bit misleading since it will not actually create the split for real and just throw another message.

gnomad_qc/v3/resources.py
```diff
@@ -1,131 +1,153 @@
...
 
-v3_qc_mt_path = 'gs://gnomad/sample_qc/mt/genomes_v3/gnomad_v3_qc_mt_v2_sites_dense.mt'
-v3_pc_relate_pca_scores_ht_path = 'gs://gnomad/sample_qc/mt/genomes_v3/gnomad_v3_qc_mt_v2_sites_pc_scores.ht' # TODO -> move to /ht
-v3_relatedness_ht_path = 'gs://gnomad/sample_qc/mt/genomes_v3/gnomad_v3_qc_mt_v2_sites_relatedness.ht' # TODO -> move to /ht
-v3_sex_ht_path = 'gs://gnomad/sample_qc/ht/genomes_v3/gnomad_v3_sex.ht'
+# MatrixTableResources
+v3_qc = MatrixTableResource('gs://gnomad/sample_qc/mt/genomes_v3/gnomad_v3_qc_mt_v2_sites_dense.mt')
+
+# TableResources
+gnomad_v3_genotypes = MatrixTableResource("gs://gnomad/raw/hail-0.2/mt/genomes_v3/gnomad_genomes_v3.repartitioned.mt")
```
mike-w-wilson: Not a big deal but this should probably move up to the MatrixTableResources section if you are going to have the comment separating them

gnomad_qc/v3/resources.py
```diff
@@ -1,131 +1,153 @@
...
+    'final', # TODO: Make sure "final" is the best label once the family scripts are in
+    {
+        'raw': PedigreeResource('gs://gnomad/metadata/genomes_v3/gnomad_v3_trios_raw.fam'),
+        'final': PedigreeResource('gs://gnomad/metadata/genomes_v3/gnomad_v3_trios.fam')
+    }
+)
+
+ped_mendel_errors = TableResource('gs://gnomad/sample_qc/temp/genomes_v3/gnomad_v3_ped_chr20_mendel_errors.ht')
+qc_ac = TableResource('gs://gnomad/annotations/hail-0.2/ht/genomes_v3/gnomad_genomes_qc_ac.ht')
+fam_stats = TableResource('gs://gnomad/annotations/hail-0.2/ht/genomes_v3/gnomad_genomes_qc_fam_stats.ht')
```
mike-w-wilson: How do you feel about these going into a similar structure as`gnomad_annotations.py` in v2? https://github.com/macarthur-lab/gnomad_hail/blob/master/gnomad_hail/resources/grch37/gnomad_annotations.py

gnomad_qc/v3/resources.py
```diff
@@ -1,131 +1,153 @@
...
+            mt = mt.filter_cols(mt.meta.release)
+
+    elif release_only:
+        mt = mt.filter_cols(meta.ht()[mt.col_key].release)
 
     return mt
 
 
-def sample_qc_path(strat: str = ""):
+def get_sample_qc(strat: str = "") -> TableResource:
```
mike-w-wilson: what is strat here?

gnomad_qc/v3/annotations/generate_qc_annotations.py
```diff
@@ -85,7 +94,7 @@ def split_info() -> hl.Table:
     :return: Info table with split multi-allelics
     :rtype: Table
     """
-    info_ht = hl.read_table(get_info_ht_path(split=False))
+    info_ht = get_info(split=False).ht()
 
     # Create split version
     info_ht = hl.split_multi(info_ht)
```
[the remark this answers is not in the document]
lfrancioli: This creates the split version, so it can't yet read it

gnomad_qc/v3/load_data/load_v2_qc_sites.py
lfrancioli: Well, the position will change between the builds. I'm removing the few loci where the contig changes as I don't trust these quite as much -- but to be entirely fair this is more of a gut feeling than a strict evaluation. What I can say is that those that move contig don't lift-under well afterwards so it made me suspicious; it's also a very small number of loci.

gnomad_qc/v3/load_data/split_multi.py
lfrancioli: yeah, that's fair. I think I'll uncomment and let the script runnable

gnomad_qc/v3/resources.py
lfrancioli: Good point, let me add comments on these public functions

gnomad_qc/v3/resources/release.py
```diff
@@ -0,0 +1,28 @@
...
+    '''
+    Fetch filepath for release (variant-only) Hail Tables
+    :param str data_type: 'exomes' or 'genomes'
+    :param str release_tag: String describing release version for use in output filepaths
+    :param bool nested: If True, fetch Table in which variant annotations (e.g., freq, popmax, faf, and age histograms)
+        are in array format ("nested"); if False, fetch Table in which nested variant annotations are unfurled
+    :param bool with_subsets: If True, fetch Table in which all gnomAD subsets are present; if False, fetch Table containing
+        only gnomAD annotations
+    :param bool temp: If True, fetch Table in which nested variant annotations are unfurled but listed under 'info' rather
+        than at the top level; used for sanity-checking sites
```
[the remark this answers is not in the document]
lfrancioli: oof, these docstring is totally out of date!

gnomad_qc/v3/resources/release.py
```diff
@@ -0,0 +1,28 @@
...
+    :return: Filepath for desired Hail Table
+    :rtype: str
+    '''
+    release = release_tag.lstrip('r')
```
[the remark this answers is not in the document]
lfrancioli: we had it for v2.0 and used it in some places. I do agree that looking at this it makes no sense though!

PR #34, opened 2020-02-07
Added requirements.txt

lfrancioli:


lfrancioli:
a few more things than I thought at first... not sure how pylint works exactly, but might be easier to wait until gnomad_hail is in PyPI

requirements.txt
```diff
@@ -0,0 +1,7 @@
+gnomad_hail
```
nawatts:
```suggestion
git+https://github.com/macarthur-lab/gnomad_hail.git@master
```

Since gnomad_hail is not published yet, we'll need to use the git format for it.

https://pip.pypa.io/en/stable/reference/pip_install/#git

nawatts:
I think we also want to leave out anything that Hail depends on (bokeh, numpy, pandas, and pyspark) to avoid version conflicts with it. I tried pip installing this and got errors:
```
ERROR: hail 0.2.32 has requirement bokeh<1.3,>1.1, but you'll have bokeh 1.4.0 which is incompatible.
ERROR: hail 0.2.32 has requirement pandas<0.26,>0.24, but you'll have pandas 1.0.1 which is incompatible.
ERROR: hail 0.2.32 has requirement pyspark<2.4.2,>=2.4, but you'll have pyspark 2.4.5 which is incompatible.
```

nawatts:
Finally got it to install with no version warnings by specifying the version of SciPy in requirements.txt

PR #39, opened 2020-02-24
Added script to create variant annotations histograms

lfrancioli:


gnomad_qc/v3/create_release/make_var_annot_hists.py
```diff
@@ -0,0 +1,86 @@
...
+
+    ht = hl.read_table(release_ht_path())
+    # NOTE: histogram aggregations are done on the entire callset (not just PASS variants), on raw data
+
+    # NOTE: run the following code in a first pass to determine bounds for metrics
+    # Evaluate minimum and maximum values for each metric of interest
+    if args.first_pass:
+        minmax_dict = {}
+        for metric in define_hist_ranges(ht).keys():
+            minmax_dict[metric] = hl.struct(min=hl.agg.min(ht[metric]), max=hl.cond(hl.agg.max(ht[metric])<1e10, hl.agg.max(ht[metric]), 1e10))
```
ch-kr: `hl.cond` -> `hl.if_else`

gnomad_qc/v3/create_release/make_var_annot_hists.py
```diff
@@ -0,0 +1,86 @@
...
+        minmax = ht.aggregate(hl.struct(**minmax_dict))
+        print(minmax)
```
ch-kr: do you think it's worth adding a `logger` and using that instead of print? it doesn't look like anything else gets printed to stdout, so just wondering

lfrancioli: I think in this specific context, because the goal of the function is to print these values `print` makes sense

lfrancioli:
Thanks for your review @ch-kr ! I took another look at this and updated it to use gnomad_hail.utils.annotations. 
I also took another look at how the code was written, and it was just terrible (this was taken straight out of v2): It made two passes through the data and was going back and forth with python for no reason. Now optimized, clean and beautiful ;)

gnomad_qc/v3/create_release/make_var_annot_hists.py
```diff
@@ -0,0 +1,65 @@
...
+    # NOTE: histogram aggregations are done on the entire callset (not just PASS variants), on raw data
+
+    hist_dict = ANNOTATIONS_HISTS
+    hist_dict['MQ'] = (20, 60, 40) # Boundaries changed for v3, but could be a good idea to settle on a standard
+    hist_ranges_expr = get_annotations_hists(
+        ht,
+        ANNOTATIONS_HISTS
+    )
+
+    # NOTE: run the following code in a first pass to determine bounds for metrics
```
ch-kr: it might be helpful to mention this part of the code doesn't actually need to be run each time

PR #42, opened 2020-02-26
V3 coverage

lfrancioli:


gnomad_qc/v3/load_data/compute_coverage.py
```diff
@@ -0,0 +1,49 @@
...
+        )
+
+        coverage_ht = coverage_ht.checkpoint('gs://gnomad-tmp/gnomad.genomes_v3.coverage.summary.ht', overwrite=True)
+        coverage_ht = coverage_ht.naive_coalesce(5000)
+
+        coverage_ht.write(coverage('genomes').versions['3.0'].path, overwrite=args.overwrite)
+
+    if args.export_coverage:
+        ht = coverage('genomes').versions['3.0'].ht()
+        if 'count_array' in ht.row_value:
```
ch-kr: `compute_coverage_stats` doesn't add `count_array` as an annotation, does it?

lfrancioli: Not any more. I've left it here because for v3 it did but then I realized that the `count_array` was taking so much space and wasn't immediately useful so decided to remove it from default computation

lfrancioli: I'm adding a comment regarding this so it's clear

PR #45, opened 2020-03-05
Added v2/v3 PC relate

lfrancioli:


gnomad_qc/v3/sample_qc/v2_pc_relate.py
```diff
@@ -0,0 +1,58 @@
+from gnomad_qc.v2.resources import qc_mt_path, get_gnomad_meta
+from gnomad_qc.v3.resources import gnomad_v2_qc_sites, v3_qc, meta, v2_v3_pc_relate_pca_scores, v2_v3_relatedness
```
ch-kr: are these imports still accurate? should this be `from gnomad_qc.v3.resources.sample_qc import ...`?

gnomad_qc/v3/sample_qc/v2_pc_relate.py
```diff
@@ -0,0 +1,58 @@
...
+import hail as hl
+import argparse
+import logging
+
+logger = logging.getLogger("v2_pc_relate")
+
+
+def liftover_v2_qc_mt() -> hl.MatrixTable:
```
ch-kr: is this the only code that requires the lifted over v2 qc mt? just wondering why this function is in this code and not in a more general location

gnomad_qc/v3/sample_qc/v2_pc_relate.py
```diff
@@ -0,0 +1,58 @@
...
+        v3_qc_mt = v3_qc.mt()
+        v3_qc_mt = v3_qc_mt.filter_cols(meta.ht()[v3_qc_mt.col_key].release)
+        v3_qc_mt = v3_qc_mt.select_rows().select_cols()
+        v3_qc_mt = v3_qc_mt.key_cols_by(s=v3_qc_mt.s, data_type="v3_genomes")
+        joint_qc_mt = v2_qc_mt_liftover.union_cols(v3_qc_mt)
+        joint_qc_mt.write("gs://gnomad-tmp/v2_exomes_v3_joint_qc.mt", overwrite=args.overwrite)
+
+    if args.run_pc_relate:
+        logger.info('Running PC-Relate')
+        logger.warn("PC-relate requires SSDs and doesn't work with preemptible workers!")
```
ch-kr: I think `logger.warn` is deprecated, maybe change this to `logger.warning`?

gnomad_qc/v3/sample_qc/v2_pc_relate.py
```diff
@@ -0,0 +1,58 @@
...
+        joint_qc_mt = hl.read_matrix_table("gs://gnomad-tmp/v2_exomes_v3_joint_qc.mt")
+        joint_qc_mt = joint_qc_mt.sample_rows(0.1)
```
ch-kr: why do you need to downsample the rows?

gnomad_qc/v3/sample_qc/v2_pc_relate.py
```diff
@@ -0,0 +1,58 @@
...
+        eig, scores, _ = hl.hwe_normalized_pca(joint_qc_mt.GT, k=10, compute_loadings=False)
+        scores = scores.checkpoint(v2_v3_pc_relate_pca_scores.path, overwrite=args.overwrite, _read_if_exists=not args.overwrite)
+        relatedness_ht = hl.pc_relate(joint_qc_mt.GT, min_individual_maf=0.01, scores_expr=scores[joint_qc_mt.col_key].scores,
```
ch-kr: since you're running `pc_relate`, would it be helpful to initalize hail with something like `tmp_dir="hdfs:///pc_relate.tmp/"`?

gnomad_qc/v3/sample_qc/v2_pc_relate.py
lfrancioli: These come  from different submodules, so yes I  think they're good as-is.

gnomad_qc/v3/sample_qc/v2_pc_relate.py
lfrancioli: hmmmm, that's a good point! How about moving this to `gnomad_qc.v2.resources.sample_qc.py` ?

gnomad_qc/v3/sample_qc/v2_pc_relate.py
lfrancioli: it's not really a "need", just that ~10k sites is plenty to estimate relatedness and is computationally cheaper..

gnomad_qc/v3/sample_qc/v2_pc_relate.py
lfrancioli: I'm not sure, what's the rational for this?
fwiw, it ran like this.

gnomad_qc/v3/sample_qc/v2_pc_relate.py
```diff
@@ -0,0 +1,58 @@
...
+                                      block_size=4096, min_kinship=0.1, statistics='all')
```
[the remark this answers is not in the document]
lfrancioli: turns out that 0.05 gives an insane number of pairs and we never use them

gnomad_qc/v3/sample_qc/v2_pc_relate.py
```diff
@@ -0,0 +1,58 @@
...
+import hail as hl
+import argparse
+import logging
+
+logger = logging.getLogger("v2_pc_relate")
+
+
+def liftover_v2_qc_mt() -> hl.MatrixTable:
+    b37_v2_qc_sites = gnomad_v2_qc_sites.ht().key_by('locus37', 'alleles')
+    v2_qc_mt = hl.read_matrix_table(qc_mt_path("exomes"))
```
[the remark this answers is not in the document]
lfrancioli: no, that was an overlook due to a redundant `qc_mt_path` found in `gnomad_qc.v2.resources.basic.py` (vs the correct one in `gnomad_qc.v2.resources.sample_qc.py`)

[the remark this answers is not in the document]
lfrancioli: I'll remove the path from basic

lfrancioli:
All right, @ch-kr @jkgoodrich -- I've tried and streamline the v2 liftover as it seems like this was the major issue here. Back to you!

gnomad_qc/v3/sample_qc/v2_pc_relate.py
ch-kr: we've needed to use SSDs to run `pc_relate` on ukb, and my understanding was the cluster doesn't use the SSDs unless you set up the `tmp_dir`. guess this isn't an issue if it ran?

gnomad_qc/v2/resources/basics.py
```diff
@@ -507,10 +515,6 @@ def vep_config_path(ref: str = 'GRCh37'):
 
 
 # Sample QC files
-def qc_mt_path(data_type: str):
-    return 'gs://gnomad/sample_qc/mt/gnomad.{}.high_callrate_common_biallelic_snps.mt'.format(data_type)
-
-
 def qc_ht_path(data_type: str):
```
ch-kr: why does this path exist in both `basics` and `sample_qc.py`?

gnomad_qc/v2/resources/sample_qc.py
```diff
@@ -1,17 +1,61 @@
...
+def get_liftover_v2_qc_mt(
+        data_type: str,
+        ld_pruned: bool,
+        release_only: bool = False,
+        overwrite: bool = False
+) -> hl.MatrixTable:
     """
-    Returns MatrixTable for sample QC purposes: can be exomes, genomes, or joint (joint dataset can also be ld_pruned=True)
+    Returns MatrixTable for sample QC purposes on build 38: can be exomes, genomes, or joint (joint dataset can also be ld_pruned=True)
+    Criteria: callrate > 0.99, AF > 0.001, SNPs only, bi-allelics only
```
ch-kr: slightly more concise: `callrate > 0.99, AF > 0.001, bi-allelic SNPs only`

gnomad_qc/v2/resources/sample_qc.py
```diff
@@ -1,17 +1,61 @@
...
+        grch38_qc_mt = hl.read_matrix_table(qc_mt_path(data_type, ld_pruned=ld_pruned))
+        get_liftover_genome(grch38_qc_mt)
+        grch38_qc_mt = grch38_qc_mt.key_rows_by()
+        grch38_qc_mt = grch38_qc_mt.transmute_rows(
+            locus=hl.liftover(grch38_qc_mt.locus, 'GRCh38'),
+            locus37=grch38_qc_mt.locus
+        )
+        grch38_qc_mt = grch38_qc_mt.filter_rows(
+            grch38_qc_mt.locus.contig == 'chr' + grch38_qc_mt.locus37.contig
+        ).key_by('locus', 'alleles')
```
ch-kr: should this `key_by` be removed?

gnomad_qc/v2/resources/sample_qc.py
```diff
@@ -1,17 +1,61 @@
...
+        grch38_qc_mt = grch38_qc_mt.key_rows_by(locus=grch38_qc_mt.locus38, alleles=grch38_qc_mt.alleles)
+        grch38_qc_mt = grch38_qc_mt.checkpoint(path, overwrite=overwrite)
+
+    if release_only:
+        meta = get_gnomad_meta(data_type)
+        grch38_qc_mt = grch38_qc_mt.filter_cols(meta[grch38_qc_mt.col_key].release)
+        grch38_qc_mt = grch38_qc_mt.select_cols()
```
ch-kr: why do you need this `select_cols()`? are there column fields you need to drop?

gnomad_qc/v2/resources/sample_qc.py
```diff
@@ -1,17 +1,61 @@
...
+    Returns MatrixTable path for sample QC purposes: can be exomes, genomes, or joint (joint dataset can also be ld_pruned=True)
     Criteria: callrate > 0.99, AF > 0.001, SNPs only, bi-allelics only
     """
     if data_type not in ('exomes', 'genomes', 'joint'):
         raise DataException("Select data_type as one of 'genomes' or 'exomes' or 'joint'")
     if ld_pruned and data_type != 'joint':
         raise DataException("ld_pruned = True is only available for 'joint'")
+    ref_str = ''
+    if reference_genome == 'GRCh38':
+        ref_str='.grch38'
```
ch-kr: does this mean only the b38 version has a reference string appended? would it make sense to label both mts?

gnomad_qc/v3/sample_qc/v2_pc_relate.py
ch-kr: ah I didn't realize they were imported in the __init__. thanks!

gnomad_qc/v2/resources/basics.py
lfrancioli: not sure... the history of these files are quite complex. Clearly it makes no sense though.

gnomad_qc/v2/resources/sample_qc.py
lfrancioli: no, you're right -- this was a merging oversight!

gnomad_qc/v2/resources/sample_qc.py
lfrancioli: I don't think so because:
1) All of v2 is on b37 -- so it makes sense that this is the implicit default
2) The files already exist with the current name and might be referenced by others

gnomad_qc/v3/sample_qc/v2_pc_relate.py
lfrancioli: So I've double-checked with Tim and the default file scheme for hail is HDFS, so the default tmp_dir=/tmp really means hdfs:///tmp. So this is not needed.

ch-kr: oh, cool! thanks for sharing

gnomad_qc/v2/resources/sample_qc.py
ch-kr: good point. works for me!

PR #47, opened 2020-03-06
Fix issues found by Pylint

lfrancioli:


requirements.txt
```diff
@@ -1,4 +1,4 @@
-gnomad==0.1.0
+git+https://github.com/broadinstitute/gnomad_methods@master
```
nawatts: Since we've switched from git to PyPI and back, it would probably be worth adding a comment here on why we're using master from GitHub instead of a PyPI release.

PR #48, opened 2020-03-12
Initial subset script for v3

mike-w-wilson:
This is an initial PR for the gnomAD subsetting script. @ch-kr told me she is working on a more general subset script that will be committed to the general repo using similar logic. Once that is in, this script will be updated. However, since the subsets requests are coming in I wanted to make sure this was reviewed before handing over more VCFs to collaborators.


gnomad_qc/v3/subset/subset.py
```diff
@@ -0,0 +1,127 @@
...
+    meta = hl.read_table(metadata)
+
+    if pop and pop in GENOME_POPS:
+        logger.info("Subsetting samples to {pop} population")
+        mt = mt.annotate_cols(pop=meta[mt.col_key].pop)
+        mt = mt.filter_cols(mt.pop == pop)
+        mt = mt.cols().drop("pop")
+
+    if subset:
+        mt = subset_samples_and_variants(mt, subset, sparse=True)
```
lfrancioli: add `gt_expr = "LGT"` ?

gnomad_qc/v3/subset/subset.py
```diff
@@ -0,0 +1,127 @@
...
+    mt = hl.experimental.sparse_split_multi(mt, filter_changed_loci=True)
+    mt = hl.experimental.densify(mt)
+
+    mt = mt.filter_rows(hl.agg.any(mt.GT.is_non_ref()))
+    mt = mt.filter_rows(hl.len(mt.alleles) > 1)
+
+    mt = mt.drop(mt.gvcf_info)
+    mt = mt.annotate_rows(info=info_ht[mt.row_key].info)
+
+    mt = mt.naive_coalesce(1000)
```
lfrancioli: I wonder if this should be a param or a function of the resulting size of the data?

gnomad_qc/v3/subset/subset.py
```diff
@@ -0,0 +1,127 @@
...
+            info_ht = info_ht.annotate(
+                info=info_ht.info.annotate(
+                    **{
+                        f: info_ht.info[f].map(
+                            lambda x: hl.float32(hl.min(2 ** 31 - 1, x))
+                        )
+                    }
+                )
+            )
+        elif ft == hl.dtype("array<array<int32>>"):
```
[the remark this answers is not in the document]
mike-w-wilson: Yes, thanks for catching this. This was for the SB field and I realize you handle that separately in the `ht_to_vcf_mt`.  I've add that in.

gnomad_qc/v3/subset/subset.py
mike-w-wilson: Right!

gnomad_qc/v3/subset/subset.py
mike-w-wilson: I gave this a shot using the recommended 128MB per partition but there isnt currently a built in way to get the actual size of the MT so I also used an entry estimate I calculated from the size of vcf shard.

mike-w-wilson:
I've added in the vcf header. I used the info and format descriptions from v2 that overlapped and also what I found here: https://github.com/broadinstitute/gatk/blob/master/src/test/resources/org/broadinstitute/hellbender/tools/walkers/GnarlyGenotyper/twoSampleASDB.vcf 
I took a shot at making the number of partitions for naive coalesce a function. I set a minimum to 20 because anything lower than that was taking a very long time. Back to you @lfrancioli !

gnomad_qc/v3/subset/subset.py
```diff
@@ -0,0 +1,243 @@
...
+    mt_disk_est = rows * columns * entry_size
+    n_partitions = (
+        int(mt_disk_est / partition_size) if (mt_disk_est / partition_size) > 20 else 20
+    )
+    return n_partitions
+
+
+def main(args):
+    hl.init(
+        log="/subset.log", tmp_dir="hdfs:///subset.tmp/", default_reference="GRCh38"
```
lfrancioli: I double-checked this with Tim: `/tmp` (the hail default) is the same as `hdfs:///tmp`. Happy to keep the `tmp_dir` arg here if you want but just thought I'd clarified this as I think that there is some confusion about this (and the fact that using the default uses HDFS / SSDs if present)

gnomad_qc/v3/subset/subset.py
```diff
@@ -0,0 +1,243 @@
...
+    ):
+        info_ht = info_ht.annotate(
+            info=info_ht.info.annotate(SB=hl.flatten(info_ht.info.SB))
+        )
+
+    meta = metadata.ht()
+
+    if pop and pop in GENOME_POPS:
+        logger.info(f"Subsetting samples to {pop} population")
+        mt = mt.annotate_cols(pop=meta[mt.col_key].pop)
```
lfrancioli: These 3 lines could be a single one:
```
mt = mt.filter_cols(meta[mt.col_key].pop == pop)
```

gnomad_qc/v3/subset/subset.py
```diff
@@ -0,0 +1,243 @@
...
+        mt = subset_samples_and_variants(mt, subset, sparse=True, gt_expr="LGT")
+
+    if args.pull_meta:
+        meta = meta.semi_join(mt.cols())
+        meta.export(f"{args.output_path}/metadata.tsv.bgz")
+
+    mt = hl.experimental.sparse_split_multi(mt, filter_changed_loci=True)
+    mt = hl.experimental.densify(mt)
+
+    mt = mt.filter_rows(hl.agg.any(mt.GT.is_non_ref()))
```
lfrancioli: These two lines should be switched because `mt = mt.filter_rows(hl.len(mt.alleles) > 1)` is less costly than `mt = mt.filter_rows(hl.agg.any(mt.GT.is_non_ref()))` since it doesn't operate on the entries. In most programming langs, when evaluating logical statements with multiple conditions (which these will be collapsed into (`mt = mt.filter_rows(hl.agg.any(mt.GT.is_non_ref()) & (hl.len(mt.alleles) > 1))`), the cheaper (or most common, depending on the situation) should always come first as oftentimes the last part of the statement is only evaluated as needed (in this case, `hl.agg.any(mt.GT.is_non_ref())` would only run iff `hl.len(mt.alleles) > 1)`. I asked Tim on Zulip to be sure that hail is sensitive to ordering and he said we should switch (without providing much details though). In any case, I think it's a good habit to take to think about this and there is no cost associated with ordering them based on their cost.

mike-w-wilson: This makes complete sense

mike-w-wilson:
@lfrancioli , I agree with moving much of this code to `gnomad_methods`  and happy to take on the subsetting/vcf module. For now I think it makes sense to merge this here and once the `gnomad_methods`  reorg is complete, which I'm working on now, I will move the sensible portions of this code to a `gnomad_methods.utils.vcf` submodule. Is there any downside to merging this to `gnomad_qc` now? If not, I can add the comment but will also work on getting this into gnomad_methods very soon.

mike-w-wilson:
Changed my mind on this. I'm going to reorg gnomad_methods, move the header, compute partitions and format  function to a vcf module in gnomad_method and then update this script. Stay tuned!

gnomad_qc/v3/subset/subset.py
```diff
@@ -0,0 +1,292 @@
...
+            "Description": "allele specific Z-score from Wilcoxon rank sum test of each Alt vs. Ref read position bias"
+        },
+        "AS_MQRankSum": {
+            "Description": "Allele-specifc Z-score from Wilcoxon rank sum test of alternate vs. reference read mapping qualities"
+        },
+        "AS_RAW_MQ": {
+            "Description": "Allele-specifc raw root mean square of the mapping quality of reads across all samples"
+        },
+        "AS_QUALapprox": {
+            "Description": "Allele-specifc sum of PL[0] values; used to approximate the QUAL score"
```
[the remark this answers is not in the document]
mike-w-wilson: Most of these definitions were taken from: https://github.com/broadinstitute/gatk/blob/master/src/test/resources/org/broadinstitute/hellbender/tools/walkers/GnarlyGenotyper/twoSampleASDB.vcf. I combine the QUALapprox definition with Allele-specific but I think in this case it might be best to leave it to the default description to avoid confusion?

gnomad_qc/v3/subset/subset.py
```diff
@@ -0,0 +1,292 @@
...
+        },
+        "AS_MQ_DP": {
+            "Description": "Allele-specifc depth over variant samples for better MQ calculation"
+        },
+        "AS_VarDP": {
+            "Description": "Allele-specific (informative) depth over variant genotypes -- including ref, RAW format"
```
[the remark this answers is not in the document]
mike-w-wilson: Similarly, definition was taken from https://github.com/broadinstitute/gatk/blob/master/src/test/resources/org/broadinstitute/hellbender/tools/walkers/GnarlyGenotyper/twoSampleASDB.vcf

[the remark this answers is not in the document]
jkgoodrich: Currently @ch-kr has this in the common code `"Description": "Depth over variant genotypes (does not include depth of reference samples)"` I think it is fine to keep as is for now for this callset (gnomad v3.0 currently looks like there is no description), but we should decide on the best explanation and be consistent within our callsets.

gnomad_qc/v3/subset/subset.py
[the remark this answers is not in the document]
jkgoodrich: I actually don't see `QUALapprox` or `AS_QUALapprox` in ether of the UKBB or gnomad v3.0 headers. I am fine leaving as is if this is the default then

gnomad_qc/v3/subset/subset.py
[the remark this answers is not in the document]
mike-w-wilson: I agree -- gnomad_methods should be the truth data.

gnomad_qc/v3/subset/subset.py
```diff
@@ -0,0 +1,292 @@
...
+            "Description": "Z-score from Wilcoxon rank sum test of alternate vs. reference read position bias"
+        },
+        "MQRankSum": {
+            "Description": "Z-score from Wilcoxon rank sum test of alternate vs. reference read mapping qualities"
+        },
+        "RAW_MQ": {
+            "Description": "Raw root mean square of the mapping quality of reads across all samples"
+        },
+        "QUALapprox": {
+            "Description": "Sum of PL[0] values; used to approximate the QUAL score"
```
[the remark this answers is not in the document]
mike-w-wilson: I'll leave this based off of
> I actually don't see `QUALapprox` or `AS_QUALapprox` in ether of the UKBB or gnomad v3.0 headers. I am fine leaving as is if this is the default then

gnomad_qc/v3/subset/subset.py
```diff
@@ -0,0 +1,292 @@
...
+    if subset:
+        gt_expr = "LGT" if sparse else "GT" #assumes matrix table is not split - infer 
+        mt = subset_samples_and_variants(mt, subset, sparse=sparse, gt_expr=gt_expr)
+
+    if args.add_meta:
+        meta = meta.semi_join(mt.cols())
+        meta.export(f"{args.output_path}/metadata.tsv.bgz")
+
+    if sparse:
+        mt = hl.experimental.sparse_split_multi(mt, filter_changed_loci=True)
```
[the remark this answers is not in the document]
mike-w-wilson: I'm not sure if these were originally filtered out but I havent had issues with the two previous subsets I ran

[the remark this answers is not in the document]
jkgoodrich: OK, sounds good

gnomad_qc/v3/subset/subset.py
```diff
@@ -0,0 +1,292 @@
...
+        mt = mt.annotate_rows(info=mt.info.annotate(subset_AC=mt.subset_callstats.AC[1],
+                                                    subset_AN=mt.subset_callstats.AN,
+                                                    subset_AF=hl.float32(mt.subset_callstats.AF[1])))
+        mt = mt.drop('subset_callstats')
+        header_dict['info'].update({"subset_AC": {"Description": "Alternate Allele count in subset post-filtering"} , 
+                                    "subset_AN": {"Description": "Ref and Alternate allele number in subset post-filtering"},
+                                    "subset_AF": {"Description": "Alternate Allele frequency in subset post-filtering"}})
+    
+    if args.add_gnomad_freqs:
+        from gnomad_qc.v3.resources.annotations import freq
```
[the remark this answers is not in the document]
mike-w-wilson: To not crowd the namespace with unnecessary variables/methods but I'm not sure it's really necessary since this is not yet generalized for wider use. Im going to move it up.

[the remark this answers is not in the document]
jkgoodrich: Thank you!

gnomad_qc/v3/subset/subset.py
```diff
@@ -0,0 +1,292 @@
...
+    if "SB" in info_ht.info and not isinstance(
+        info_ht.info.SB, hl.expr.ArrayNumericExpression
+    ):
+        info_ht = info_ht.annotate(
+            info=info_ht.info.annotate(SB=hl.flatten(info_ht.info.SB))
+        )
+
+    meta = metadata.ht()
+
+    if pop and pop in GENOME_POPS:
```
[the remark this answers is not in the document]
mike-w-wilson: I've added choices to the pop arg which errors out when the specified pop is not in GENOME_POPS as soon as the script is first started.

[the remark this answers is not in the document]
jkgoodrich: That is great, thank you!

gnomad_qc/v3/subset/subset.py
```diff
@@ -0,0 +1,292 @@
...
+    mt = mt.drop(mt.gvcf_info)
+    mt = mt.annotate_rows(info=info_ht[mt.row_key].info)
+
+    if args.subset_call_stats:
+        mt = mt.annotate_rows(subset_callstats=hl.agg.call_stats(mt.GT,mt.alleles))     
+        mt = mt.annotate_rows(info=mt.info.annotate(subset_AC=mt.subset_callstats.AC[1],
+                                                    subset_AN=mt.subset_callstats.AN,
+                                                    subset_AF=hl.float32(mt.subset_callstats.AF[1])))
+        mt = mt.drop('subset_callstats')
+        header_dict['info'].update({"subset_AC": {"Description": "Alternate Allele count in subset post-filtering"} ,
```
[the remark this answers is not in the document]
mike-w-wilson: I added in the adj and raw callstats for subset as well as updated the header dict to include number where necessary.

mike-w-wilson:
Thanks for the quick feedback @jkgoodrich , back to you!

gnomad_qc/v3/subset/subset.py
```diff
@@ -21,56 +21,84 @@
...
-        "DP": {"Description": "Approximate read depth"},
         "RGQ": {"Description": ""},
-    },
+        },
     "info": {
-        "AC": {"Description": "Alternate Allele count in gnomAD post-filtering"},
-        "AC_raw": {"Description": "Raw alternate allele count in gnomAD"},
+        "AC": {
+            "Number": "A",
+            "Description": "Alternate Allele count in gnomAD after filtering out low-confidence genotypes (GQ < 20; DP < 10; and AB < 0.2 for het calls)"
```
jkgoodrich, line 70: Allele count -> allele count

gnomad_qc/v3/subset/subset.py
```diff
@@ -181,62 +211,112 @@ def main(args):
...
     if args.subset_call_stats:
-        mt = mt.annotate_rows(subset_callstats=hl.agg.call_stats(mt.GT,mt.alleles))     
-        mt = mt.annotate_rows(info=mt.info.annotate(subset_AC=mt.subset_callstats.AC[1],
-                                                    subset_AN=mt.subset_callstats.AN,
-                                                    subset_AF=hl.float32(mt.subset_callstats.AF[1])))
-        mt = mt.drop('subset_callstats')
-        header_dict['info'].update({"subset_AC": {"Description": "Alternate Allele count in subset post-filtering"} , 
-                                    "subset_AN": {"Description": "Ref and Alternate allele number in subset post-filtering"},
-                                    "subset_AF": {"Description": "Alternate Allele frequency in subset post-filtering"}})
+        mt = mt.annotate_rows(subset_callstats_raw=hl.agg.call_stats(mt.GT,mt.alleles))
```
jkgoodrich, line 236: nitpick add space after comma

gnomad_qc/v3/subset/subset.py
```diff
@@ -181,62 +211,112 @@ def main(args):
...
-        header_dict['info'].update({"subset_AC": {"Description": "Alternate Allele count in subset post-filtering"} , 
-                                    "subset_AN": {"Description": "Ref and Alternate allele number in subset post-filtering"},
-                                    "subset_AF": {"Description": "Alternate Allele frequency in subset post-filtering"}})
+        mt = mt.annotate_rows(subset_callstats_raw=hl.agg.call_stats(mt.GT,mt.alleles))     
+        mt = mt.annotate_rows(info=mt.info.annotate(subset_AC_raw=mt.subset_callstats.AC[1],
+                                                    subset_AN_raw=mt.subset_callstats.AN,
+                                                    subset_AF_raw=hl.float32(mt.subset_callstats.AF[1])))
+        mt = mt.drop('subset_callstats_raw')
+        mt = annotate_adj(mt)
+        mt = mt.annotate_rows(subset_callstats_adj=hl.agg.filter(mt.adj, hl.agg.call_stats(mt.GT,mt.alleles)))
```
jkgoodrich, line 242: nitpick add space after comma

gnomad_qc/v3/subset/subset.py
```diff
@@ -181,62 +211,112 @@ def main(args):
...
+                "Number": "A",
+                "Description": "Alternate allele count in subset post-filtering of low-confidence genotypes (GQ < 20; DP < 10; and AB < 0.2 for het calls)",
+                },
+             "subset_AN_adj": {
+                "Number": "1",
+                "Description": "Total number of alleles in subset post-filtering of low-confidence genotypes (GQ < 20; DP < 10; and AB < 0.2 for het calls)"
+                },
+            "subset_AF_adj": {
+                "Number": "A",
+                "Description": "Alternate allele frequency in subset post-filtering of low-confidence genotypes (GQ < 20; DP < 10; and AB < 0.2 for het calls)",
```
jkgoodrich, line 272: Above and below you have `after filtering` here you have `post-filtering`, either is fine, but maybe make the same

gnomad_qc/v3/subset/subset.py
```diff
@@ -181,62 +211,112 @@ def main(args):
...
-                                    "gnomAD_AN_raw": {"Description": "Raw allele number. The total number of called alleles in gnomAD post-filtering"},
-                                    "gnomAD_AF_raw": {"Description": "Raw alternate allele frequency in gnomAD post-filtering"}})
+        freq_ht = freq.ht()
+        mt = mt.annotate_rows(info=mt.info.annotate(gnomAD_AC_adj=freq_ht[mt.row_key].freq[0].AC, 
+                                                    gnomAD_AN_adj=freq_ht[mt.row_key].freq[0].AN,
+                                                    gnomAD_AF_adj=hl.float32(freq_ht[mt.row_key].freq[0].AF),
+                                                    gnomAD_AC_raw=freq_ht[mt.row_key].freq[1].AC, 
+                                                    gnomAD_AN_raw=freq_ht[mt.row_key].freq[1].AN,
+                                                    gnomAD_AF_raw=hl.float32(freq_ht[mt.row_key].freq[1].AF)))
+        header_dict['info'].update({
```
jkgoodrich, line 285: What is the difference between these and the AC, AN, AF... in the `HEADER_DICT` above?

gnomad_qc/v3/subset/subset.py
```diff
@@ -256,27 +336,21 @@ def main(args):
     )
 
     parser.add_argument(
-        "-s", "--subset", help="Path to subset table of sample IDs with header: s"
+        "-s", "--subset", help="Path to a text file with sample IDs for subsetting and a header: s"
     )
-    parser.add_argument("--pop", help="Subset from specific population") # add choices for gnomad pops
+    parser.add_argument("--pop", help=f"Subset to a specific population", choices=GENOME_POPS)
```
jkgoodrich, line 341: Thank you for adding!

gnomad_qc/v3/subset/subset.py
mike-w-wilson: Good question, I've reached out to Laurent for help answering this. I'm thinking of dropping these to avoid any confusion.

mike-w-wilson: After speaking with Laurent and investigating the `AC` and `AC_raw` in the info file compared to the frequencies file, `AC` and `AC_raw` are computed on all samples while the frequencies ACs are computed on release only. I am dropping `AC` and `AC_raw` from the info_ht before annotating.

jkgoodrich: Ah, yes, because this is the raw table. For the sites table I think he resets these to only the release samples (and high quality unrelated). I am assuming AN and AF also should be dropped. The only question is for subsets will we want AC, AN, and AF to now be reset to the numbers in the subset rather than having a subset_AC,... Also I think we can drop the '_adj' on everything. The way it is in gnomAD and UKBB is that we specify if it is raw, but otherwise it is adj and it is described in the header what it is

mike-w-wilson: There isn't a AN or AF in the raw table. I'm not sure I understand this sentence "The only question is for subsets will we want AC, AN, and AF to now be reset to the numbers in the subset rather than having a subset_AC" Are you suggesting renaming the `subset_AC` annotation to `AC` so on for the other callstat annotations? dropping `_adj` from the annotations and updating headers makes sense to me, especially if it is consistent with gnomAD and UKBB. Thanks again!

mike-w-wilson: Sorry, reread the comment and now I understand changing the subset_AC to AC. I think we'll just have to be clear in the email but that works for me.

jkgoodrich: Yeah I was wondering if you think we instead of `subset_AC` we name it `AC`,  `subset_AN` we name it `AN`, and `subset_AF` we name it `AF`. Since they will be a reflection of the samples in the VCF.

mike-w-wilson: This makes sense. I've updated the annotations so if you wouldnt mind one last look, I'd appreciate it

gtiao:
Clarifying note: this script is more focused on creating separate sub-callsets of gnomAD and not about creating frequency array annotations for subsets of gnomAD

gtiao:
May no longer be relevant

mike-w-wilson:
Replaced by #566

PR #53, opened 2020-05-07
Add picard filters to sample_qc

mike-w-wilson:
Small PR. Added picard metric filtering to sample_qc and the HT to resources. The thresholds for each metric are the same from v2. Sample callrate is not included but I can add it in. Since this would require a densify, we could use a similar approach to what KC did in UKBB: https://github.com/macarthur-lab/ukbb_qc/blob/d8fcc39342f9e203db3a7972bec688a236b1931e/ukbb_qc/utils/sparse_utils.py#L16

### DIFF
--- a/gnomad_qc/v3/sample_qc/sample_qc.py
+++ b/gnomad_qc/v3/sample_qc/sample_qc.py
@@ -103,7 +103,6 @@ def compute_qc_mt() -> hl.MatrixTable:
 
 def compute_hard_filters(cov_threshold: int) -> hl.Table:
     ht = get_gnomad_v3_mt(remove_hard_filtered_samples=False).cols()
-
     hard_filters = dict()
 
     # Remove samples failing fingerprinting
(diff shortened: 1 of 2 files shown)


jkgoodrich:
For this code it was very specific to exomes since we looked at callrate per interval per sample and used it in other steps of thee pipeline too that were exome specific. It did still require a densify step though. https://github.com/macarthur-lab/ukbb_qc/blob/d8fcc39342f9e203db3a7972bec688a236b1931e/ukbb_qc/utils/sparse_utils.py#L16

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -135,6 +134,13 @@ def compute_hard_filters(cov_threshold: int) -> hl.Table:
     hard_filters['ambiguous_sex'] = (sex_ht.sex_karyotype == 'Ambiguous')
     hard_filters['sex_aneuploidy'] = ~hl.set({'Ambiguous', 'XX', 'XY'}).contains(sex_ht.sex_karyotype)
 
+    # Remove samples that fail picard metric thresholds, percents are not divided by 100, e.g. 5% == 5.00, %5 != 0.05
```
jkgoodrich, line 137: Thanks for adding this comment!

PR #100, opened 2020-09-11
Update v3 resources to account for other versions

jkgoodrich:


mike-w-wilson:
I've updated the resources to use versioned table and matrixtable resources. One odd piece is the use of `"v3.1_raw" in the `create_fam.py` code. Because each release has a `final` and a `raw` version of the ped/trios, I opted to add the version's to the `raw` strings, used `v3_final` for the final version of v3 ped and trios, and left `final` for the current version. I am open to suggestions around this but after discussing options with Laurent, this seemed the best and least clunky.

gnomad_qc/v3/annotations/generate_qc_annotations.py
```diff
@@ -30,6 +27,7 @@ def compute_info() -> hl.Table:
     mt = get_gnomad_v3_mt(key_by_locus_and_alleles=True, remove_hard_filtered_samples=False)
     mt = mt.filter_rows((hl.len(mt.alleles) > 1))
     mt = mt.transmute_entries(**mt.gvcf_info)
+    mt = mt.annotate_rows(alt_alleles_range_array=hl.range(1, hl.len(mt.alleles)))
```
nawatts, line 30: Might be worth adding a comment about why this is necessary. It's not obvious that this is used by a function in gnomad_methods. (#109 and broadinstitute/gnomad_methods#258 for context)

gnomad_qc/v3/resources/annotations.py
```diff
@@ -1,42 +1,137 @@
...
     """
-    path = '{}/{}_filtering{}.ht'.format(
-        ANNOTATIONS_ROOT,
-        model_id,
-        '.split' if split else ''
+    return VersionedTableResource(
+        CURRENT_RELEASE,
+        {
+            release: TableResource(
+                f"{_annotations_root(release)}/filtering/{model_id}{'.finalized' if finalized else ''}{'.split' if split else ''}.ht"
```
nawatts, line 54: Nitpick about consistency of using `format` vs f strings. I find `format` a bit easier to read with the ternary expressions.

gnomad_qc/v3/resources/basics.py
```diff
@@ -0,0 +1,67 @@
...
+    :param key_by_locus_and_alleles: Whether to key the MatrixTable by locus and alleles (only needed for v3)
+    :param remove_hard_filtered_samples: Whether to remove samples that failed hard filters (only relevant after sample QC)
+    :param release_only: Whether to filter the MT to only samples available for release (can only be used if metadata is present)
+    :param samples_meta: Whether to add metadata to MT in 'meta' column
+    :return: gnomAD v3 dataset with chosen annotations and filters
+    """
+    mt = gnomad_v3_genotypes.mt()
+    if key_by_locus_and_alleles:
+        mt = hl.MatrixTable(
+            hl.ir.MatrixKeyRowsBy(mt._mir, ["locus", "alleles"], is_sorted=True)
```
nawatts, line 30: Why use this vs `mt.key_rows_by`? Might be worth a comment.

gnomad_qc/v3/resources/meta.py
```diff
@@ -1,37 +1,113 @@
...
+            "min_partitions": 100,
+        },
+    ),
+}
+_project_meta_versions["v3"] = _project_meta_versions["09-09-2019"]
+_project_meta_versions["v3.1"] = _project_meta_versions["2020-09-11"]
+
+
+_pedigree_versions = {
+    "final": PedigreeResource(
```
nawatts, line 68: Why use "final" here vs "v3.1"? Same question for `_trios_versions`.

gnomad_qc/v3/resources/release.py
```diff
@@ -1,13 +1,19 @@
...
 # TODO: Remove if not used after all python files are in
 # internal_ht_path = 'gs://gnomad/release/3.0/ht/gnomad.genomes.r3.0.nested.no_subsets.sites.ht'
 
 
-def release_ht_path(data_type: str = 'genomes', release_version: str = V3_RELEASE_VERSION, public: bool = True) -> str:
+def release_ht_path(
+    data_type: str = "genomes",
+    release_version: str = CURRENT_RELEASE,
+    public: bool = True,
+) -> TableResource:
```
nawatts, line 16: Looks like this returns a string instead of a `TableResource`.

gnomad_qc/v3/resources/release.py
```diff
@@ -1,13 +1,19 @@
-V3_RELEASE_VERSION = 'r3.0'
+from gnomad.resources.resource_utils import TableResource
 
-qual_hists_json_path = 'gs://gnomad/release/3.0/json/gnomad.genomes.r3.json'
+from gnomad_qc.v3.resources.constants import CURRENT_RELEASE
+
+qual_hists_json_path = "gs://gnomad/release/3.0/json/gnomad.genomes.r3.json"
```
nawatts, line 5: Does this need to parameterized with a version similar to other paths? (Is this going to get overwritten in a v3.1 pipeline)

gnomad_qc/v3/resources/sample_qc.py
```diff
@@ -1,69 +1,112 @@
...
-        '_with_unreleasable_samples' if include_unreleasable_samples else '',
-        'txt' if part == 'eigenvalues' else 'ht'
+        "_with_unreleasable_samples" if include_unreleasable_samples else "",
+        "txt" if part == "eigenvalues" else "ht",
     )
 
 
-def get_ancestry_pca_loadings(include_unreleasable_samples: bool = False) -> TableResource:
+def ancestry_pca_loadings(include_unreleasable_samples: bool = False
+) -> TableResource:
```
nawatts, line 60: This returns a `VersionedTableResource` but is annotated as returning `TableResource`. It's not wrong, a `VersionedTableResource` is a `TableResource`.

gnomad_qc/v3/resources/sample_qc.py
```diff
@@ -1,69 +1,112 @@
...
+        {release: TableResource(_get_ancestry_pca_ht_path("scores", release, include_unreleasable_samples)) for release in RELEASES}
+    )
 
 
-def get_ancestry_pca_eigenvalues_path(include_unreleasable_samples: bool = False) -> str:
+def ancestry_pca_eigenvalues_path(include_unreleasable_samples: bool = False
+) -> VersionedTableResource:
     """
-    Gets the ancestry PCA eigenvalues path.
+    Gets the path to file containing ancestry PCA eigenvalues
```
nawatts, line 91: Docstring isn't quite accurate... this function returns a resource instead of a path and has no `version` parameter.

gnomad_qc/v3/resources/sample_qc.py
```diff
@@ -73,55 +116,124 @@ def get_v3_relatedness_annotated_ht() -> Table:
...
 
 # Hard-filtered samples
-hard_filtered_samples = TableResource(f'{SAMPLE_QC_ROOT}/gnomad_v3_hard_filtered_samples.ht')
+hard_filtered_samples = VersionedTableResource(
+    CURRENT_RELEASE,
+    {release: TableResource(f"{get_sample_qc_root(release)}/gnomad_v{release}_hard_filtered_samples.ht") for release in RELEASES}
+)
 
-# Results of running population-based metrics filtering
+# Results of running population-based metrics filtering	    :param version: Version of sample QC path to return
```
nawatts, line 192: Looks like `:param version: ...` was unintentionally pasted here? Same thing for `release_samples_rankings` below.

gnomad_qc/v3/resources/sample_qc.py
```diff
@@ -73,55 +116,124 @@ def get_v3_relatedness_annotated_ht() -> Table:
...
-# Inferred populations
-pop = TableResource(f'{SAMPLE_QC_ROOT}/gnomad_v3_pop.ht')
-pop_tsv_path = 'gs://gnomad/sample_qc/temp/genomes_v3/gnomad_v3_RF_pop_assignments.txt.gz'
-pop_rf_path = 'gs://gnomad/sample_qc/temp/genomes_v3/gnomad_v3_pop.RF_fit.pickle'
+
+def pop_tsv_path(version: str = CURRENT_RELEASE) -> str:
+    """
+    Path to tab delimited file indicating inferred sample populations
+
+    :param version: Version of sample QC path to return
```
nawatts, line 171: Is "sample QC" the correct description here? Same question for `pop_rf_path`.

nawatts (approved):

gnomad_qc/v3/resources/basics.py
mike-w-wilson: The MT is already sorted by locus and here each locus only appears once so this prevents hail from running it's sorting algorithm to save on runtime. Will add the comment.

gnomad_qc/v3/resources/meta.py
mike-w-wilson: I was trying to reduce changes to the code, e.g. L103 and L104, moving forward but I think the trade off with the current approach is less clarity and consistency so I'll update to "v3.1" for both.

gnomad_qc/v3/resources/release.py
mike-w-wilson: Good catch, I'll add this!

gnomad_qc/v3/resources/sample_qc.py
mike-w-wilson: I think gnomAD version is more clear.

gnomad_qc/v3/resources/sample_qc.py
mike-w-wilson: Yes, thank you -- this slipped by me.

PR #112, opened 2020-09-17
hotfix for depletion of homozygous alternate genotypes to frequency code

jkgoodrich:


gnomad_qc/v3/annotations/generate_freq_data.py
```diff
@@ -42,6 +42,24 @@ def main(args):
     mt = hl.experimental.densify(mt)
     mt = mt.filter_rows(hl.len(mt.alleles) > 1)
 
+    logger.info("Setting het genotypes at sites with >1% AF and > 0.9 AB to homalt...")
+    # hotfix for depletion of homozygous alternate genotypes
+    # Use AC / 2*n_samples for AF. This doesn't take missing into account but avoids extra frequency calculation
+    # TODO: this is fine for genomes, but will need to be modified for v4
+    info_ht = get_info(split=True).ht()
```
gtiao, line 49: Since you're pulling the v3 AFs for other parts of the code that call for an on-the-fly AF calculation, what do you think about doing it here as well, for consistency? I'm also thinking that since we'll be revising the pipeline in the near future to do incremental QC, making this the standard move (pulling the previous callset's AFs) might make sense.

The only catch would be that if we had a very large increment of new samples, AFs on the border might tip in one direction or another. What do you think? We could also discuss in the general QC channel

gtiao (commented):
Hotfix code looks good, just a question about using v3 AFs instead of using the 2*AC/n_samples approximation

gnomad_qc/v3/annotations/generate_freq_data.py
jkgoodrich: I agree especially since this is only changing common variants. I will add a note though about needing to think about this method for future releases. At this point I will just hardcode the use of v3, but we will likely want to parameterize this in the future

gnomad_qc/v3/annotations/generate_freq_data.py
```diff
@@ -1,7 +1,7 @@
 from gnomad.utils.annotations import annotate_freq, qual_hist_expr, pop_max_expr, faf_expr
 from gnomad.utils.annotations import get_adj_expr
 from gnomad.sample_qc.sex import adjusted_sex_ploidy_expr
-from gnomad_qc.v3.resources import get_gnomad_v3_mt, meta, freq
+from gnomad_qc.v3.resources import get_gnomad_v3_mt, get_info, meta, freq
```
gtiao, line 4: I think you don't need `get_info` any more now that we're pulling v3 AFs

gtiao (approved):

PR #114, opened 2020-09-17
Make modifications to the compute_sex code

jkgoodrich:
Changed to use v3 AF instead of AC / (2 * n_samples)


gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -152,42 +152,26 @@ def compute_hard_filters(cov_threshold: int) -> hl.Table:
 
 
 def compute_sex() -> hl.Table:
-    # Compute sex chrom poloidy
-    ht = impute_sex_ploidy(
-        get_gnomad_v3_mt(remove_hard_filtered_samples=False),
-        excluded_calling_intervals=telomeres_and_centromeres.ht()
+    mt = get_gnomad_v3_mt(
```
ch-kr, line 155: this is imported with `from gnomad_qc.v3.resources import *`, right? basic question, is there a reason why this is better here than importing each function individually?

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -152,42 +152,26 @@ def compute_hard_filters(cov_threshold: int) -> hl.Table:
...
+        key_by_locus_and_alleles=True,
+        remove_hard_filtered_samples=False,
     )
-    ht = ht.checkpoint('gs://gnomad-tmp/sex_depth.ht', overwrite=True)
+    mt = mt.filter_rows((hl.len(mt.alleles) == 2))
```
ch-kr, line 159: I don't think you need this filter, since you want to compute depth from non ref sites and ref blocks when imputing ploidies

ch-kr (changes requested):
one question and one small change

gnomad_qc/v3/sample_qc/sample_qc.py
jkgoodrich: Yes it is, I think I would prefer to import each individually, but will need to go back and change all that in another PR at somepoint

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -159,13 +159,14 @@ def compute_sex() -> hl.Table:
 
     # Use AF from v3
     freq_ht = freq(versions="3").ht()
-    mt = mt.annotate_rows(AF=freq_ht[mt.row_key].freq[0].AF)
+    freq_ht = freq_ht.select(AF=freq_ht[mt.row_key].freq[0].AF)
```
ch-kr, line 162: maybe you should filter this to the AAF threshold too?

ch-kr (approved):
just one last thing, then this looks good to me!

PR #120, opened 2020-09-22
Add inbreeding sample qc calculation using population AF

jkgoodrich:
Untested, but wanted to get it looked at before I get to that point


gtiao (approved):
LGTM!

PR #136, opened 2020-12-09
Add vqsr loading

jkgoodrich:


gnomad_qc/v3/load_data/load_vqsr.py
```diff
@@ -0,0 +1,125 @@
...
+        args.header_path,
+    )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--vqsr_vcf_path", help="Path to VQSR VCF")
+    parser.add_argument(
+        "--vqsr_type",
+        help="Path to VQSR VCF. Can be specified as Hadoop glob patterns",
```
williamphu, line 96: I think this help description doesn't match the description from the docstrings above? Looks like this is the description for `--vqsr_vcf_path`

gnomad_qc/v3/load_data/load_vqsr.py
```diff
@@ -0,0 +1,125 @@
...
+        default="alleleSpecificTrans",
+        choices=["classic", "alleleSpecific", "alleleSpecificTrans"],
```
williamphu, line 98: Do the choices line up with expected inputs for `vqsr_type`? The docstring makes it seem like the input should be "AS" or "AS_TS"

williamphu (changes requested):
Loading looks good. There might be some things to improve about the argument descriptions and making sure the arguments from `argparser` line up with the function `import_vqsr.`

williamphu (approved):
Looks good!

PR #138, opened 2020-12-09
Add make var annotation hists

jkgoodrich:


gnomad_qc/v3/create_release/make_var_annot_hists.py
```diff
@@ -1,61 +1,164 @@
...
+                ).map(lambda x: x[1].annotate(metric="AS_QUALapprox-"+x[0]))
             ),
-            _localize=False
+            _localize=False,
         )
 
-        with hl.hadoop_open(qual_hists_json_path(CURRENT_RELEASE), 'w') as f:
-            f.write(hl.eval(hl.json(hists)))
+        # Defining hist range and bins for allele frequency groups because they needed different ranges
+        inbreeding_bin_ranges = {'under_0.0005': [-0.0005, 0.0005, 50], 'over_0.0005': [-0.25, 1, 50]}
```
mike-w-wilson, line 141: Could this be parameterized? Or maybe defined as a global after LOG10_ANNOTATIONS so it is a bit easier to read? Or maybe just a new line?

gnomad_qc/v3/create_release/make_var_annot_hists.py
```diff
@@ -1,61 +1,164 @@
...
+                        ht.info.InbreedingCoeff,
+                        *inbreeding_bin_ranges[x],
+                    )
+                )
+            ).annotate(metric="InbreedingCoeff" + "-" + x) for x in inbreeding_bin_ranges
+        ]
+
+        hists = hl.eval(hl.json(hists))
+        inbreeding_hists = hl.eval(hl.json(inbreeding_hists))
+        hists = hists[:-1] + "," + inbreeding_hists[1:]
```
mike-w-wilson, line 157: For my own knowledge, why is the last item remove from hists and first from inbreeding_hists?

gnomad_qc/v3/create_release/make_var_annot_hists.py
```diff
@@ -1,61 +1,164 @@
...
+                ]
             ).extend(
                 hl.array(
                     hl.agg.group_by(
                         create_frequency_bins_expr(
-                            AC=ht.freq[1].AC,
-                            AF=ht.freq[1].AF
+                            AC=ht.freq[1].AC, AF=ht.freq[1].AF
+                        ),
+                        # Decided to use QUALapprox because its formula is easier to interpret than QUAL's
```
mike-w-wilson, line 117: Should this be removed? Or made a NOTE:?

mike-w-wilson (commented):
Mostly style and a couple things on notes and loggers. I think black may need to be run as well.

gnomad_qc/v3/create_release/make_var_annot_hists.py
```diff
@@ -1,61 +1,164 @@
...
+    return bin_expr
+
+
 def main(args):
     hl.init(default_reference='GRCh38', log='/variant_histograms.log')
 
-    ht = hl.read_table(release_ht_path())
+    logger.info("Loading ANNOTATIONS_HISTS dictionary...")
+    if not file_exists(annotation_hists_path()):
+        raise DataException(
```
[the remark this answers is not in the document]
nawatts: I think it's preferable to stick with `DataException` or some other user-defined exception class here since this is wrapping both `hl.hadoop_exists` and `os.path.exists`.

Not sure that OSError (and by extension FileNotFoundError) would be appropriate since those are defined as being raised by "system functions"
> This exception is raised when a system function returns a system-related error, including I/O failures such as “file not found” or “disk full” (not for illegal argument types or other incidental errors).
> -- https://docs.python.org/3/library/exceptions.html#OSError

We could define some more specific exception classes in gnomAD, including a "not found" error class.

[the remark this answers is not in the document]
mike-w-wilson: Thanks for explaining as always Nick! Please ignore my suggestion Julia!

gnomad_qc/v3/create_release/make_var_annot_hists.py
```diff
@@ -1,61 +1,164 @@
...
+
+    :param AC: Field in input that contains the allele count information
+    :param AF: Field in input that contains the allele frequency information
+    :return: Expression containing bin name
+    :rtype: hl.expr.StringExpression
+    """
+    bin_expr = (
+        hl.case()
+            .when(AF < 0.0005, "under_0.0005")
+            .when((AF >= 0.0005) & (AF <= 1), "over_0.0005")
```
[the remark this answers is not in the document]
jkgoodrich: nope certainly shouldn't, will change

gnomad_qc/v3/create_release/make_var_annot_hists.py
jkgoodrich: I'm going to mess around with this one again. The reason is I am just removing the `}` from the first and the `{` from the second. I think there is a better way to do this, but at the time of running it I was getting an error because the types of hists (before hl.json) didn't match the type of inbreeding hists and trying to cast to the same type was still giving an error. This was my quick and dirty way of getting the hists quickly because all I really needed was to append the two together.

jkgoodrich: Yeah so the problem I had run into was that inbreeding_hists has this structure `<ArrayStructExpression of type array<struct{bin_edges: array<float64>, bin_freq: array<int32>, n_smaller: int32, n_larger: int32, metric: str}>>`  and hists has `<ArrayStructExpression of type array<struct{bin_edges: array<float64>, bin_freq: array<int64>, n_smaller: int64, n_larger: int64, metric: str}>>` so I can't extend the hists array as is. `ArrayExpression.extend' expects 'a' to be the same type as the caller`

gnomad_qc/v3/create_release/make_var_annot_hists.py
jkgoodrich: I struggled with the best way to handle this because currently the ranges for the other metrics are defined as a json file and each metric has one list with min max and bins. For inbreeding we needed the ranges to be different because of how different the values are compared to frequency. I could either do the global, but then we need to actually modify the code if it changes. Or try changing the code to be able to handle something like this is the json and handle list vs dict differently.

```
{'AS_FS': [0, 30, 50],
 'AS_MQ': [50, 60, 40],
 'AS_MQRankSum': [-5, 5, 60],
 'AS_QD': [0, 40, 40],
 'AS_ReadPosRankSum': [-5, 5, 60],
 'AS_SOR': [0, 5, 50],
 'AS_VarDP': [1, 10, 40],
 'AS_VQSLOD': [-30, 30, 60],
 'AS_pab_max': [0, 1, 50],
 'QUALapprox': [1, 10, 40],
 'AS_QUALapprox': [1, 10, 40],
 'InbreedingCoeff': {"under_0.0005": [-0.0005, 0.0005, 50], "over_0.0005": [-0.25, 1, 50],
}}```

File: gnomad_qc/v3/create_release/make_var_annot_hists.py [hunk 6]
```diff
@@ -1,61 +1,164 @@
...
+                    1e10,
+                ),
+            )
         minmax = ht.aggregate(hl.struct(**minmax_dict))
-        print(minmax)
+        logger.info(f"Metrics bounds: {minmax}")
     else:
-        # Aggregate hists over hand-tooled ranges
+        logger.info(
+            "Aggregating hists over ranges determined using first pass run..."
```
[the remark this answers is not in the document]
jkgoodrich: the defaults (`annotation_hists_path`) can be used without a first pass, but we usually run the first pass and I also usually make histograms visually too just because I find it helpful. But yes that logger is confusing as it implies only the first pass values will be used not that you actually need to change the values in `annotation_hists_path` after doing the first pass if you want to use them

gnomad_qc/v3/create_release/make_var_annot_hists.py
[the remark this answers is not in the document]
nawatts: Then again... I couldn't really find any guidance on this. Hail uses a lot of OSError subclasses, such as FileNotFoundError, in their code.

¯\\_(ツ)_/¯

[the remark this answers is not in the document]
jkgoodrich: I'm fine with either, FileNotFoundError does make more sense though

gnomad_qc/v3/create_release/make_var_annot_hists.py
williamphu: Also for my knowledge, what error did it give when you tried to cast one of the hists to the same type as the other? Is the current issue appending int32 vs int64?

gnomad_qc/v3/create_release/make_var_annot_hists.py
```diff
@@ -1,71 +1,173 @@
...
+    hl.init(default_reference="GRCh38", log="/variant_histograms.log")
+
+    logger.info("Loading ANNOTATIONS_HISTS dictionary...")
+    if not file_exists(annotation_hists_path()):
+        raise DataException(
+            "Annotation hists JSON file not found. Need to create this JSON before running script!"
+        )
+
+    with hl.hadoop_open(annotation_hists_path()) as a:
+        ANNOTATIONS_HISTS = json.loads(a.read())
```
williamphu, line 67: Is `ANNOTATIONS` in `ANNOTATIONS_HISTS` supposed to be plural while it's singular `annotation` in `annotation_hists_path()`?

williamphu (commented):
Just a few thoughts on style.

gnomad_qc/v3/create_release/make_var_annot_hists.py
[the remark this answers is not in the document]
jkgoodrich: @mike-w-wilson what is your preference?

gnomad_qc/v3/create_release/make_var_annot_hists.py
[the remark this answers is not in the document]
jkgoodrich: @mike-w-wilson I changed to `determine_bounds`, let me know if you think the changes I made address this comment, or if there is something else that would help

gnomad_qc/v3/create_release/make_var_annot_hists.py
jkgoodrich: @mike-w-wilson thoughts on these options?

gnomad_qc/v3/create_release/make_var_annot_hists.py
jkgoodrich: This is imported from `gnomad_methods` and I think used in other repos so I think I will leave `ANNOTATIONS_HISTS` I think it should probably be singular though

gnomad_qc/v3/create_release/make_var_annot_hists.py
mike-w-wilson: I've struggled to come up with an alternative here without adding way more code. The python is just so much shorter without casting things. I think this works but maybe just add a comment on what is being removed? Alternatively, join and then just hist.remove('{') and hist.remove('}') ?

mike-w-wilson (commented):
This looks good, just two things for more discussion as I'm also struggling to come up with a different approach to the inbreedingCoeff hists and the join with the hists.

gnomad_qc/v3/create_release/make_var_annot_hists.py
jkgoodrich: Agree that sounds good. I was thinking that the other option provided the ability to have multiple metrics with the dict form if needed, and wouldn't require `InbreedingCoeff` to have only under and over, though we have already specified that in `create_frequency_bins_expr_inbreeding` so I actually think this would be better for the current code and more thought could be put in to generalize it later if/when needed

gnomad_qc/v3/create_release/make_var_annot_hists.py
jkgoodrich: Sound's good to me, agree I couldn't find a better way currently and this is overall relatively clean and represents what was done in v3.1. I will add a comment

gnomad_qc/v3/create_release/make_var_annot_hists.py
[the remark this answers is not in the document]
jkgoodrich: OK, leaving as is, it won't impact things overall as there is still na error, but we can cleanup in the future if we would rather something else for our standards

gnomad_qc/v3/create_release/make_var_annot_hists.py
jkgoodrich: Because of some issues in code above this I went with my first suggestion and added a note. I'm still not sure I really like it, but I am struggling to think of better ways. This code is certainly not very generalized, but I suppose it is in the QC repo so its not as big of a deal.

gnomad_qc/v2/variant_qc/prepare_data_release.py
```diff
@@ -1,4 +1,4 @@
-from gnomad.utils.slack import try_slack
+#from gnomad.utils.slack import try_slack
```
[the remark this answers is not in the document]
jkgoodrich: nope, thanks

gnomad_qc/v3/create_release/make_var_annot_hists.py
mike-w-wilson: That makes sense to me!

mike-w-wilson (approved):
LGTM!

PR #184, opened 2021-03-10
Fixed slack import in v2 prepare data release

ch-kr:
This was causing import issues. Did not fix anything else in this script

### DIFF
--- a/gnomad_qc/v2/variant_qc/prepare_data_release.py
+++ b/gnomad_qc/v2/variant_qc/prepare_data_release.py
@@ -1,4 +1,4 @@
-from gnomad.utils.slack import try_slack
+from gnomad.utils.slack import slack_notifications
 from gnomad_qc.v2.resources.variant_qc import *
 import copy
 import itertools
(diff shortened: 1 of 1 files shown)


gnomad_qc/v2/variant_qc/prepare_data_release.py
```diff
@@ -1051,13 +1051,15 @@ def main(args):
     parser.add_argument('--liftover', help='Liftover final sites file', action='store_true')
     parser.add_argument('--verbose', help='Run sanity checks function with verbose output', action='store_true')
     parser.add_argument('--slack_channel', help='Slack channel to post results and notifications to.')
+    parser.add_argument('--slack_token', help='Slack token to enable slack notifications. Must be set if slack_channel is provided')
```
jkgoodrich, line 1054: Need to add check that this is set if `slack_channel` is set or throw error

jkgoodrich (changes requested):
just one thing

jkgoodrich (approved):
lgtm

PR #187, opened 2021-03-31
Updated v2 public paths

mike-w-wilson:
Updated the public path functions to point to the google hosted HTs.

### DIFF
--- a/gnomad_qc/v2/resources/basics.py
+++ b/gnomad_qc/v2/resources/basics.py
@@ -87,14 +87,14 @@ def get_release_file(file_path: str, version: str = CURRENT_RELEASE) -> str:
 
 def public_exomes_ht_path(split=True, version=CURRENT_RELEASE):
     if int(version[0]) > 1 and int(version[2]) > 0:
-        return get_release_file('gs://gnomad-public/release/{0}/ht/exomes/gnomad.exomes.r{0}.sites.ht')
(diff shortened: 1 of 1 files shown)


gnomad_qc/v2/resources/basics.py
```diff
@@ -87,14 +87,14 @@ def get_release_file(file_path: str, version: str = CURRENT_RELEASE) -> str:
 
 def public_exomes_ht_path(split=True, version=CURRENT_RELEASE):
     if int(version[0]) > 1 and int(version[2]) > 0:
-        return get_release_file('gs://gnomad-public/release/{0}/ht/exomes/gnomad.exomes.r{0}.sites.ht')
+        return get_release_file('gs://gcp-public-data--gnomad/release/{0}/ht/exomes/gnomad.exomes.r{0}.sites.ht')
```
klaricch, line 90: what does {0} refer to and why not just use "version"?

mike-w-wilson: The {0} is version and is formatted in the function get_release_file: https://github.com/broadinstitute/gnomad_qc/blame/76b71e280e7c04a81a73564fd17392672ab934a9/gnomad_qc/v2/resources/basics.py#L60 . This is pretty old so I'm guessing it was before f-strings became the norm for the group. I'm just changing the main bucket here within the path. I don't think it is worth rewriting the module.

PR #212, opened 2021-08-20
Replace try_slack function in v2 scripts

nawatts:
After https://github.com/broadinstitute/gnomad_methods/pull/404, Pylint now notices these broken imports.

This replaces the old `try_slack` function with the newer [`slack_notifications`](https://broadinstitute.github.io/gnomad_methods/api_reference/utils/slack.html#gnomad.utils.slack.slack_notifications).


ch-kr (approved):
this is great, thank you so much for looking into both why these weren't getting caught and fixing them!!

PR #213, opened 2021-08-31
Jg/fix resources

jkgoodrich:
Splitting the resource related changes out of the HGDP + TGP PR to make that PR a bit less daunting.

This PR changes resources because files have been moved around since the script was first created and adds Martin group files to the resources.

It also modifies the resource constants to reflect the fact that for v3.1.1 we used different versions for annotations than the release version, this will also be the case for v3.1.2.

Note: Some resources will no longer work because the files have been deleted.


gnomad_qc/v3/resources/release.py
```diff
@@ -166,50 +165,59 @@ def append_to_vcf_header_path(
...
 
     return VersionedMatrixTableResource(
         default_version=CURRENT_HGDP_TGP_RELEASE,
         versions={
             release: MatrixTableResource(
-                f"gs://gnomad/release/{release}/mt/gnomad.genomes.v{release}.hgdp_1kg_subset{f'_dense' if dense else '_sparse'}.mt"
+                f"{qc_temp_prefix(version=release) if test else f'gs://gnomad/release/{release}/mt/'}/gnomad.genomes.v{release}.hgdp_1kg_subset{f'_dense' if dense else '_sparse'}.mt"
             )
             for release in HGDP_TGP_RELEASES
             if release != "3"
```
nawatts, line 186: This `if` could be removed. "3" is not in HGDP_TGP_RELEASES.

gnomad_qc/v3/resources/constants.py
```diff
@@ -1,8 +1,23 @@
...
-CURRENT_META_VERSION = "3.1"
+
 META_VERSIONS = ["3", "3.1"]
-CURRENT_PROJECT_META_VERSION = "3.1"
+CURRENT_META_VERSION = "3.1"
+
 PROJECT_META_VERSIONS = ["3", "3.1"]
+CURRENT_PROJECT_META_VERSION = "3.1"
+
+ANNOTATION_VERSIONS = ["3", "3.1"]
```
nawatts, line 13: It would probably be worth adding some comments here about why there are so many different version lists when they're all very similar.

nawatts (approved):

gnomad_qc/v3/resources/constants.py
jkgoodrich: Would it be better if I made something like a `DEFAULT_VERSIONS = ["3", "3.1", "3.1.1"]` and then removed where needed with added comments?

nawatts: I think the way it is now is possibly easier to read and understand than adding/removing from other version lists.

Also, maybe the issue is just that I'm not familiar with the different pieces that go into making a gnomAD release. If it's obvious what these are for someone who is familiar, feel free to ignore my comment.

jkgoodrich: No I think this is a good comment, I changed it up a little to simplify it, let me know if it looks better

PR #214, opened 2021-10-06
V3.1.2 runtime fixes

jkgoodrich:
I haven't finished the v3.1.2 release, but there are a few changes that are bigger that I would like to have looked at regarding the HGDP + 1KG files.

-Mary corrected my misunderstanding of the duplicate sample, and it actually doesn't apply to our MT, only the MT that Konrad created, so I removed this
-I had to put in fixes for when we need the "v3.1::" prefix on sample names and when we don't. Basically the outlier file and relatedness files I got from Mary need to have these removed for matching, but our callset MT needs them present when filtering. So I made a few changes to fix this in the frequency code and the HGDP + 1KG subset code

### DIFF
--- a/gnomad_qc/v3/annotations/generate_freq_data.py
+++ b/gnomad_qc/v3/annotations/generate_freq_data.py
@@ -37,7 +37,9 @@ from gnomad_qc.v3.resources.basics import (
     get_gnomad_v3_mt,
     qc_temp_prefix,
 )
+from gnomad_qc.v3.resources.meta import meta
 from gnomad_qc.v3.resources.release import hgdp_1kg_subset_annotations, release_sites
+from gnomad_qc.v3.resources.variant_qc import SYNDIP
 from gnomad_qc.v3.utils import hom_alt_depletion_fix
 
 logging.basicConfig(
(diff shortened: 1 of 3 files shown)


gnomad_qc/v3/annotations/generate_freq_data.py
```diff
@@ -132,10 +134,25 @@ def main(args):
                 "Filtering MT columns to HGDP + 1KG samples that pass specialized sample QC"
             )
             hgdp_1kg_meta = hgdp_1kg_subset_annotations().ht()
+
+            # Note: Need to use sample names with the v3.1:: prefix
```
ch-kr, line 138:
```suggestion
            # Note: Sample IDs in MT have v3.1:: prefix, but sample IDs in hgdp tgp meta do not
            # Need to add prefix to hgdp tgp meta IDs to filter samples in MT correctly
```
this seems more descriptive (assuming I understood this part correctly)

gnomad_qc/v3/create_release/create_hgdp_tgp_subset.py
```diff
@@ -215,6 +213,13 @@ def prepare_sample_annotations() -> hl.Table:
     )
 
     relatedness_ht = get_relatedness_set_ht(relatedness_ht)
+
+    # Note: Needs to be done before adding the relatedness info because the relatedness HT doesn't have the prefix
+    logger.info(
+        "Removing 'v3.1::' from the sample names, these were added because there are duplicates of some 1KG samples"
+        " in the full gnomAD dataset..."
+    )
+    meta_ht = meta_ht.key_by(s=meta_ht.s.replace("v3.1::", ""))
```
ch-kr, line 222: I might not have followed the sample complexities, but it seems like there are some tgp samples that are duplicated in the callset mt/meta ht. keying by just "s" here would create two rows for the same sample in the meta ht, right? 

for example, this table:
```
s    info
sample_1    some_info_here
v3.1::sample_1    more_info_here
```
would become this:
```
s    info
sample_1    some_info_here
sample_1    more_info_here
```
would adding `meta_ht = meta_ht.distinct()` help here? (assuming their annotations are all identical)

PR #217, opened 2021-10-27
Add script to compare freq versions by population

jkgoodrich:
I am not positive that this is the best location for this script, so thoughts are very welcome. I added it to annotations because it creates a HT with chi squared test and fisher exact test annotations comparing frequencies. This script is to get these annotations so that we can start exploring differences in v3 genome frequencies compared to v2 exome frequencies and will be redone when we have v4 exome frequencies. 

ZenHub ticket with additional details: https://app.zenhub.com/workspaces/gnomad-5f4d127ea61afc001d6be50b/issues/broadinstitute/gnomad_qc/190


gnomad_qc/v3/annotations/compare_freq.py
```diff
@@ -0,0 +1,175 @@
...
+        except ResourceNotAvailable:
+            gnomad_public_resource_configuration.source = (
+                GnomadPublicResourceSource.GNOMAD
+            )
+            ht = release_resource_map[v](d).ht()
+
+        logger.info("Filtering %s to only variants that are PASS...", version)
+        ht = ht.filter(hl.len(ht.filters) == 0)
+        freq_index_dict = hl.eval(ht.freq_index_dict)
+        freq_idx = [0, 1] + [
```
williamphu, line 78: What do the `[0,1]` reference/access?

williamphu (changes requested):

williamphu (approved):
LGTM!

PR #225, opened 2022-02-04
switch to using VDS

klaricch:


klaricch:
The test on chr20 completed successfully for compute_subpop_qc_mt.

Cluster:
hailctl dataproc start kmlw --worker-machine-type n1-highmem-8 --num-workers 10 --num-secondary-workers 10 --init gs://gnomad-kristen/mitochondria/broad_master-init.sh --max-idle=30m --requester-pays-allow-all

gnomad_qc/v3/sample_qc/subpop_analysis.py
```diff
@@ -55,19 +55,41 @@ def compute_subpop_qc_mt(
...
+    # Filter to only sites defined in the qc_sites HT
     mt = mt.select_entries(
-        "END", GT=mt.LGT, adj=get_adj_expr(mt.LGT, mt.GQ, mt.DP, mt.LAD)
+        "END", "LA", GT=mt.LGT, adj=get_adj_expr(mt.LGT, mt.GQ, mt.DP, mt.LAD)
     )
 
-    logger.info(f"Checkpointing the QC sites HT of biallelic SNVs that are not in low-confidence regions and have a popmax above the specified minimum allele frequency of {min_popmax_af}.")
+    # Filter to QC sites
+    last_END_positions_ht = hl.read_table(last_END_position.path)
+    qc_sites = qc_sites.key_by("locus")
```
jkgoodrich, line 65: Already done above

jkgoodrich (changes requested):
After discussion with Kristen, we think it might be best to change this around a bit. The `qc_sites` has already been written, and it will likely be faster to:

- Convert to the VDS
- Use `filter_variants` https://github.com/hail-is/hail/blob/bd3cf7c73f3fc648d03d6d393f0437364fe5fb5f/hail/python/hail/vds/methods.py#L460
- Densify

Rather than filtering to the intervals

gnomad_qc/v3/sample_qc/subpop_analysis.py
```diff
@@ -53,21 +54,30 @@ def compute_subpop_qc_mt(
         & ~release_ht.was_split
         & hl.is_snp(release_ht.alleles[0], release_ht.alleles[1])
         & hl.is_missing(lcr_intervals.ht()[release_ht.locus])
-    ).key_by("locus")
```
jkgoodrich, line 56: Ah, yes, this is needed for `filter_variants`. Since the VDS will be keyed by both. We are using only bi-allelic variants so this will work without a split of the VDS.

gnomad_qc/v3/sample_qc/subpop_analysis.py
```diff
@@ -53,21 +54,30 @@ def compute_subpop_qc_mt(
...
     qc_sites = qc_sites.checkpoint(
         get_checkpoint_path("qc_sites"),
         overwrite=args.overwrite,
         _read_if_exists=not args.overwrite,
     )
 
-    mt = densify_sites(mt, qc_sites, hl.read_table(last_END_position.path))
+    logger.info("Converting MT to VDS...")
+    mt = mt.select_entries(
+        "END", "LA", "LGT", adj=get_adj_expr(mt.LGT, mt.GQ, mt.DP, mt.LAD)
```
jkgoodrich, line 70: Just got thinking, do we need `adj`? We can add it later.

gnomad_qc/v3/sample_qc/subpop_analysis.py
```diff
@@ -53,21 +54,30 @@ def compute_subpop_qc_mt(
...
+    )
+
+    vds = hl.vds.VariantDataset.from_merged_representation(mt)
+    vds = hl.vds.split_multi(vds)
```
jkgoodrich, line 74: Do we need to do the split_multi if this is only on bi-allelics and derived from the same MT as the release HT though?

gnomad_qc/v3/sample_qc/subpop_analysis.py
klaricch: are we ever going to not want adj?

jkgoodrich: Your right, if we exclude the other entries we can't compute it later.

jkgoodrich (approved):
LGTM

PR #226, opened 2022-02-04
Update release to gcp public path

mike-w-wilson:
This path needs to be updated to the GCP version now that the requester pays has been deleted.

### DIFF
--- a/gnomad_qc/v3/resources/release.py
+++ b/gnomad_qc/v3/resources/release.py
@@ -60,7 +60,7 @@ def release_ht_path(
     if public:
         if het_nonref_patch:
             DataException("The patch HT will not be made public")
-        return f"gs://gnomad-public-requester-pays/release/{release_version}/ht/{data_type}/gnomad.{data_type}.{version_prefix}{release_version}.sites.ht"
+        return f"gs://gcp-public-data--gnomad/release/{release_version}/ht/{data_type}/gnomad.{data_type}.{version_prefix}{release_version}.sites.ht"
     else:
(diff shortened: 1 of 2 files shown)


gnomad_qc/v3/resources/release.py
```diff
@@ -60,7 +60,7 @@ def release_ht_path(
     if public:
         if het_nonref_patch:
             DataException("The patch HT will not be made public")
-        return f"gs://gnomad-public-requester-pays/release/{release_version}/ht/{data_type}/gnomad.{data_type}.{version_prefix}{release_version}.sites.ht"
+        return f"gs://gcp-public-data--gnomad/release/{release_version}/ht/{data_type}/gnomad.{data_type}.{version_prefix}{release_version}.sites.ht"
```
jkgoodrich, line 63: Just one small comment that no longer matters since this has been written. We should switch to write to `gnomad-public-requester-pays` and read from `gcp-public-data--gnomad` if it exists. This matters more for v4.

jkgoodrich: Completely feel free to ignore, but now that I see this modification, I wonder if we should pull from https://github.com/broadinstitute/gnomad_methods/blob/b95578056f8051d00dcbb749a10e585d48888f82/gnomad/resources/grch38/gnomad.py#L263 if it exists for reading. I still think for writing we should go to gs://gnomad-public-requester-pays

mike-w-wilson: Thats a good idea, I wasnt thinking about the write integration, I'll update this!

mike-w-wilson:
@jkgoodrich, I've updated the code to read from the public bucket if the table's path exists and req pays if not but I'm wondering if this should be a passed arg rather than happening silently. Is there a use case for passing this read/write param? I couldn't think of one but wanted to confirm.

jkgoodrich (approved):
LGTM

PR #231, opened 2022-02-22
add assign_subpop

klaricch:


gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -474,6 +476,8 @@ def assign_pops(
     max_mislabeled_training_samples: int = 50,  # TODO: Think about this parameter and add it to assign_population_pcs. Maybe should be a fraction? fraction per pop?
     n_pcs: int = 16,
     withhold_prop: float = None,
+    subpops: bool = False,
+    subpops_pca_scores: hl.Table = None,
```
jkgoodrich, line 480: I think we might actually want to change this. Let's add a `high_quality` option to be passed to `ancestry_pca_scores` and a `pop` with default of None. Then `pop_pca_scores_ht = ancestry_pca_scores(include_unreleasable_samples, high_quality, pop).ht()` 

Then you don't need the check for `subpops_pca_scores` and if it doesn't exist, that will be obvious from trying to get `ancestry_pca_scores`

Then all that really changes is how `training_pop` is defined

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -490,22 +494,46 @@ def assign_pops(
...
+                    hl.is_defined(pop_pca_scores_ht.project_meta.subpop_description),
+                    pop_pca_scores_ht.project_meta.subpop_description,
+                )
+                .or_missing()
             )
-            .when(project_meta_ht.v2_pop != "oth", project_meta_ht.v2_pop)
-            .or_missing()
         )
-    )
+        pop_field = "subpop_description"
```
jkgoodrich, line 520: Let's keep it simple and just change to `subpop`
```suggestion
        pop_field = "subpop"
```

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -530,11 +558,17 @@ def assign_pops(
         pop_pca_scores_ht,
         pc_cols=pop_pca_scores_ht.scores[:n_pcs],
         known_col="training_pop",
+        output_col=pop_field,
         min_prob=min_prob,
     )
 
+    # Set "<NA>" string to missing
+    pop_ht = pop_ht.annotate(
+        training_pop=hl.or_missing(pop_ht.training_pop != "<NA>", pop_ht.training_pop)
```
jkgoodrich, line 567: I'm not following why this is needed. I would have assumed before being put into `assign_population_pcs` `training_pop` would either be missing or defined to a global or subpop. Is it missing before `assign_population_pcs` and somehow set to a string in `assign_population_pcs`?

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -549,11 +583,12 @@ def assign_pops(
         pop_ht = pop_ht[pop_pca_scores_ht.key]
```
jkgoodrich, line 583: Can't add suggestion, but change logger to say "pop/subpop"

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -571,13 +606,22 @@ def assign_pops(
             pc_cols=pop_pca_scores_ht.scores[:n_pcs],
             known_col="training_pop",
             min_prob=min_prob,
+            output_col=pop_field,
         )
+
+        # Set "<NA>" string to missing
```
jkgoodrich, line 612: Same comment as above

gnomad_qc/v3/sample_qc/subpop_analysis.py
```diff
@@ -2,6 +2,9 @@
 import hail as hl
 import logging
 
+# import pandas as pd
```
jkgoodrich, line 5: Not used so remove?

gnomad_qc/v3/sample_qc/subpop_analysis.py
```diff
@@ -262,11 +264,21 @@ def main(args):  # noqa: D103
                 overwrite=args.overwrite,
             )
 
-        # TODO: Need to use annotate_subpop_meta before subpop assignment
         if args.assign_subpops:
-            raise NotImplementedError(
-                "Sub-population assignment is not currently implemented."
+            logger.info("Assigning subpops...")
+            pop_pca_scores_ht = ancestry_pca_scores(
```
jkgoodrich, line 269: see comment in `assign_pops`. Alternatively, we can just change it so that no matter what, the HT is passed in to `assign_pops`? I just don't like different functionality for the global pops vs subpops

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -474,6 +476,8 @@ def assign_pops(
     max_mislabeled_training_samples: int = 50,  # TODO: Think about this parameter and add it to assign_population_pcs. Maybe should be a fraction? fraction per pop?
     n_pcs: int = 16,
     withhold_prop: float = None,
+    subpops: bool = False,
```
jkgoodrich, line 479: Let's change n_pcs to PCs with default: `pcs=range(1, 17)` The reason is that for subpops we sometimes skip PCs, so we want to be able to specify the exact PCs we want to use

jkgoodrich (changes requested):
A few thoughts

gnomad_qc/v3/sample_qc/sample_qc.py
klaricch: I was debating those options at first, but would that break what was previously done for population assignment then?

gnomad_qc/v3/resources/sample_qc.py
```diff
@@ -166,6 +166,16 @@ def filtered_subpop_qc_mt(pop: str, version: str = CURRENT_VERSION) -> str:
...
 
 
+def assigned_subpops_path(pop: str, version: str = CURRENT_VERSION) -> str:
+    """
+    Path to Table of inferred sample subpopulations
+    :param pop: Population in which subpops have been assgined
+    :param version: Version of assigned subpops path to return
+    :return: String path to Table of assigned subpops within population specified by 'pop' parameter 
+    """
+    return f"{get_sample_qc_root(version)}/subpop_analysis/gnomad_v{version}_assigned_subpops.{pop}.ht"
```
jkgoodrich, line 176: Should we make this a VersionedTableResource?

gnomad_qc/v3/resources/sample_qc.py
```diff
@@ -166,6 +166,16 @@ def filtered_subpop_qc_mt(pop: str, version: str = CURRENT_VERSION) -> str:
     return f"{get_sample_qc_root(version)}/subpop_analysis/gnomad_v{version}_filtered_subpop_qc_mt.{pop}.mt"
 
 
+def assigned_subpops_path(pop: str, version: str = CURRENT_VERSION) -> str:
+    """
+    Path to Table of inferred sample subpopulations
```
jkgoodrich, line 171:
```suggestion
    Path to Table of inferred sample subpopulations
    
```

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -468,12 +470,29 @@ def run_pca(
     return run_pca_with_relateds(qc_mt, samples_to_drop, n_pcs=n_pcs)
 
 
+def annotate_subpop_meta(ht: hl.Table) -> hl.Table:
+    meta_ht = meta.ht()
+    ht = ht.annotate(**meta_ht[ht.key])
+    ht = ht.annotate(
+        subpop_description=ht.project_meta.subpop_description,
```
jkgoodrich, line 477: I know I did this, but looking over again, all we really need is `subpop_description` for this script right? So maybe we just do keep that and then for the plotting you might want to add a little more info if needed? Open to other thoughts though

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -468,12 +470,29 @@ def run_pca(
...
+
+    return ht
+
+
 def assign_pops(
     min_prob: float,
     include_unreleasable_samples: bool,
     max_mislabeled_training_samples: int = 50,  # TODO: Think about this parameter and add it to assign_population_pcs. Maybe should be a fraction? fraction per pop?
-    n_pcs: int = 16,
+    pcs: list = list(range(1, 17)),
```
jkgoodrich, line 492: I think we typically use the `from typing import List`
```suggestion
    pcs: List(int) = list(range(1, 17)),
```

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -488,24 +507,48 @@ def assign_pops(
     :param min_prob: Minimum RF probability for pop assignment
     :param include_unreleasable_samples: Should unreleasable samples be included in the PCA
     :param max_mislabeled_training_samples: Maximum number of training samples that can be mislabelled
-    :param n_pcs: Number of PCs to use in the RF
+    :param pcs: List of PCs to use in the RF
     :param withhold_prop: Proportion of training pop samples to withhold from training will keep all samples if `None`
-    :return: Table of pop assignments and the rf model
+    :param pop: Population that the PCA was restricted to. When set to None, the PCA on the full dataset is returned
+    :param high_quality: Whether the file includes PCA info for only high-quality samples
+    :return: Table of pop or subpop assignments and the rf model
```
jkgoodrich, line 514:
```suggestion
    :return: Table of pop or subpop assignments and the RF model
```

gnomad_qc/v3/sample_qc/subpop_analysis.py
```diff
@@ -262,11 +246,24 @@ def main(args):  # noqa: D103
...
+                withhold_prop=args.withhold_prop,
+                pop=pop,
+                high_quality=high_quality,
+            )
+
+            joint_pca_ht = joint_pca_ht.checkpoint(
+                assigned_subpops_path(pop),
+                overwrite=args.overwrite,
+                _read_if_exists=not args.overwrite,
             )
```
jkgoodrich, line 265: No used after so write instead of checkpoint
```suggestion
            joint_pca_ht.write(assigned_subpops_path(pop), overwrite=args.overwrite)
```

gnomad_qc/v3/sample_qc/subpop_analysis.py
```diff
@@ -358,9 +350,32 @@ def main(args):  # noqa: D103
...
+    )
+    parser.add_argument(
+        "--max-mislabeled-training-samples",
+        help="Maximum number of training samples that can be mislabelled",
+        type=int,
+        default=50,
+    )
+    parser.add_argument(
+        "--withhold-prop",
+        help="Proportion of training pop samples to withhold from training will keep all samples if `None`",
```
jkgoodrich, line 369:
```suggestion
        help="Proportion of training pop samples to withhold from training, all samples will be kept if this flag is not used",
```

gnomad_qc/v3/sample_qc/subpop_analysis.py
```diff
@@ -358,9 +350,32 @@ def main(args):  # noqa: D103
...
+        type=float,
+        default=None,
+    )
+    parser.add_argument(
+        "--pcs",
+        help="List of PCs to use in the RF",
```
jkgoodrich, line 375:
```suggestion
        help="List of PCs to use in the subpop RF assignment",
```

jkgoodrich (changes requested):
Just a couple suggestions, but very close!

jkgoodrich:
Also, it still seems to be failing on pylint

gnomad_qc/v3/resources/sample_qc.py
klaricch: i was going to copy what was done for "pop", but put here instead because of the extra pop argument

gnomad_qc/v3/sample_qc/subpop_analysis.py
```diff
@@ -262,11 +246,21 @@ def main(args):  # noqa: D103
...
+                include_unreleasable_samples=False,
+                max_mislabeled_training_samples=args.max_mislabeled_training_samples,  # How to decide on this number?
+                pcs=args.pcs,
+                withhold_prop=args.withhold_prop,
+                pop=pop,
+                high_quality=high_quality,
+                missing_label="Other",
             )
+
+            joint_pca_ht.write(assigned_subpops_path(pop), overwrite=args.overwrite)
```
jkgoodrich, line 262:
```suggestion
            joint_pca_ht.write(assigned_subpops(pop).path, overwrite=args.overwrite)
```

gnomad_qc/v3/sample_qc/subpop_analysis.py
```diff
@@ -10,21 +10,21 @@
     ancestry_pca_eigenvalues,
     ancestry_pca_loadings,
     ancestry_pca_scores,
+    assigned_subpops_path,
```
jkgoodrich, line 13:
```suggestion
    assigned_subpops,
```

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -468,12 +468,29 @@ def run_pca(
...
+def annotate_subpop_meta(ht: hl.Table) -> hl.Table:
+    """
+    Add subpop_description to a Table.
+
+    :param ht: Table to which metadata annotation of subpop_description should be addded
+    :return: Table with added annotation for subpop_description
+    """
+    meta_ht = meta.ht()
+    ht = ht.annotate(**meta_ht[ht.key])
+    ht = ht.annotate(subpop_description=ht.project_meta.subpop_description)
```
jkgoodrich, line 480: Sorry, one more. This is used like this below:
`pop_pca_scores_ht.project_meta.subpop_description`

So we really only need the meta_ht annotated on, I actually vote we just remove this function and do 

```
    meta_ht = meta.ht()
    ht = ht.annotate(**meta_ht[ht.key])
```

right in the function that uses `subpop_description`

```suggestion
```

jkgoodrich (changes requested):
Really, just a couple more tiny changes

jkgoodrich (approved):
LGTM!

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -472,7 +472,7 @@ def assign_pops(
     min_prob: float,
     include_unreleasable_samples: bool,
     max_mislabeled_training_samples: int = 50,  # TODO: Think about this parameter and add it to assign_population_pcs. Maybe should be a fraction? fraction per pop?
-    pcs: List(int) = list(range(1, 17)),
+    pcs: List[int] = list(range(1, 17)),
```
jkgoodrich, line 475: Ah, yes, good catch

PR #242, opened 2022-05-04
pass in pc_names param to assign_population_pcs

klaricch:


jkgoodrich (changes requested):
2 small requests. One I can't put in because it doesn't involve a line in this PR, but lets add to the `--pop_pcs` description that this is a 1-based number

jkgoodrich:
@klaricch I think this one is back to you

jkgoodrich (approved):
LGTM

PR #244, opened 2022-05-20
Fix to basic resources to split after sample filtering of UKBB samples

jkgoodrich:
We need to split after the filter sample calls if we want to use the remove_dead_alleles parameter. I also removed the test return. I know there are 0 samples removed, but I would have caught the split error on testing.

### DIFF
--- a/gnomad_qc/v4/sample_qc/hard_filters.py
+++ b/gnomad_qc/v4/sample_qc/hard_filters.py
@@ -212,6 +212,7 @@ def main(args):
         default_reference="GRCh38",
         tmp_dir="gs://gnomad-tmp-4day",
     )
+    # NOTE: remove this flag when the new shuffle method is the default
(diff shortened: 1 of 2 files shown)


gnomad_qc/v4/sample_qc/hard_filters.py
```diff
@@ -207,7 +207,13 @@ def compute_hard_filters(
 
 
 def main(args):
-    hl.init(log="/gnomad_hard_filters.log", default_reference="GRCh38")
+    hl.init(
+        log="/gnomad_hard_filters.log",
+        default_reference="GRCh38",
+        tmp_dir="gs://gnomad-tmp-4day",
+    )
+    hl._set_flags(use_new_shuffle='1')
```
klaricch, line 215: is this going to be temporary? could add comment here

klaricch (approved):

PR #248, opened 2022-07-18
Add options to change the number of computed platform PCs and the number to use for assignment

jkgoodrich:


gnomad_qc/v4/sample_qc/platform_inference.py
```diff
@@ -141,7 +141,7 @@ def main(args):
                 scores_ht = platform_pca_scores.ht()
 
             platform_ht = assign_platform_from_pcs(
-                scores_ht,
+                scores_ht.annotate(scores=scores_ht.scores[:args.n_assignment_pcs]),
```
mike-w-wilson, line 144: Nit pick for consistency: args.n_assignment_pcs could be assigned to a variable since it i used twice but I do not have strong feelings on this at all --  am unsure if we have some threshold for when arg parse variables should be assigned, whether they are referenced two time or three times. I'm fine with this merging as is though!

mike-w-wilson (approved):
LGTM, tiny nit pick but no strong feelings on it.

mike-w-wilson (approved):

PR #251, opened 2022-08-01
use only manually curated subpops in training for subpop inference

klaricch:


gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -534,6 +631,17 @@ def assign_pops(
...
+        pop_pca_scores_ht = pop_pca_scores_ht.annotate(
+            project_meta=pop_pca_scores_ht.project_meta.annotate(
+                subpop_description=hl.or_missing(
+                    hl.literal(CURATED_SUBPOPS[pop]).contains(
+                        pop_pca_scores_ht.project_meta.subpop_description
+                    ),
+                    pop_pca_scores_ht.project_meta.subpop_description,
+                )
+            )
+        )
```
jkgoodrich, line 644: Adding comment here because I can't add below, but just noticed it. Does this work if not restricting to a population? 

```
pop_pca_scores_ht = pop_pca_scores_ht.annotate(
        withheld_sample=hl.is_defined(pop_pca_scores_ht.project_meta.subpop_description)
        & (~hl.is_defined(pop_pca_scores_ht.training_pop))
    )
```

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -84,6 +84,101 @@
 Default list of subsets to include in the sample QC meta HT from the project meta HT
 """
 
+CURATED_SUBPOPS = {
```
jkgoodrich, line 87: I don't really like having these here, I think they would be better in the subpops script

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -84,6 +84,101 @@
...
+        "Orcadian",
+        "Russian",
+        "Sardinian",
+        "Swedish",
+        "Toscani",
+        "Utah Residents (European Ancestry)",
+    ],
+}
+"""
+Manually curated list of which subpops to included within each pop. Manual curation was needed as some known subpop labels may be inaccurate, especially those that are cohort-level annotations."
```
jkgoodrich, line 179:
```suggestion
Manually curated list of which subpops to include within each global pop. Manual curation was needed as some known subpop labels may be inaccurate, especially those that are cohort-level annotations."
```

jkgoodrich, line 179: Maybe add an example as well

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -534,6 +631,17 @@ def assign_pops(
     if pop is not None:
         meta_ht = meta.ht()
         pop_pca_scores_ht = pop_pca_scores_ht.annotate(**meta_ht[pop_pca_scores_ht.key])
+        # Set subpop_descritpion to missing if the subpop is not in the manually curated list of subpops for the given pop
```
jkgoodrich, line 634:
```suggestion
        # Set subpop_descriptpion to missing if the subpop is not in the manually curated list of subpops for the given pop
```

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -534,6 +631,17 @@ def assign_pops(
...
+        pop_pca_scores_ht = pop_pca_scores_ht.annotate(
+            project_meta=pop_pca_scores_ht.project_meta.annotate(
```
jkgoodrich, line 636: Is `subpop_description` all you need from the `meta_ht`?

If so, I would actually just use the project meta, so I think this line below can be moved outside the if statement

```
project_meta_ht = project_meta.ht()[pop_pca_scores_ht.key]
```

Then instead of

```
pop_pca_scores_ht = pop_pca_scores_ht.annotate(**meta_ht[pop_pca_scores_ht.key])
```
do
```
pop_pca_scores_ht = pop_pca_scores_ht.annotate(
                    subpop_description=hl.or_missing(
                    hl.literal(CURATED_SUBPOPS[pop]).contains(
                        pop_pca_scores_ht.subpop_description
                    ),
                    pop_pca_scores_ht.subpop_description,
                )
    )
```

and change uses of `pop_pca_scores_ht.project_meta.subpop_description` to `pop_pca_scores_ht.subpop_description`

jkgoodrich: I think you might want to also add the curated subpops as an optional list parameter to the function then within the subpops script pass in `curated_subpops = CURATED_SUBPOPS[pop]`. Then if passed in use it otherwise just use `pop_pca_scores_ht.subpop_description`

jkgoodrich (changes requested):

gnomad_qc/v3/sample_qc/sample_qc.py
klaricch: yeah i was back and forth on this, i'll move it to subpop_analysis.py and import here instead

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -497,6 +499,7 @@ def assign_pops(
     pcs: List[int] = list(range(1, 17)),
     withhold_prop: float = None,
     pop: str = None,
+    curated_subpops: list = [],
```
jkgoodrich, line 502: Maybe this should be set to None as a default?

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -519,34 +522,48 @@ def assign_pops(
     :param pcs: List of PCs to use in the RF
     :param withhold_prop: Proportion of training pop samples to withhold from training will keep all samples if `None`
     :param pop: Population that the PCA was restricted to. When set to None, the PCA on the full dataset is returned
+    :param curated_subpops:  List of the subpops that are accepted as known lables for the pop. If not specified, all values of subpop_description will be used as known labels when inferring subpops.
```
jkgoodrich, line 525:
```suggestion
    :param curated_subpops: List of the subpops that are accepted as known labels for the pop. If not specified, all values of subpop_description will be used as known labels when inferring subpops.
```

gnomad_qc/v3/sample_qc/sample_qc.py
```diff
@@ -558,6 +575,10 @@ def assign_pops(
                 .or_missing()
             )
         )
+        # Keep track of defined pops, useful if decide to withhold samples
```
jkgoodrich, line 578:
```suggestion
        # Keep track of defined pops, useful if samples are withheld
```

jkgoodrich (changes requested):
Just a couple small things

jkgoodrich (approved):
LGTM

jkgoodrich (approved):
LGTM, the notes are very clear and helpful

PR #252, opened 2022-08-08
Initial relatedness script for v4

lgruen:
Hopefully this is a good starting point to see whether this is going in the right direction.


gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+            ht = ht.repartition(64)
+            ht = ht.key_by(ht.i, ht.j)
+            ht.write(relatedness.path)
+
+        if args.compute_related_samples_to_drop:
+            # compute_related_samples_to_drop uses a rank table as a tie breaker when
+            # pruning samples. We determine rankings as follows: filtered samples are
+            # ranked lowest, followed by non-releasable samples, followed by presence in
+            # v3 (favoring v3 over v4), and then by coverage.
+            def annotate_project_meta_ht(
```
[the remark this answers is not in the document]
jkgoodrich: v4 will not have an `exclude`, and I don't think we need to worry about including it for v3 either since I think we mostly care about just making sure the `release` samples are prioritized. We can still keep releasable, we have that for both v3 and v4 so doesn't hurt to keep it

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+            )
+            annotated_project_meta_v4_ht = annotate_project_meta_ht(
+                project_meta_v4.ht(), sex_v4.ht, hard_filtered_samples_v4.ht()
+            )
+
+            rank_ht = annotated_project_meta_v3_ht.union(
+                annotated_project_meta_v4_ht, unify=True
+            )
+
+            rank_ht = rank_ht.order_by(
```
jkgoodrich, line 172: Need to add that we want to favor v3 `release` samples, then `v4 samples` over v3 non release samples

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+            from gnomad_qc.v3.resources.meta import project_meta as project_meta_v3
+            from gnomad_qc.v4.resources.meta import project_meta as project_meta_v4
+            from gnomad_qc.v3.resources.sample_qc import (
+                hard_filtered_samples as hard_filtered_samples_v3,
+            )
+            from gnomad_qc.v4.resources.sample_qc import (
+                hard_filtered_samples as hard_filtered_samples_v4,
+            )
+            from gnomad_qc.v3.resources.sample_qc import sex as sex_v3
+            from gnomad_qc.v4.resources.sample_qc import sex as sex_v4
```
jkgoodrich, line 159: Why here instead of at the beginning of the file?

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+                    annotated_project_meta_v3_ht[rank_ht.s]
+                ),
+                hl.desc(rank_ht.chr20_mean_dp),
+            ).add_index(name="rank")
+
+            rank_ht = rank_ht.key_by(rank_ht.s)
+            rank_ht = rank_ht.select(rank_ht.filtered, rank_ht.rank)
+
+            rank_ht.write(samples_rankings.path)
+            rank_ht = samples_rankings.ht()
```
jkgoodrich, line 185:
```suggestion
            rank_ht = rank_ht.checkpoint(samples_rankings.path, overwrite=overwrite)
```

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+            )
+
+            samples_to_drop = compute_related_samples_to_drop(
+                relatedness.ht(),
+                rank_ht,
+                args.second_degree_kin_cutoff,
+                filtered_samples=filtered_samples,
+            )
+            samples_to_drop = samples_to_drop.key_by(samples_to_drop.s)
+            samples_to_drop.write(related_samples_to_drop.path)
```
jkgoodrich, line 200:
```suggestion
            samples_to_drop.write(related_samples_to_drop.path, overwrite=overwrite)
```

gnomad_qc/v4/sample_qc/README.md
```diff
@@ -0,0 +1,89 @@
...
+gcloud services enable \
+    artifactregistry.googleapis.com \
+    batch.googleapis.com \
+    cloudbuild.googleapis.com 
+```
+
+Create an Artifact Registry repository for Docker images:
+
+```sh
+gcloud artifacts repositories create images --location=us-central1
```
jkgoodrich, line 27:
```suggestion
gcloud artifacts repositories create images --location=us-central1 --repository-format=docker
```

gnomad_qc/v4/sample_qc/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+Start an autoscaling Hail Dataproc cluster with 20 secondary workers and submit `relatedness.py --prepare-inputs`. This should take about an hour.
+
+### Run `cuKING`
+
+This step doesn't need a Dataproc cluster. Run the command printed by `relatedness.py --print-cuking-command` to submit a Cloud Batch job. This should take about two hours.
+
+### Convert outputs
+
+`cuKING` writes Parquet files. To convert these to a standard Hail table, run `relatedness.py --create-relatedness-table` on a small Dataproc cluster. This should only take a minute.
```
jkgoodrich, line 81:
```suggestion
`cuKING` writes Parquet files. To convert these to a standard Hail Table, run `relatedness.py --create-relatedness-table` on a small Dataproc cluster. This should only take a minute.
```

gnomad_qc/v4/resources/sample_qc.py
```diff
@@ -148,18 +155,35 @@ def ancestry_pca_eigenvalues(
...
+                & ((ht.kin < 0.17) | (ibs0_2 < -1e-2 * ht.kin + 2.2e-3)),
+                SECOND_DEGREE_RELATIVES,
+            )
+            .when(ht.kin < 0.15, AMBIGUOUS_RELATIONSHIP)
+            .when(ibs0_2 < -3e-3 * ht.kin + 9.2e-4, PARENT_CHILD)
+            .when(ibs0_2 > 4e-5, SIBLINGS)
+            .default(AMBIGUOUS_RELATIONSHIP),
         )
+        .default(AMBIGUOUS_RELATIONSHIP)
     )
```
jkgoodrich, line 187: I think we want to add all of this to a function probably also in `gnomad_methods` and with these options as defaults

gnomad_qc/v4/resources/sample_qc.py
```diff
@@ -5,9 +5,16 @@
...
 )
-from gnomad.sample_qc.relatedness import get_relationship_expr
+from gnomad.sample_qc.relatedness import (
+    UNRELATED,
+    SECOND_DEGREE_RELATIVES,
+    PARENT_CHILD,
+    SIBLINGS,
+    DUPLICATE_OR_TWINS,
+    AMBIGUOUS_RELATIONSHIP,
+)
```
jkgoodrich, line 15: Just a note that we order the imports alphabetically

gnomad_qc/v4/resources/sample_qc.py
```diff
@@ -333,23 +357,12 @@ def get_predetermined_qc(version: str = CURRENT_VERSION, test: bool = False):
...
     CURRENT_VERSION,
     {
         version: TableResource(
-            f"{get_sample_qc_root(version)}/gnomad.exomes.v{version}.related_samples_to_drop_for_pca.ht"
-        )
-        for version in VERSIONS
-    },
-)
-
-# Related samples to drop for release
```
jkgoodrich, line 347: We will actually still need both. One is used for the population PCA and the other is used for release and includes an option to prioritize samples that pass outlier filtering which happens after population inference

gnomad_qc/v4/resources/sample_qc.py
```diff
@@ -378,6 +391,16 @@ def get_predetermined_qc(version: str = CURRENT_VERSION, test: bool = False):
 )
 
 
+def cuking_input_path(version: str = CURRENT_VERSION, test: bool = False) -> str:
```
jkgoodrich, line 394: add docstring

gnomad_qc/v4/resources/sample_qc.py
```diff
@@ -378,6 +391,16 @@ def get_predetermined_qc(version: str = CURRENT_VERSION, test: bool = False):
...
+    # cuKING inputs can be easily regenerated, so use a temp location.
+    return f"{qc_temp_prefix(version)}cuking_input{'_test' if test else ''}.parquet"
+
+
+def cuking_output_path(version: str = CURRENT_VERSION, test: bool = False) -> str:
```
jkgoodrich, line 399: add docstring

gnomad_qc/v4/sample_qc/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+### Determine samples to drop
+
+To prune samples up to second degree relatedness, run `relatedness.py --compute-related-samples-to-drop` on a small Dataproc cluster. This should only take a few minutes.
+
+### Inferring relationships for closely related samples
+
+`get_relatedness_annotated_ht` annotates the relatedness table for closely related samples based on the kinship and IBS values computed by cuKING.
```
jkgoodrich, line 89:
```suggestion
`get_relatedness_annotated_ht` annotates the relatedness Table for closely related samples based on the kinship and IBS values computed by cuKING.
```

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+                raise DataException("number of sites doesn't match between v3 and v4")
+
+            # Combine v3 and v4 dense QC matrices, as we need to compute relatedness
+            # across the union of samples.
+            mt = mt_v3.union_cols(mt_v4)
+
+            # Compute the field required for KING.
+            mt = mt.select_entries(n_alt_alleles=mt.GT.n_alt_alleles())
+
+            # Create a table based on entries alone. By adding a row and column index,
```
jkgoodrich, line 90:
```suggestion
            # Create a Table based on entries alone. By adding a row and column index,
```

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+
+            spark = Env.spark_session()
+            df = spark.read.parquet(cuking_output_path(test=test))
+            ht = hl.Table.from_spark(df)
+            ht = ht.repartition(64)
+            ht = ht.key_by(ht.i, ht.j)
+            ht.write(relatedness.path)
+
+        if args.compute_related_samples_to_drop:
+            # compute_related_samples_to_drop uses a rank table as a tie breaker when
```
jkgoodrich, line 130:
```suggestion
            # compute_related_samples_to_drop uses a rank Table as a tie breaker when
```

gnomad_qc/v4/resources/sample_qc.py
```diff
@@ -449,23 +472,12 @@ def _import_related_samples_to_drop(**kwargs):
     },
 )
 
-# Ranking of all samples based on quality metrics. Used to remove relateds for PCA.
```
jkgoodrich, line 452: Same comment as above

jkgoodrich (changes requested):
This is great!! I'm still testing out the pipeline (still queued for the batch step), but here are my initial comments.

gnomad_qc/v4/resources/sample_qc.py
lgruen: Thanks, I'm surprised `black` doesn't do that automatically. Fixed now.

gnomad_qc/v4/resources/sample_qc.py
lgruen: I wasn't sure what the best way to change the `gnomad_methods` implementation is, given that IBS values will behave quite differently from the PC-Relate IBD estimates. Should this be a new method within `gnomad_methods`, e.g. `get_ibs_relationship_expr`, so we don't break any code that currently relies on `get_relationship_expr`? Or should we add some sort of versioning / mode to the existing `get_relationship_expr`?

gnomad_qc/v4/resources/sample_qc.py
lgruen: Thanks, reverted this. I keep forgetting that we should be able to run the PCA (even though PC-Relate would have been problematic).

gnomad_qc/v4/resources/sample_qc.py
lgruen: Done.

gnomad_qc/v4/resources/sample_qc.py
lgruen: Done.

gnomad_qc/v4/resources/sample_qc.py
lgruen: Restored.

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+                    --input_uri={cuking_input_path(test=test)} \\
+                    --output_uri={cuking_output_path(test=test)} \\
+                    --requester_pays_project=$PROJECT_ID \\
+                    --kin_threshold={args.second_degree_kin_cutoff} \\
+                    --split_factor=4 && \\
+                cd ..
+                """
+            )
+        )
+        return
```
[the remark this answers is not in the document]
lgruen: Good idea, done!

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+
+    hl.init(
+        log="/relatedness.log",
+        default_reference="GRCh38",
+    )
```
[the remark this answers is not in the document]
lgruen: The only reason not to put `hl.init` at the top of `main()` is the `--print-cuking-commmand` mode, which doesn't need Hail at all (and also doesn't need to be run on a Dataproc cluster).

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+
+            # Check that the number of sites match.
+            row_count_v3 = mt_v3.count_rows()
+            row_count_v4 = mt_v4.count_rows()
+            if row_count_v3 != row_count_v4:
+                raise DataException("number of sites doesn't match between v3 and v4")
+
+            # Combine v3 and v4 dense QC matrices, as we need to compute relatedness
+            # across the union of samples.
+            mt = mt_v3.union_cols(mt_v4)
```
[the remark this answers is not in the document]
lgruen: Oh, that's awesome! If `qc.mt()` will contain both v3 + v4, we can actually just use the [`dense_mt_to_parquet.py`](https://github.com/populationgenomics/cuKING/blob/44814da050a96659942050ead815c0f6ca983d4e/dense_mt_to_parquet.py) code directly.

I've added comments regarding the metadata and the `--test` behavior below.

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+            entries = entries.key_by()
+            entries = entries.select(
+                entries.row_idx, entries.col_idx, entries.n_alt_alleles
+            )
+
+            # Export one compressed Parquet file per partition.
+            logger.info("Writing Parquet tables...")
+            entries.to_spark().write.option("compression", "zstd").parquet(input_path)
+
+            # Write metadata that's useful for postprocessing. Map `col_idx` to `s`
```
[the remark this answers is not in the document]
lgruen: cuKING uses sample indices instead of sample strings in the input data, to reduce size. The output contains sample strings though, for making the import easier. I'll improve the comment in the code if we decide to go with calling `dense_mt_to_parquet.py` directly.

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+            col_idx_mapping.sort(key=lambda e: e.col_idx)
+            metadata = {
+                "num_sites": row_count_v4,
+                "samples": [e.s for e in col_idx_mapping],
+            }
+            with hadoop_open(f"{input_path}/metadata.json", 'w') as f:
+                json.dump(metadata, f)
+
+        if args.create_relatedness_table:
+            from hail.utils.java import Env
```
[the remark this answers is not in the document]
lgruen: No strong reason, it's just used anywhere else. I'll use the fully referenced path (`hl.utils.java.Env`) instead.

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+
+            spark = Env.spark_session()
+            df = spark.read.parquet(cuking_output_path(test=test))
+            ht = hl.Table.from_spark(df)
+            ht = ht.repartition(64)
```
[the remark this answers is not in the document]
lgruen: The number of entries in the relatedness Table depends on the kinship threshold, but it's hard to predict the exact number (which in turn influences how many partitions there should be). 64 just seemed to be an okay number for a 0.05 - 0.1 threshold, without any special reasoning. I've now made this configurable with a default of 50.

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+            ht = ht.key_by(ht.i, ht.j)
+            ht.write(relatedness.path)
+
+        if args.compute_related_samples_to_drop:
```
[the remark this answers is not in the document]
lgruen: Yes, makes sense if it stays this complicated, but maybe we can make this easier if we don't need to merge the metadata (see the comment below).

gnomad_qc/v4/sample_qc/relatedness.py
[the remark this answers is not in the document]
lgruen: A lot of the complexity here comes from merging the v3 and v4 project meta tables. If `qc.mt()` will contain both v3 + v4 samples, would it make sense to also create meta tables across both v3 and v4? I'd imagine that this would simplify any downstream analysis (including this code) quite a bit.

gnomad_qc/v4/sample_qc/relatedness.py
lgruen: Just because of the v3 + v4 renaming, which isn't used anywhere else in the code. But happy to move that to the top (if still required).

gnomad_qc/v4/sample_qc/relatedness.py
lgruen: Thanks for catching that! I'll fix that once you've confirmed whether we could potentially merge the metadata.

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -0,0 +1,251 @@
...
+
+    finally:
+        logger.info("Copying hail log to logging bucket...")
+        hl.copy_log(get_logging_path("relatedness"))
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "--test", help="Use a test MatrixTableResource as input.", action="store_true"
```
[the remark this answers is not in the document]
lgruen: Does that mean the semantics of `test` for `get_predetermined_qc` will also change? Or do you think that using the test subset for `get_predetermined_qc` isn't useful for testing this relatedness script?

I was assuming that any test dataset had already been prepared before running this script. Certainly no problem to add chromosome filtering here, but just curious what your usual strategy is to test the various stages.

lgruen:
Thanks very much for the great review feedback!

lgruen:
I've rebased to `main`, now that `jg/generate_qc_mt` has been merged.

gnomad_qc/v4/sample_qc/relatedness.py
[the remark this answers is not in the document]
jkgoodrich: I added a PR for this: https://github.com/broadinstitute/gnomad_qc/pulls

gnomad_qc/v4/sample_qc/relatedness.py
jkgoodrich: Added PR for the metadata merge: https://github.com/broadinstitute/gnomad_qc/pulls

gnomad_qc/v4/sample_qc/relatedness.py
jkgoodrich: No longer needed with the new PR I just put in

gnomad_qc/v4/resources/sample_qc.py
jkgoodrich: Yeah, I was thinking a `get_ibs_relationship_expr` in `gnomad_methods` would be reasonable

gnomad_qc/v4/sample_qc/relatedness.py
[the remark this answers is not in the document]
jkgoodrich: I'm thinking I might change the QC MT test to use the test dataset instead (can then also be used in the pop inference), so actually this will still work as is.

gnomad_qc/v4/sample_qc/relatedness.py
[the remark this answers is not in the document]
lgruen: Awesome, thanks very much! That simplifies all of this greatly.

gnomad_qc/v4/resources/sample_qc.py
lgruen: In contrast to the previous `get_relationship_expr`, the current IBS parameters for v4 unfortunately don't generalize well -- it's basically just an empirically determined set of lines to form clusters. Do you still think we should move it to `gnomad_methods` nevertheless?

lgruen:
@jkgoodrich Thanks so much for getting this merged!

> I think there may need to be an update to the script that makes the job JSON 

Interesting! This error seems new, but I guess some breaking changes had to be expected with Cloud Batch just having moved to GA less than two weeks ago.

I found the following new documentation: https://cloud.google.com/batch/docs/create-run-job#create-job-service-account

I'll send a PR to add this new section to the JSON:

```json
    "allocationPolicy": {
        "serviceAccount": {
            "email": "SERVICE_ACCOUNT_EMAIL"
        }
    }
```

lgruen:
I've just sent https://github.com/broadinstitute/gnomad_qc/pull/286.

PR #255, opened 2022-08-17
Add `n_unsplit_alleles` to gnomAD v4 subset script

jkgoodrich:


gnomad_qc/v4/subset.py
```diff
@@ -168,6 +166,19 @@ def main(args):
             metadata=HEADER_DICT,
```
mike-w-wilson, line 166: Not part of this PR but since were editting it, wondering if it makes sense to add the TODO num-vcf-shards with `parallel` option here to shard the VCF? If needs more thought -- totally get skipping it but seems like a quick todo? Although we'd probably want a tabix option too so maybe this is just another feature request/PR.

mike-w-wilson (approved):
LGTM! One note on sharding that is not necessarily tied to this PR. If you don't think its worth tackling now, I'm good merging this.

gnomad_qc/v4/subset.py
jkgoodrich: Yeah, let's make this another PR, this subset script will need more thought anyway for subsets that will have the final annotations also added on.

PR #257, opened 2022-08-19
subpop_description to training_pop

klaricch:
reannotating with subpop_description will undo the manual curation done in assign_pops, so switching to using training_pop


jkgoodrich (approved):
LGTM

PR #258, opened 2022-08-22
Fix qc mts paths typos

mike-w-wilson:
Small PR to fix two qc mt paths. The files will need to be moved before the PR is merged.


jkgoodrich (approved):
LGTM, thanks for making the change!

PR #260, opened 2022-08-29
Modify generate QC MT test to use v3 and v4 test datasets

jkgoodrich:
Currently it just filters the full datasets to the first 20 partitions


averywpx (approved):
LGTM!

PR #261, opened 2022-08-29
Add joint QC meta HT for relatedness and population inference

jkgoodrich:
Accidentally closed the other PR without actually having any of these changes

### DIFF
--- a/gnomad_qc/v4/resources/sample_qc.py
+++ b/gnomad_qc/v4/resources/sample_qc.py
@@ -214,12 +214,34 @@ qc = VersionedMatrixTableResource(
     },
 )
 
+# v3 and v4 combined sample metadata Table for relatedness and population inference
+qc_meta = VersionedTableResource(
+    CURRENT_VERSION,
+    {
+        version: TableResource(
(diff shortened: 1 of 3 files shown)


gnomad_qc/v4/sample_qc/generate_qc_mt.py
```diff
@@ -204,6 +277,9 @@ def main(args):
                 overwrite=overwrite,
             )
 
+        if args.generate_qc_meta:
+            generate_qc_meta_ht().write(qc_meta.path)
```
klaricch, line 281: could use an overwrite option here

gnomad_qc/v4/resources/sample_qc.py
```diff
@@ -214,6 +214,17 @@ def get_predetermined_qc(version: str = CURRENT_VERSION, test: bool = False):
     },
 )
 
+# v3 and v4 combined sample metadata Table for relatedness and population inference
+qc_meta = VersionedTableResource(
```
klaricch, line 218: maybe a name like "joint_qc_meta" would make it more clear that this doesn't just consist of v4 samples?

klaricch (changes requested):

klaricch (approved):
lgtm

PR #262, opened 2022-08-30
Add contamination estimate computation to hardfilters

jkgoodrich:
Wenhan, I'm not positive that I have the filtering right. I added the DP filter, but I wasn't sure if that filter ended up in the final script that you ran.


wlu04 (commented):
This looks good to me! And in terms of the DP  yes, we did have both DP and GQ filters for running this. But I think we once had this discussion and decided that DP > 10 and GQ > 20 should have a similar impact, so adding just the DP filter should be enough.

wlu04 (approved):

PR #265, opened 2022-09-13
Add callstats to v4 subset script

jkgoodrich:
Adding callstats for the subset to meet a new v4 subset request.


mike-w-wilson (approved):
One tiny thing but LGTM!

PR #269, opened 2022-09-15
Split `compute_sex` into two steps: ploidy imputation and then karyotype sex inference

jkgoodrich:
Note: This PR is stacked on #267 so #267 should be reviewed first. This is also dependent on parameterizing karyotype assignment in the `gnomad_methods`: https://github.com/broadinstitute/gnomad_methods/pull/478

This cleans up the code and also makes it easy to later re-annotate karyotype using different cutoffs if we need to.

### DIFF
--- a/gnomad_qc/v4/sample_qc/sex_inference.py
+++ b/gnomad_qc/v4/sample_qc/sex_inference.py
@@ -18,8 +18,8 @@ from gnomad_qc.v4.resources.basics import (
 )
 from gnomad_qc.v4.resources.sample_qc import (
     f_stat_sites,
-    interval_coverage,
     platform,
+    ploidy,
     sex,
     sex_imputation_coverage,
     sex_imputation_platform_coverage,
@@ -629,11 +629,14 @@ def main(args):
                 overwrite=args.overwrite,
             )
 
-        if args.impute_sex:
+        if args.impute_sex_ploidy:
             vds = get_gnomad_v4_vds(
                 remove_hard_filtered_samples=False,
                 remove_hard_filtered_samples_no_sex=True,
                 test=test,
+            )
+            platform_ht = load_platform_ht(
+                test, calling_interval_name, calling_interval_padding
             )
             if args.f_stat_ukb_var:
                 # The UK Biobank f-stat table contains only variants that were high callrate (0.99) and common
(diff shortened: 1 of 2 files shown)


gnomad_qc/v4/sample_qc/sex_inference.py
```diff
@@ -409,15 +425,11 @@ def _annotate_sex(vds, calling_intervals_ht):
...
-    if platform_ht is not None:
-        logger.info("Collecting platform list...")
-        platforms = platform_ht.aggregate(
-            hl.agg.collect_as_set(platform_ht.qc_platform)
-        )
-
-    if high_cov_intervals or high_cov_all_platforms:
+    if high_cov_intervals or high_cov_all_platforms or high_cov_per_platform:
         logger.info(
             "Running sex ploidy and sex karyotype estimation using only high coverage intervals: %d percent of samples "
```
klaricch, line 434:
```suggestion
            "Running sex ploidy estimation using only high coverage intervals: %d percent of samples "
```

gnomad_qc/v4/sample_qc/sex_inference.py
```diff
@@ -347,8 +345,24 @@ def compute_sex(
...
+    if high_cov_per_platform and high_cov_all_platforms:
+        raise ValueError(
+            "Only one of 'high_cov_per_platform' or 'high_cov_all_platforms' can be True!"
+        )
+
+    if high_cov_intervals and high_cov_all_platforms:
+        logger.warning(
+            "Both 'high_cov_intervals' and 'high_cov_all_platforms' are True, high coverage intervals will be defined"
+            " as intervals that match high coverage criteria within all platforms."
+        )
```
klaricch, line 364: is this warning no longer relevant?

klaricch (changes requested):

gnomad_qc/v4/sample_qc/sex_inference.py
jkgoodrich: It's just saying that if high_cov_intervals and high_cov_all_platforms are both set it will default to performing `high_cov_all_platforms`. It's not really relevant because of the `add_mutually_exclusive_group` below, but I think it's OK to keep

klaricch (approved):
lgtm

PR #283, opened 2022-10-17
Small runtime fix to hard filtering with sex imputation

jkgoodrich:
Need to check that sample has defined sex_karyotype to determine `sex_aneuploidy` filter


averywpx (approved):
LGTM!

PR #286, opened 2022-10-21
Add --service-account parameter to cloud_batch_submit.py invocation

lgruen:
To fix the failure described here: https://github.com/broadinstitute/gnomad_qc/pull/252#pullrequestreview-1148291734


jkgoodrich (approved):
Thanks for adding the fix @lgruen!

PR #288, opened 2022-10-26
Write a _SUCCESS file to the relatedness Parquet output directory

lgruen:
Note: requires a fairly recent Google Cloud SDK (`gcloud`) installation.

### DIFF
--- a/gnomad_qc/v4/sample_qc/relatedness.py
+++ b/gnomad_qc/v4/sample_qc/relatedness.py
@@ -57,6 +57,7 @@ def main(args):
                     --project-id=$PROJECT_ID \\
                     --tag-name=$(git describe --tags) \\
                     --service-account=cuking@$PROJECT_ID.iam.gserviceaccount.com \\
+                    --write-success-file \\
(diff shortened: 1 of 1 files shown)


jkgoodrich (approved):
This looks good to me and it worked. One thing is that we are still mostly using python 3.8. so I needed to change the `action=argparse.BooleanOptionalAction` to `action='store_true'` in `cloud_batch_submit.py`. Thank you so much for adding!!

lgruen:
> This looks good to me and it worked. One thing is that we are still mostly using python 3.8. so I needed to change the `action=argparse.BooleanOptionalAction` to `action='store_true'` in `cloud_batch_submit.py`. Thank you so much for adding!!

Great! I've changed the `action` upstream and updated the `cuKING` tag in this PR.

Generally I'd highly recommend switching to Python 3.10 (or now 3.11, which got released recently as is significantly faster), as there are lots of great new features. If Dataproc is holding you back, we've got a workaround for that too (although it's a bit of a hack).

PR #289, opened 2022-10-31
Add IBD and pc-relate to relatedness

jkgoodrich:
Also cleans the script up a bit

### DIFF
--- a/gnomad_qc/v4/sample_qc/relatedness.py
+++ b/gnomad_qc/v4/sample_qc/relatedness.py
@@ -71,7 +71,11 @@ def main(args):
         )
         return
 
-    hl.init(log="/relatedness.log", default_reference="GRCh38")
+    hl.init(
+        log="/relatedness.log",
+        default_reference="GRCh38",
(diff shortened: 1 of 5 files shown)


gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -176,26 +341,78 @@ def main(args):
...
+    cuking_args.add_argument(
         "--create-cuking-relatedness-table",
         help="Convert the cuKING outputs to a standard Hail Table.",
         action="store_true",
     )
-    parser.add_argument(
-        "--relatedness-n-partitions",
-        help="Number of desired partitions for the relatedness Table.",
-        default=50,
+    print_cuking_ibd = cuking_args.add_argument_group(
```
klaricch, line 349: is the "print" part here necessary? Looks like the args in this group are for actually running steps not just printing?

klaricch (changes requested):

klaricch (approved):
lgtm

PR #290, opened 2022-11-09
Modify sample QC resources so files are stored in subdirectories

jkgoodrich:
It was getting very difficult to navigate the sample QC directory, so this cleans it up by adding subdirectories for each step


gnomad_qc/v4/resources/sample_qc.py
```diff
@@ -654,52 +707,22 @@ def _import_related_samples_to_drop(**kwargs):
...
+######################################################################
+# Other resources
+######################################################################
+# Number of clinvar variants per sample.
+sample_clinvar_count = VersionedTableResource(
     CURRENT_VERSION,
     {
         version: TableResource(
-            f"{get_sample_qc_root(version)}/gnomad.exomes.v{version}.fingerprintcheck_failures.ht"
+            f"{get_sample_qc_root(version)}/gnomad.exomes.v{version}.clinvar.ht"
```
mike-w-wilson, line 725: Everything else, with the exception of `pre_ld_pruning_qc_variants.ht` are in subdirectories, how do you feel about making another subdirectory under gnomad/v4.0/sample_qc of `additional_resources` or some better name?

mike-w-wilson (approved):
This is beautiful and makes me so happy. I confirmed the already generated resources are sitting in their bucket locations. One small idea -- but not critical to this going in -- everything is neatly under subdirectories with the exception of `pre_ld_pruning_qc_variants.ht` and the yet to be created clinvar resource. How do you feel about one more subdirectory in `gs://gnomad/v4.0/sample_qc/` for `other_resources` or `additional_resources`? Thank you for this!

mike-w-wilson (approved):
LGTM!

PR #293, opened 2022-12-15
Fixes made for a vcf export and meta HT export

jkgoodrich:


gnomad_qc/v4/subset.py
```diff
@@ -272,6 +275,10 @@ def main(args):
                     nhomalt=ht.subset_callstats_adj.homozygote_count[1:],
                 )
             )
+            ht = ht.checkpoint(
```
mike-w-wilson, line 278: Does this help with runtime? Just curious why it's written out.

mike-w-wilson (approved):
Just one question FMK but otherwise LGTM.

gnomad_qc/v4/subset.py
jkgoodrich: Not really, but I figured it would be nice to have as a table, and I also ended up needing to rerun the export and wanted to avoid needing to recompute callstats if I needed to do it again.

PR #297, opened 2023-02-02
Generate sample QC metadata ht

jkgoodrich:
This script is not complete because I am missing some of the input, but I have summarized that in TODOs and I think it's fine to have the current script PRed and then make new PRs for the additions.

- Add annotation documentation to globals
- How to handle PCs in platform and population Tables? For example in the platform Table the number of PCs will be 9 because that is what was used, should we modify to have all 30 PCs, or add all 30 PCs to another annotation, or only keep the 9 since that is all that was used?
- Keep or drop is_female from sex HT?
- Add relatedness relationship expression code.
- Modify for any changes to get_pop_ht, ideally we change so that there is a single final pop HT that keeps track of the parameters used.
- Where to compute the release related samples to drop?
- Add more nearest neighbor info?
- Add trio info?
- Add annotation for relatedness to a v3 sample?
- joint that has v3 info?


jkgoodrich:
Fixed relatedness and ran with some paths hard coded just to get a sense of the format. that can be found here: gs://gnomad/v4.0/metadata/exomes/gnomad.exomes.v4.0.sample_qc_metadata.ht

Job ID is ```21a6ec841fb84e818aad581ea899d523``` if you want to see the logger output

Note: the related_release annotation was commented out when I ran this so it's currently missing in this HT

gnomad_qc/v4/sample_qc/hard_filters.py
```diff
@@ -340,13 +335,57 @@ def main(args):
...
+                min_af=args.qc_mt_callrate_min_af,
+                min_site_callrate=args.qc_mt_callrate_min_site_callrate,
+            )
+            ht.write(
+                get_checkpoint_path("test_gnomad.exomes.qc_mt_callrate")
+                if test
+                else sample_qc_mt_callrate.path,
+                overwrite=overwrite,
+            )
+        # add globals
```
mike-w-wilson, line 377: Is this for the on L130 and 131?

mike-w-wilson (commented):
I think this is great. I have some comments/suggestions on some of the structure but I don't think they are critical, just could help with readability. I also found the annotate_relatedness function to be pretty dense and I had to go through it slowly in order to tract what tables where being accessed and how the data was being manipulated so I think a few more inline comments could be helpful there. I'm not sure if it makes sense to break the function into separate functions though. I also gave my opinion on some of your TODOs/questions but I think discussing in person as you suggested would be the best.

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht.py
```diff
@@ -263,9 +382,6 @@ def add_annotations(
 
     :param ht: Table to annotate.
     :param ann_ht: Table with annotations to add to `ht`.
```
mike-w-wilson, line 384:
```suggestion
    :param ht: Dictionary where key is the Table name and value is the Table to annotate.
    :param ann_ht: Dictionary where the key is annotation name and the value is Table 
    with annotations to add to `ht`'s Table.
```

mike-w-wilson (approved):
This look good to me. A couple questions and suggestions but I think this is good once those are addressed! Also, just to note, my approach here was to make sure this works and makes sense. I do think it is complex and could be cleaner (basically the dictionaries are really clunky to me, both the big step and passed arg dictionary and my idea for dictionaries in the add_annotation module, but I'm not sure how much effort we want to put in to the class approach where I think in the future we'd like to move towards a version-proof repo that will likely update them and possibly use and external library like luigi for dependency management.

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht.py
```diff
@@ -215,43 +275,102 @@ def annotate_relatedness(ht: hl.Table) -> hl.Table:
...
                 )
                 for rel, rel_val in rel_dict.items()
-                for k, v in rel_set_expr_dict.items()
             }
-        )
+        ),
+        release_relatedness_filters=hl.struct(
+            **{
+                rel: get_relationship_filter_expr(
+                    ht.outlier_filtered,
```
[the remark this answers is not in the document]
jkgoodrich: good point, it should be `hard_filtered | outlier_filtered` though I think

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht.py
```diff
@@ -278,8 +394,12 @@ def add_annotations(
         tables. Default is True.
     :return: Table with additional annotations.
     """
-    logger.info("\n\nAnnotating with the %s Table.", label)
-    label = label.lower().replace(" ", "_")
+    if len(ht) != 1 or len(ann_ht) != 1:
+        raise ValueError("Both input dicts `ht` and `ann_ht` must be of length 1!")
+
+    ht_label, ht = list(ht.items())[0]
```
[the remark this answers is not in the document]
jkgoodrich: I decided to go back to the original way, but with a little bit of additional clean-up

jkgoodrich (commented):
OK, I went back to how I had it for the annotation function, but cleaned things up a little so main isn't as long. Also, fixed NA issues for hard filtering and relatedness annotations.

I will add the GATK version, but I will do that in a PR after this one goes in.

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht.py
```diff
@@ -437,35 +433,117 @@ def _sample_check(
...
+def get_hard_filter_metric_ht(base_ht: hl.Table) -> hl.Table:
+    """
+    Combine sample contamination, chr20 mean DP, and QC MT callrate into a single Table.
+
+    :param base_ht: Input Table to add annotations to.
+    :return: Table with hard filter metric annotations added.
+    """
+    logger.info("Combining hard-filter metric Tables for 'hard_filter_metrics' struct.")
+
+    # TODO: Add drop of `gq_thresholds` to the sample_chr20_mean_dp code.
```
mike-w-wilson, line 522: Post v4 -- but what are your thoughts on creating issues for all the TODOs that are going to be merged to main and adding the ticket number to the TODO? I think this could help with getting to the TODO done but obviously is not a priority, just a thought on workflow/organization.

mike-w-wilson (approved):
This looks great and it's much easier to read, I'm glad you back to the original ann function, sorry for throwing a wrench in things!

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht.py
jkgoodrich: Yeah that's a great idea, I will do that for the TODO's that we want for this script. For this specific TODO I might change this to a note that it wasn't dropped, and make the ones in the actual files sound more like a warning. I didn't want to actually change the v4 code since the current files will still have them, but wanted to warn others if they copy or modify the code later.

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht.py
```diff
@@ -60,96 +64,60 @@ def get_sex_imputation_ht() -> hl.Table:
...
+        ann_expr[f] = hl.or_else(hf_expr.contains(f), False)
+
+    # If the sample was filtered before sex imputation, it should have a missing value
+    # for the sex imputation filters.
+    for f in filters["hf_sex"]:
+        ann_expr[f] = hl.if_else(
+            hf.hf_no_sex,
+            hl.missing(hl.tbool),
+            hl.or_else(hf_expr.contains(f), False),
+            missing_false=True,
```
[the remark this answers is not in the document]
jkgoodrich: because of `hf = hf_ht[ht.key]`, it will be missing for all values in `ht` that are not in `hf_ht`

PR #298, opened 2023-02-08
Add finalize relatedness HT option that also adds a relationship annotation for each pair

jkgoodrich:
This PR will not pass tests until the related PR in gnomad_methods gets merged: https://github.com/broadinstitute/gnomad_methods/pull/511

The cuking relatedness results are here: gs://gnomad-tmp/julia/relatedness/gnomad_v4_relatedness_final.html


jkgoodrich:
I added an annotation for v3/v4 overlap to the finalized relatedness HT

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -540,6 +663,97 @@ def main(args):
...
+        type=float,
+    )
+
+    drop_related_samples = parser.add_argument_group(
+        "Compute related samples to drop",
+        "Arguments used to determine related samples that should be dropped from "
+        "the ancestry PCA or release.",
+    )
+    drop_related_samples.add_argument(
+        "--compute-related-samples-to-drop",
```
mike-w-wilson, line 744: Since this option or `release` are the two for the related group and they use some of the same resource calls, I think it would help with readability if the args at the command line were phrased similarly, 
```suggestion
        "--compute-related-samples-to-drop-for-pca",
```

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -540,6 +663,97 @@ def main(args):
...
+        help="Determine the minimal set of related samples to prune for ancestry PCA.",
+        action="store_true",
+    )
+    drop_related_samples.add_argument(
+        "--release",
```
mike-w-wilson, line 749: If the above suggestion is accepted then 
```suggestion
        "--compute-related-samples-to-drop-for-release",
```
I  think this can be reassigned to release in the script but from a command line help perspective, it helps show these are doing similar thing.

mike-w-wilson (approved):
Just a few suggestions. Overall, I think main could be shortened a bit since right now its over 500 lines. I know there is a lot going on in here and  our formatting obviously makes our scripts longer but some extra functions under the conditionals like `if args.compute_related_samples_to_drop:`  that handle some extra logic could help with readbility, though I know this means passing a ton of args to those other small functions. I don't think this needs to prevent merging just something I think would help moving forward.

gnomad_qc/v4/sample_qc/relatedness.py
```diff
@@ -130,32 +141,45 @@ def compute_rank_ht(ht: hl.Table) -> hl.Table:
...
-        hl.desc(ht.chr20_mean_dp),
-    ).add_index(name="rank")
+    rank_order = []
+    ht_select = ["rank"]
+    if filter_ht is not None:
+        ht = ht.annotate(
+            filtered=hl.coalesce(filter_ht[ht.key].outlier_filtered, False)
+        )
+        rank_order = [ht.filtered]
+        ht_select.append("filtered")
```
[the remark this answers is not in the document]
mike-w-wilson: Sorry this was a note to myself to see where it was used later and forgot to delete it after seeing it passed to compute_related _to_drop

mike-w-wilson:
Looks like black is failing -- curious if your local instance says the file is fine?

gnomad_qc/v4/sample_qc/relatedness.py
jkgoodrich: I kept this and the one below as is and changed the help for this one. release is only used if this one is also used

gnomad_qc/v4/resources/basics.py
```diff
@@ -366,3 +367,217 @@ def check_resource_existence(
...
             )
+
+
+class PipelineStepResourceCollection:
+    """A collection of resources used in a specific step of a gnomad_qc pipeline."""
+
+    def __init__(
+        self,
+        step_name: str,
+        previous_pipeline_steps: List = [],
```
mike-w-wilson, line 378: So when I first went through the class, I assumed this would contain any upstream steps , obviously didnt read the doc on L392 since you specify input, but I wondering if theres another name that would make this a bit more clear, like `previous_step_w_input` or something. Again I don't think it's critical but the variable name doesnt mention the steps output is required.

mike-w-wilson (approved):
I think this looks good. The new resource classes really help clean up the main and all the file checking that was previously being done inside the main or each module. I do think the cost of this is an added layer of fairly complex code/data structures. We discussed it briefly but I'd like to revisit this post MVP because it is fairly complex and I think there are some external libraries like luigi that could do some of this dependency checking for us. Having said that, I don't think all of our code needs to be extremely approachable and understood at first glance. Similar to the VersionedTableResources, I think this will take a little bit of getting use to by getting hands dirty. :shipit:

gnomad_qc/v4/resources/basics.py
```diff
@@ -366,3 +367,217 @@ def check_resource_existence(
                     f"rerun {step} with --overwrite. The following files already exist:"
                 ),
             )
+
+
+class PipelineStepResourceCollection:
+    """A collection of resources used in a specific step of a gnomad_qc pipeline."""
+
```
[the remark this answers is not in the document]
jkgoodrich: Added

gnomad_qc/v4/resources/basics.py
```diff
@@ -366,3 +367,217 @@ def check_resource_existence(
...
+            will be set to False.
+        :return: None.
+        """
+        if overwrite is None:
+            if self.overwrite is None:
+                overwrite = False
+            else:
+                overwrite = self.overwrite
+
+        self.pipeline_steps[step].check_resource_existance(overwrite=overwrite)
```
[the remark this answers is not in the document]
jkgoodrich: added a TODO

PR #299, opened 2023-02-09
Generate v4 release HT

mike-w-wilson:
This will not run as is as we have not yet generated the resources it is trying to reference. If you want to see the code run, I can send the notebook which uses v3 resources and fields.

I've commented in a large number of TODOs, some are essential, e.g., insilico scores ht, and a couple are more exploratory and for future releases, e.g. structuring this so selecting specific fields is simpler with versioning --right now we can re-release but it requires changing the select code if there is a new field in a resource.  

This script is based around the main CONFIG dictionary. It is currently in the middle of the script because after a code review suggestion, I removed the quotes from around the custom_select functions and if the CONFIG is at the top, this breaks python as it builds the dictionary and does not know what the custom select functions are unless they defined above it in the code, thus the CONFIG is in the middle. If I return to the old way, the CONFIG dict can move back the top.

I also need to fill out the FIELD_DESCRIPTIONS dictionary but as was discussed, this is a big PR so getting started on it without that since well need to make more changes to the code anyways. 

This does not yet have code to process/remove any freq array entries as we don't know if subsets will stay in that main array or not. I think this will be an easy add using the custom_select flexibility.


mike-w-wilson:
I'm not sure why this is failing black, my pre-commit isn't flagging anything. I also ran a new install of black 22.10 and the file remains unchanged.

mike-w-wilson:
I put the global README with annotation definitions in the create release ht script but I think this should live in some sort of resource instead. It is a ~1k line dictionary so the script looks enormous now but it is only about 400 lines once you remove this README dict.

mike-w-wilson:
I moved the dict to resources

mike-w-wilson:
To be replaced by work on branch mw/v4_exome_release

PR #301, opened 2023-02-16
Simplify related samples to drop and ranking resources

jkgoodrich:
This was in my other PR, but I think it makes sense to get it in before that one.


mike-w-wilson (approved):
LGTM

PR #307, opened 2023-02-23
For ancestry inference change exomes inferred ami to remaining

klaricch:


gnomad_qc/v4/sample_qc/assign_ancestry.py
```diff
@@ -564,6 +564,25 @@ def main(args):
...
 
+        # TODO : Parameterize a cutoff to apply to all pops and decide which
+        # filters to incorporate beforehand
+        if args.set_ami_exomes_to_remaining:
+            logger.info("Reassigning exomes inferred as 'amish' to 'remaining'...")
+            joint_meta = joint_qc_meta.ht()
+            exome_samples = hl.literal(
+                joint_meta.filter(joint_meta.data_type == "exomes").s.collect()
+            )
+            pop_ht = get_pop_ht(test=test).ht()
```
jkgoodrich, line 575: need to write this out to a temp file like of L563 since you are writing to the same path

jkgoodrich (changes requested):

gnomad_qc/v4/sample_qc/assign_ancestry.py
jkgoodrich: nvm, you just added it below

PR #318, opened 2023-03-22
Add the Versioned Table Resource for vrs_annotations

matren395:
Add Versioned Table Resource for VRS Annotations for v3 - adding to end of  gnomad_qc/v3/resources/annotations.py


klaricch (approved):
lgtm

PR #320, opened 2023-03-24
Update to use global version dictionaries

mike-w-wilson:
Updates the v3 code to use the new gnomad_method global dictionaries instead of previously used lists. This PR depends on https://github.com/broadinstitute/gnomad_methods/pull/519


jkgoodrich (changes requested):
Needs changes for SUBSETS and DOWNSAMPLINGS

PR #325, opened 2023-04-07
Remove unused final outlier QC stats

jkgoodrich:
When finalizing the outlier QC HT the globals for "qc_metrics_stats" and "lms" need to be filtered to only the stats requested


klaricch (approved):
lgtm

PR #331, opened 2023-04-24
Fix sample QC metadata HT to include hard-filter cutoffs in the globals

jkgoodrich:
I also noticed we needed additional global annotations on the outlier detection HTs so I added them to this PR as well


mike-w-wilson (approved):
:shipit:

PR #332, opened 2023-04-25
Add bi-allelic sample QC to metadata HT

jkgoodrich:
bi-allelic sample QC was used for hard-filtering, this PR includes it in the sample QC metadata HT


mike-w-wilson (approved):
LGTM

PR #333, opened 2023-04-26
Add test sample annotation to metadata HT

jkgoodrich:
We need to make sure these are included in variant QC so it's easiest to have an annotation on the metadata HT for them.

### DIFF
--- a/gnomad_qc/v4/resources/variant_qc.py
+++ b/gnomad_qc/v4/resources/variant_qc.py
@@ -18,18 +18,28 @@ from gnomad_qc.v4.resources.constants import CURRENT_VERSION, VERSIONS
 
 SYNDIP = "CHMI_CHMI3_Nex1"
 """
-String representation for syndip truth sample
+String representation for syndip truth sample.
 """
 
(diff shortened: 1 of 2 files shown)


gnomad_qc/v4/resources/variant_qc.py
```diff
@@ -16,15 +16,19 @@
...
 
-# TODO: Confirm syndip name for v4 (copied current name from UKBB).
 SYNDIP = "CHMI_CHMI3_Nex1"
 """
-String representation for syndip truth sample
+String representation for syndip truth sample.
 """
 
-NA12878 = "NA12878_1075619236"
+NA12878 = "ASC-4Set-1573S_NA12878@1075619236"
```
mike-w-wilson, line 24: gross

mike-w-wilson (approved):
:shipit:

gnomad_qc/v4/resources/variant_qc.py
jkgoodrich: very

PR #335, opened 2023-04-28
Mwqh/generate frequencies

mike-w-wilson:
This is still a WIP (needs resource collection added/subsets) but to decide our approach to the homalt fix we want to test the densify and  annotate freq and high ab strata outputs on a small chr. @jkgoodrich could you review the freq and high ab het count before we run this? Thank you!


mike-w-wilson:
I've run the script with all options, here's the tiny test output: gs://gnomad-tmp-4day/freq//gnomad.exomes.v4.0.frequencies.hom_alt_adjusted.ht . Feel free to overwrite in your own tests!

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -0,0 +1,441 @@
...
+        vds.reference_data,
+        vds.variant_data.annotate_cols(meta=meta_ht[vds.variant_data.col_key]),
+    )
+
+    if test or chrom:
+        if test:
+            logger.info("Filtering to DRD2 in test VDS for testing purposes...")
+            test_interval = [
+                hl.parse_locus_interval("chr11:113409605-113475691")
+            ]  # NOTE: test on gene DRD2 for now, will update to chr22 when original PR goes in
```
jkgoodrich, line 274: just confirming the plan is to run this script without `--test` (so it runs on the full VDS not the test dataset) and with `--chrom chr20` (or 22, whichever)?

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -0,0 +1,441 @@
...
+            vds = hl.vds.filter_intervals(vds, test_interval)
+            variants, samples = vds.variant_data.count()
+            logger.info(
+                "Test VDS has %s variants in DRD2 in %s samples...", variants, samples
+            )
+        else:
+            logger.info("Filtering to chromosome %s...")
+            vds = hl.vds.filter_chromosomes(vds, keep=f"chr{chrom}")
```
jkgoodrich, line 282: Not that it matters for this run, but I prefer having the user pass in "chr20" rather than 20

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -0,0 +1,441 @@
...
+def annotate_non_ref_het(vds: hl.vds.VariantDataset) -> hl.vds.VariantDataset:
+    """
+    Annotate non-ref heterozygous calls to prevent incorrect call adjustment post-split.
+
+    :param vds: Hail VDS to annotate non-ref het onto variant data.
+    :return: Hail VDS with _non_ref_het annotation.
+    """
+    vds.variant_data = vds.variant_data.annotate_entries(
+        _het_non_ref=vds.variant_data.LGT.is_het_non_ref()
+    )
```
jkgoodrich, line 62: Nice, I didn't realize this worked, I have always used something like:

```
vmt = vds.variant_data
vmt = vmt.annotate_entries(_het_non_ref=vmt.LGT.is_het_non_ref())
vds = hl.vds.VariantDataset(vds.reference_data, vmt)
```

good to know

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -0,0 +1,441 @@
...
+            variants, samples = vds.variant_data.count()
+            logger.info(
+                "Test VDS has %s variants in DRD2 in %s samples...", variants, samples
+            )
+        else:
+            logger.info("Filtering to chromosome %s...")
+            vds = hl.vds.filter_chromosomes(vds, keep=f"chr{chrom}")
+
+    logger.info("Annotating non_ref hets pre-split...")
+    vds = annotate_non_ref_het(vds)
```
jkgoodrich, line 285: Don't need it for now, but it might be helpful to add this as an option to `get_gnomad_v4_vds` then you don't need to do the split in this code

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -0,0 +1,441 @@
...
+    # if args.subsets:
+    #     vds = hl.vds.filter_samples(
+    #         vds,
+    #     )  # TODO: Where is subset information coming meta.project_meta.ukb is defined for non-ukb and ukd, how do we determine non-topmed v3.1 meta did (topmed=ht.s.startswith("NWD"))
+
+    logger.info("Splitting VDS....")
+    vds = hl.vds.split_multi(vds, filter_changed_loci=True)
+
+    logger.info("Densifying VDS...")
+    mt = hl.vds.to_dense_mt(vds)
```
jkgoodrich, line 296: Again, not for this run, but could also add this as an option to the `get_gnomad_v4_vds`

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -0,0 +1,441 @@
...
+
+    :param mt: Hail MatrixTable to annotate.
+    :param ab_cutoff: Alelle balance threshold at which the GT changes to homalt. Default is 0.9.
+    :return mt:
+    """
+    return (
+        (mt.AD[1] / mt.DP > ab_cutoff)
+        & mt.adj
+        & mt.GT.is_het_ref()
+        & mt.meta.project_meta.fixed_homalt_model
```
jkgoodrich, line 111: I think this might need to be 
```suggestion
        & ~mt.meta.project_meta.fixed_homalt_model
```

True means it has a fixed homalt model in the GATK version used

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -0,0 +1,441 @@
...
+    """
+    Correct frequencies at sites with an AF greater than the af_threshold.
+
+    :param ht: Hail Table containing freq and high_ab_het annotations.
+    :param af_threshold: AF threshold at which to correct frequency. Default is 0.01.
+    :return: Hail Table with adjusted frequencies.
+    """
+    mt = mt.annotate_rows(
+        ab_adjusted_freq=hl.if_else(
+            mt.freq[0].AF < af_threshold,
```
jkgoodrich, line 128: Not a big deal, but before we changed if `(af_expr > af_cutoff)` , so to be the same I think this needs to be `mt.freq[0].AF <= af_threshold`

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -0,0 +1,441 @@
...
+            mt.freq,
+            hl.map(
+                lambda f, g: hl.struct(
+                    AC=hl.int32(f.AC - g),
```
jkgoodrich, line 132: Maybe I'm totally thinking about this wrong, but don't we want to add? We are fixing a hom alt depletion, so we expect a higher AC

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -0,0 +1,441 @@
...
+                    AF=hl.if_else(f.AN > 0, (f.AC - g) / f.AN, hl.missing(hl.tfloat64)),
```
jkgoodrich, line 133: same comment as above

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -0,0 +1,441 @@
...
+                    AN=f.AN,
+                    homozygote_count=f.homozygote_count,
```
jkgoodrich, line 135: This will change as well since we are changing hets to homs

jkgoodrich (changes requested):
A few questions before this can be run

gnomad_qc/v4/annotations/generate_freq.py
mike-w-wilson: Yup, no --test. This is just for small code change tests

gnomad_qc/v4/annotations/generate_freq.py
mike-w-wilson: Good call, added TODOs

gnomad_qc/v4/annotations/generate_freq.py
mike-w-wilson: That does make sense!

gnomad_qc/v4/annotations/generate_freq.py
mike-w-wilson: Fixed!

gnomad_qc/v4/annotations/generate_freq.py
mike-w-wilson: Nope, you are correct

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -372,7 +370,6 @@ def main(args):  # noqa: D103
     parser.add_argument(
         "--chrom",
         help="If passed, script will only run on passed chromosome.",
```
jkgoodrich, line 372:
```suggestion
        help="If passed, script will only run on passed chromosome.",
        type=str,
```

jkgoodrich (approved):
only one small thing

mike-w-wilson:
This is an old branch, closing. New branch is mw/v4_freq

PR #337, opened 2023-05-02
Add rsid to SV hist output

mike-w-wilson:
The SV team repurposes the rsid field to be a unique identifier. This adds the field into the output for use by the browser team.

### DIFF
--- a/gnomad_qc/v3/annotations/generate_sv_hists.py
+++ b/gnomad_qc/v3/annotations/generate_sv_hists.py
@@ -36,6 +36,7 @@ def generate_hists(mt: hl.MatrixTable) -> hl.Table:
     # Remove multiallelic CNV sites as recommended by the SV team
     # (too difficult to determine carriers vs non carriers at these sites)
     hists = mt.select_rows(
+        rsid=mt.rsid,
         age_hist_het=hl.or_missing(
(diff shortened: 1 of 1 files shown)


mike-w-wilson:
Screenshot 2023-05-03 at 8 44 10 AM
This is a unique fields

gnomad_qc/v3/annotations/generate_sv_hists.py
```diff
@@ -137,6 +138,7 @@ def get_sex_and_autosome_mt() -> hl.MatrixTable:
         reference_genome="GRCh38",
         min_partitions=300,
         force_bgz=True,
+        array_elements_required=False,
```
ch-kr, line 141: why is this required now (and not previously)?

ch-kr (approved):
one small question about the `array_elements_required` param, otherwise LGTM!

gnomad_qc/v3/annotations/generate_sv_hists.py
mike-w-wilson: Sorry -- this was a runtime error, didnt come up in my small test dataset but there was a missing value in the larger full dataset, and I fixed it locally for the original run.

PR #338, opened 2023-05-04
Add use of gnomad_methods `annotate_allele_info` to add allele info into the info HT during splitting alleles

jkgoodrich:


gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -36,17 +37,23 @@ def split_info(info_ht: hl.Table) -> hl.Table:
     """
     Generate an info Table with split multi-allelic sites from the multi-allelic info Table.
 
+    .. note::
+
+        The split of multi-allelic sites before splitting the 'info' annotation is done
+        within `annotate_allele_info` in gnomad_methods so that allele info is also
+        annotated on the returned Table.
```
mike-w-wilson, line 44:
```suggestion
        gnomad_methods' `annotate_allele_info` splits multi-allelic sites before the `info` 
        annotation is split to ensure that the returned Table is annotated with allele info.
```

mike-w-wilson (approved):
One suggestion on the description but no strong feelings!

PR #339, opened 2023-05-04
Add v4 VEP to variant annotations

jkgoodrich:


gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -171,6 +185,15 @@ def main(args):
     parser.add_argument(
         "--export-info-vcf", help="Export info as VCF.", action="store_true"
     )
+    parser.add_argument(
+        "--run-vep", help="Generates vep annotations.", action="store_true"
+    )
+    parser.add_argument(
+        "--vep-version",
+        help="Version of VEPed context Table to use in vep_or_lookup_vep.",
+        action="store_true",
```
mike-w-wilson, line 194:
```suggestion
```
I dont think this will work as written. My understanding is vep-version will be stored as `True` if you pass `--vep-version` and it will consider the user passed version a new argument. I think it can be dropped and then 105 is always passed unless overwritten by the arg

mike-w-wilson (changes requested):
One change to the argparse then this looks good!

gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
jkgoodrich: oops, just a copy paste error. Thank you!

mike-w-wilson (approved):
:shipit:

PR #352, opened 2023-05-18
change CURRENT RELEASE to 3.1.4

KoalaQin:
change CURRENT RELEASE to 3.1.4, add 3.1.3 and 3.1.4 to RELEASES. 
Not sure if this will influence other codes.


jkgoodrich (changes requested):
Just a small change

KoalaQin:
> 
Actually KC wants to discuss this, if we put a note on each release in the code on the major changes or we give a link to the blogpost.

jkgoodrich:
I think we need this specific note in the code no matter what. You can make another ticket and accompanying PR for deciding on and implementing any additional documentation that should be added to describe the release changes.

jkgoodrich:
I went through the code where `CURRENT_RELEASE` is used and found some errors I had made in the initial change. I'm going to put in a PR on top of this branch for those changes.

KoalaQin:
> 

I was worried about that. Could you show me how you detect errors when change like this happen?

jkgoodrich (approved):

PR #353, opened 2023-05-19
Fixes to use `CURRENT_VERSION` instead of `CURRENT_RELEASE` in a couple places

jkgoodrich:


gnomad_qc/v3/resources/meta.py
```diff
@@ -8,33 +8,28 @@
...
 
-    :param version: gnomAD release version
-    :param meta_version: metadata version to return
+    :param version: gnomAD version
     :return: String path to the finalized metadata
     """
-    return (
-        f"{_meta_root_path(version)}/gnomad_v{version}_metadata_v{meta_version}.tsv.gz"
-    )
+    return f"{_meta_root_path(version)}/gnomad_v{version}_metadata.tsv.gz"
```
KoalaQin, line 32: I found that there's only gs://gnomad/metadata/genomes_v3/gnomad_v3_metadata_2019-09-27.tsv.gz, nothing inside gs://gnomad/metadata/genomes_v3.1/

jkgoodrich: yeah, we never wrote it out to a tsv for v3.1, we don't really need it for anything unless someone asks us for it

KoalaQin:
LGTM! Let's merge it.

PR #358, opened 2023-06-05
add validity step for VEP after running VEP

KoalaQin:
put the validity check step (count if variants are correctly annotated as in "protein-cding" biotype) right after run_vep.

PR #359, opened 2023-06-07
module to remove missing fields from VEP105 HT

KoalaQin:
Add one module to remove missing fields from VEP105 annotation HT.
Made some minor changes in annotation resources to include both exomes and genomes.

### DIFF
--- a/gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
+++ b/gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
@@ -50,3 +50,4 @@ def remove_missing_vep_fields(vep_ht: hl.Table) -> hl.Table:
                 }
             )
         )
+    return vep_ht
--- a/gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
+++ b/gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
@@ -4,11 +4,10 @@ import hail as hl
 
 def remove_missing_vep_fields(vep_expr: hl.StructExpression) -> hl.StructExpression:
     """
-    Remove fields from VEP 105 annotations that are missing in all rows.
+    Remove fields from VEP 105 annotations that have been excluded in past releases or are missing in all rows.
 
     :param vep_expr: StructExpression containing VEP 105 annotations.
     :return: StructExpression containing VEP 105 annotations with missing fields removed.
-    ...Note: "colocated_variants" is not all missing, but it has been removed in previous releases.
     """
     vep_expr = vep_expr.drop("colocated_variants", "context")
 
--- a/gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
+++ b/gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
@@ -5,6 +5,7 @@ import logging
 import hail as hl
 
 from gnomad_qc.v4.resources.annotations import get_vep
+from gnomad_qc.v4.resources.constants import CURRENT_VERSION
 
 logging.basicConfig(
     format="%(asctime)s (%(name)s %(lineno)s): %(message)s",
@@ -23,7 +24,7 @@ def remove_missing_vep_fields(vep_ht: hl.Table) -> hl.Table:
     """
     logger.info("Loading annotation tables...")
 
-    vep_ht = get_vep(data_type="genomes").ht()
+    vep_ht = get_vep(version=CURRENT_VERSION, data_type="genomes").ht()
 
     vep_ht = vep_ht.annotate(vep=vep_ht.vep.drop("colocated_variants", "context"))
 
--- a/gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
+++ b/gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
@@ -1,0 +1,44 @@
+"""script to create release sites ht for v4 genomes."""
+import argparse
+import logging
+
+import hail as hl
+
+from gnomad_qc.v4.resources.annotations import get_vep
+
+
+def add_release_annotations(freq_ht: hl.Table) -> hl.Table:
+    """
+    Load and join all Tables with variant annotations.
+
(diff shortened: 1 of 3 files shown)


mike-w-wilson:
@KoalaQin this doesn't look like its comparing to the current main. I noticed when looking at the annotations root adding in data_type. Could you pull those changes in? Looks like they all went in #358 so there is a merge conflict right now. I just want to be sure I'm comparing again the up to date version.

KoalaQin:
Hi Mike, this was pushed before Julia approved my previous PR. Sorry about the confusion. Can you see the correct version now?

mike-w-wilson:
Yes thank you! Looks like the diff is just single function now with a minor edit to another file

gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
```diff
@@ -0,0 +1,50 @@
...
+        vep_ht = vep_ht.annotate(
+            vep=vep_ht.vep.annotate(
+                **{
+                    consequence: vep_ht.vep[consequence].map(
+                        lambda x: x.drop("minimised")
+                    )
+                }
+            )
+        )
+    return vep_ht
```
mike-w-wilson, line 50: A suggestion for readability and code reuse,  since you aren't doing anything to the actual ht, you could just pass the ht vep expression, vep_expr, to the entire function and return that instead of the entire table, I think that would read the best.   
```suggestion
    vep_expr = vep_expr.drop("colocated_variants", "context")

    vep_expr = vep_expr.annotate(
            transcript_consequences=vep_expr.transcript_consequences.map(
                lambda x: x.drop("minimised", "swissprot", "trembl", "uniparc")
            )

    for consequence in [
        "intergenic_consequences",
        "motif_feature_consequences",
        "regulatory_feature_consequences",
    ]:
        vep_expr = vep_expr.annotate(
                **{
                    consequence: vep_expr[consequence].map(
                        lambda x: x.drop("minimised")
                    )
                }
            )
    return vep_expr
```

mike-w-wilson: I made this suggestion here so havent had a chance to format it, it may fail black if you accept it. It will also require updates to the docs.

mike-w-wilson (commented):
I have a suggestion on the type passed, if you accept it, I'll can do another quick pass and then this is good to go

gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
```diff
@@ -0,0 +1,42 @@
+"""Script to create release sites HT for v4 genomes."""
+import argparse
+import logging
+
+import hail as hl
+from gnomad.resources.grch38.reference_data import dbsnp
+
+from gnomad_qc.v4.resources.annotations import get_vep
+from gnomad_qc.v4.resources.constants import CURRENT_VERSION
```
mike-w-wilson, line 9: I know this will eventually be a complete script but lets not add the imports until they are used. This goes for argparse  and logging too.

gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
```diff
@@ -0,0 +1,42 @@
...
+
+    :param vep_expr: StructExpression containing VEP 105 annotations.
+    :return: StructExpression containing VEP 105 annotations with missing fields removed.
+    """
+    vep_expr = vep_expr.drop("colocated_variants", "context")
+
+    vep_expr = vep_expr.annotate(
+        transcript_consequences=vep_expr.transcript_consequences.map(
+            lambda x: x.drop("minimised", "swissprot", "trembl", "uniparc")
+        )
```
mike-w-wilson, line 31: This is purely preference here since its a small module but I generally like to define lists like this outside of the operation as I find it easier to locate when we add/remove items.

gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
```diff
@@ -0,0 +1,42 @@
...
+    format="%(asctime)s (%(name)s %(lineno)s): %(message)s",
+    datefmt="%m/%d/%Y %I:%M:%S %p",
+)
+logger = logging.getLogger("create_release_ht")
+logger.setLevel(logging.INFO)
+
+
+def remove_missing_vep_fields(vep_expr: hl.StructExpression) -> hl.StructExpression:
+    """
+    Remove fields from VEP 105 annotations that are missing in all rows.
```
mike-w-wilson, line 21: are some of these fields just not wanted or are they all missing?

mike-w-wilson (commented):
One thing with imports, a question, and a general comment!

gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
KoalaQin: Hopefully next time, we will have the vep.config correct and we don't have to remove any fields.  But I can make two lists: fields_missing_in_all and fields_missing_in_transcript_consequences, still inside the function, no?

KoalaQin: wait, there will be a 3rd list too, like context in the first layer.

mike-w-wilson: Yeah, I went back and forth on this comment. I think having the three lists may make it less readable so lets keep it as is especially since this will only be used once

gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py
mike-w-wilson: Gtocha, I just wanted to comment to reflect the decision accurately. I think this is okay. We could add ` or excluded in the last release` but I dont think it's necessary here. As long as the decision was covered in zenhub or the notebook I think we are good!

mike-w-wilson (approved):
LGTM

PR #365, opened 2023-07-07
function to merge all CADD predictions to a HT

KoalaQin:
The function merged all downloadable and gnomAD historical CADD predictions (for v3) and recent run for v4 into one single HT.


gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -0,0 +1,117 @@
...
+        - gnomad 3.0 indels: `cadd.v1.6.indels.tsv.bgz` (1.1G) downloaded from `https://krishna.gs.washington.edu/download/CADD/v1.6/GRCh38/gnomad.genomes.r3.0.indel.tsv.gz`. It contains 100,546,109 indels in gnomaD v3.0.
+        - gnomad 3.1 indels: `CADD-indels-gnomad.3.1.ht` was run by William Phu with CADD v1.6 in 2020. It contains 166,122,720 indels in gnomAD v3.1.
+        - gnomad 3.1 complex indels: `CADD-1.6-gnomad-complex-variants.ht` was run by William Phu with CADD v1.6 in 2020. It contains 2,307 complex variants that were not filtered out by Hail is.indels().
+        - gnomAD v4 indels: `cadd.v1.6.gnomAD_v4_new_indels.tsv.bgz` (368M) was run by Qin He with CADD v1.6 in 2023. It contains 32,561,253 indel that are new in gnomAD v4.
+        - 1,972,208 indels were duplicated in gnomAD v3.0 and v4.0, in gnomAD v3.1 and v4.0. However, CADD only generate a score per loci. We use anti_join to keep only the latest prediction.
+    :return: Hail Table with CADD scores for GRCh38.
```
mike-w-wilson, line 30:
```suggestion
    Create a Hail Table with CADD scores for GRCh38.

    The combined CADD scores in the returned table are from the following sources:
        - all SNVs: `cadd.v1.6.whole_genome_SNVs.tsv.bgz` (81G) downloaded from `https://krishna.gs.washington.edu/download/CADD/v1.6/GRCh38/whole_genome_SNVs.tsv.gz`. It contains 8,812,917,339 SNVs.
        - gnomad 3.0 indels: `cadd.v1.6.indels.tsv.bgz` (1.1G) downloaded from `https://krishna.gs.washington.edu/download/CADD/v1.6/GRCh38/gnomad.genomes.r3.0.indel.tsv.gz`. It contains 100,546,109 indels from gnomaD v3.0.
        - gnomad 3.1 indels: `CADD-indels-gnomad.3.1.ht` was run on gnomAD v3.1 with CADD v1.6 in 2020. It contains 166,122,720 indels from gnomAD v3.1.
        - gnomad 3.1 complex indels: `CADD-1.6-gnomad-complex-variants.ht` was run on gnomAD v3.1 with CADD v1.6 in 2020. It contains 2,307 complex variants that do not fit Hail's criteria for an indel and thus exist in a separate table than the gnomad 3.1 indels.
        - gnomAD v4 indels: `cadd.v1.6.gnomAD_v4_new_indels.tsv.bgz` (368M) was run on gnomAD v4 with CADD v1.6 in 2023. It contains 32,561,253 indels that are new in gnomAD v4.
        
         .. note:: 
         1,972,208 indels were duplicated in gnomAD v3.0 and v4.0 or in gnomAD v3.1 and v4.0. However, CADD only generates a score per loci. We keep only the latest prediction, v4.0, for these loci.
    :return: Hail Table with CADD scores for GRCh38.
```

Just a suggestion for the way this is organized and for some descriptions -- feel free to keep or edit it some more!

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+
+    # Merge the CADD predictions run for v3 versions.
+    indel3 = indel3_0.union(indel3_1, indel3_1_complex)
+
+    # This will avoid duplicated indels in gnomAD v3 and v4.
+    indel3 = indel3.anti_join(indel4)
+
+    ht = snvs.union(indel3, indel4)
+    return ht
```
mike-w-wilson, line 85: Can you add a cadd version to as a global here which would eventually be put on the final in silico HT? This is something we'll want on the final release HT for all in silico versions.

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -0,0 +1,117 @@
...
+
+
+def main(args):
+    """Generate Hail Tables with in silico predictors."""
+    hl.init(
+        default_reference="GRCh38",
+        log="/insilico_predictors.log",
+        tmp_dir="gs://gnomad-tmp-4day",
+    )
+
```
mike-w-wilson, line 95: Just curious how you plan to merge the  other tables that will be in this script? Will you have an in_silico HT that grabs the sites from the release and then annotate all of these predictors onto it?

mike-w-wilson (commented):
This is a general thought on this script and an approach but depending on how many tools you have, it may be worth while to set up a dictionary config with variables that are used in more general scripts and then run a single join ht script that could accept the keys to that dictionary. This could be very similar to https://github.com/broadinstitute/gnomad_qc/pull/299/files which just makes it easy to add new tools. I don't have strong opinion here but I think it could save you a lot of time if you adopt something similar as you add in new tools since the main logic is just joins.

KoalaQin:
> This is a general thought on this script and an approach but depending on how many tools you have, it may be worth while to set up a dictionary config with variables that are used in more general scripts and then run a single join ht script that could accept the keys to that dictionary. This could be very similar to https://github.com/broadinstitute/gnomad_qc/pull/299/files which just makes it easy to add new tools. I don't have strong opinion here but I think it could save you a lot of time if you adopt something similar as you add in new tools since the main logic is just joins.

We will have 4-5 tools, but each is a bit different, here is just to combine all the data we had for CADD.

gnomad_qc/v4/annotations/insilico_predictors.py
KoalaQin: I was planning to have a HT for each in silico predictor in this script, also a way to document what we did for each of them. CADD is the simplest. Then we'll merge all of them like you did before to `create_releate_sites_ht` for exomes and genomes respectively. What do you think about this?

KoalaQin:
I wanted to separate the data preparation from making dictionaries. We will do something similar later on to make them a dictionaries. Right now we are still deciding which strategy we will apply on the other scores.

gnomad_qc/v4/annotations/insilico_predictors.py
mike-w-wilson: I think this makes sense. I think the merge should live in this script though not in the release script so that we end up with a single insilico_predictors.ht which contains structs of the info we are pulling from each in silico predictor
```
locus
alleles
cadd: hl.struct{
    raw_score,
    phred,
    }
pangolin: hl.struct{
....
 ```
 Is that what youre envisioning?

File: gnomad_qc/v4/create_release/create_release_sites_ht_genomes.py [hunk 4]
```diff
@@ -1,5 +1,22 @@
 """Script to create release sites HT for v4 genomes."""
+
```
mike-w-wilson, line 2: I think a commit slipped in here. Or am I missing where the above output is used? Since this is a PR to merge CADD, I think any release changes are out of scope.

KoalaQin: Sorry, that was my on-going script. I removed it from this PR.

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -0,0 +1,124 @@
...
+    if args.cadd:
+        logger.info("Creating CADD Hail Table for GRCh38...")
+
+        ht = create_cadd_grch38_ht()
+        ht.checkpoint(
```
mike-w-wilson, line 104: As we discussed in person, this can be changed back to a write.

mike-w-wilson (approved):
Just one thing and then this looks good to me since we decided that this script is not going to merge the tables, just create them. 

In general, we should be reviewing any code that is creating files for production before it is run so please make sure to have the next predictor's create script reviewed before creating the file. You should build in a test case so that you can test the full code without running it on such large datasets. We do typically this with filter_partitions or filter_intervals.

gnomad_qc/v4/annotations/insilico_predictors.py
KoalaQin: I changed it back to `write`, but it won't have `_read_if_exists` anymore, does it matter?

KoalaQin:
I'm thinking where to apply the test, for each predictor, we will have multiple files to be merged into one, should we use `filter_intervals` just before write?

PR #367, opened 2023-07-12
Add function to compute AS_QUALapprox from LPL

jkgoodrich:


gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -42,6 +42,125 @@
...
+        Starting Values:
+            - alleles: [‘G’, ‘*’, ‘A’, ‘C’, ‘GCTT’, ‘GT’, ‘T’]
+            - LGT: 1/2
+            - LA: [0, 1, 6]
+            - LPL: [138, 98, 154, 26, 0, 14]
+            - QUALapprox: 138
+
+        Use `extract_as_pls` to get PLs for each allele:
+            - allele 1: [138, 98, 154] -> [40, 0, 56]
+            - allele 2: [138, 26, 14] -> [124, 12, 0]
```
mike-w-wilson, line 102: Maybe drop from `-> [....]`   in each allele example since you normalize below?

gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -42,6 +42,125 @@
...
+
+        Normalize PLs by subtracting the smallest element from all the PLs:
+            - allele 1: [138-98, 98-98, 154-98] -> [40, 0, 56]
+            - allele 2: [138-14, 26-14, 14-14] -> [124, 12, 0]
+
+        Use the first element of the allele specific PLs to generate AS_QUALapprox:
+        [None, 40, 124]
+
+        Correct for star allele: [None, 0, 124]
```
mike-w-wilson, line 111: You mention above but maybe reiterate or move the note that star alleles get set to 0.

gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -42,6 +42,125 @@
...
+    """
+    Run compute info on a MatrixTable.
+
+    ..note::
+
+        Adds a fix for AS_QUALapprox by recomputing from LPL because some were found to
+        have different lengths than LA.
+
+    :param mt: Input MatrixTable.
+    :param test: Whether to use the test dataset.
```
mike-w-wilson, line 142:
```suggestion
    :param test: Whether to use the test dataset. Default to False.
```

mike-w-wilson (commented):
Just a couple of suggestions. I do think `recompute_as_qualapprox_from_lpl` and `extract_as_pls` could move to gnomad_methods. I know we just hit this issue in v4 but it may be helpful for others if their gvcfs have the same missing data?

gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -42,6 +42,125 @@
...
+    [0/0, 0/1, 1/1, 0/2, 1/2, 2/2].
+
+    If we want to extract the PLs for each alternate allele, we need to extract:
+        - allele 1: [0/0, 0/1, 1/1]
+        - allele 2: [0/0, 0/2, 2/2]
+
+    Example:
+        - LPL: [138, 98, 154, 26, 0, 14]
+        - Extract allele 1 PLs: [0/0, 0/1, 1/1] -> [138, 98, 154]
+        - Extract allele 2 PLs: [0/0, 0/2, 2/2] -> [138, 26, 14]
```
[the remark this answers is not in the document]
jkgoodrich: I'm not positive what you mean, the reference with allele 1 is [0/0, 0/1, 1/1], ref with allele 2 is [0/0, 0/2, 2/2], 1/2 would require both alt alleles, so it doesn't matter in this case

jkgoodrich:
> I know we just hit this issue in v4 but it may be helpful for others if their gvcfs have the same missing data?

Maybe, but for now I think I feel more comfortable keeping it in gnomad_qc as v4 specific since that is what I tested on. I think I would want to do a bit more before making any claims of it being more general.

mike-w-wilson (approved):
LGTM

PR #368, opened 2023-07-12
Add new outlier detection annotation and elgh2_project descriptions

jkgoodrich:
Added the JSON just for reviewing, will remove it before merge


klaricch (approved):
lgtm

PR #376, opened 2023-07-20
v4 frequency script

mike-w-wilson:
Here is the full frequency script. I've run it on the test dataset using the `--use-test-dataset` arg. The script currently puts out all of the annotations from the split VDSs and any annotation used in correcting for the high AB hets. I plan on removing these once this is ready and have a note before the final write on which fields I plan on removing for now. I also realized this does not account for the downsamplings within the ukb subset which Konrad had asked for so I will work on adding that but figured this is huge so should not hold up review with it. As I mentioned in one of our chats, I copied Tim's freq updates and then updated them here to account for the upstream downsampling annotation. Post v4, I will rewrite the gnomad_methods annotate_freq method.

### DIFF
--- a/gnomad_qc/v4/annotations/generate_freq.py
+++ b/gnomad_qc/v4/annotations/generate_freq.py
@@ -408,19 +408,19 @@ def correct_qual_hists(ht: hl.Table) -> hl.Table:  # add ab_threshold as arg
     return ht
 
 
-def generate_faf_popmax(ht: hl.Table) -> hl.Table:
+def generate_faf_grpmax(ht: hl.Table) -> hl.Table:
     """
-    Compute filtering allele frequencies and popmax with the AB-adjusted frequencies.
+    Compute filtering allele frequencies and grpmax with the AB-adjusted frequencies.
 
     :param ht: Hail Table containing freq, ab_adjusted_freq, high_ab_het annotations.
-    :return: Hail Table with faf & popmax annotations.
+    :return: Hail Table with faf & grpmax annotations.
     """
     faf, faf_meta = faf_expr(
         ht.ab_adjusted_freq, ht.freq_meta, ht.locus, POPS_TO_REMOVE_FOR_POPMAX
     )
     ht = ht.annotate(
         faf=faf,
-        popmax=pop_max_expr(  # TODO: Update popmax to grpmax
+        grpmax=pop_max_expr(
             ht.ab_adjusted_freq, ht.freq_meta, POPS_TO_REMOVE_FOR_POPMAX
         ),
     )
@@ -429,9 +429,9 @@ def generate_faf_popmax(ht: hl.Table) -> hl.Table:
         faf_index_dict=make_faf_index_dict(faf_meta, label_delimiter="-"),
     )
     ht = ht.annotate(
-        popmax=ht.popmax.annotate(  # TODO: Update popmax to grpmax
+        grpmax=ht.grpmax.annotate(
             faf95=ht.faf[
-                ht.faf_meta.index(lambda x: x.values() == ["adj", ht.popmax.pop])
+                ht.faf_meta.index(lambda x: x.values() == ["adj", ht.grpmax.pop])
             ].faf95
         )
     )
@@ -712,8 +712,8 @@ def main(args):  # noqa: D103
         logger.info("Correcting age histograms...")
         ht = correct_age_hists(ht)
 
-        logger.info("computing FAF & popmax...")
-        ht = generate_faf_popmax(ht)
+        logger.info("computing FAF & grpmax...")
+        ht = generate_faf_grpmax(ht)
 
         logger.info("Calculating InbreedingCoeff...")
         ht = compute_inbreeding_coeff(ht)
--- a/gnomad_qc/v4/annotations/generate_freq.py
+++ b/gnomad_qc/v4/annotations/generate_freq.py
@@ -920,6 +920,79 @@ def generate_freq_and_hists_ht(
     return freq_ht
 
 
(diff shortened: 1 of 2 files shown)


gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -0,0 +1,1307 @@
...
+
+
+def compute_age_hist(mt: hl.MatrixTable) -> hl.MatrixTable:
+    """
+    Compute age histograms for each variant.
+
+    :param mt: Input MT with age annotation.
+    :return: MatrixTable with age histogram annotations.
+    """
+    mt = mt.annotate_rows(**age_hists_expr(mt.adj, mt.GT, mt.age))
```
mike-w-wilson, line 325:
```suggestion
    mt = mt.annotate_rows(**age_hists_expr(mt.adj, mt.GT, mt.age))
```
In v2, we created an age hist per subset. I did not include the logic for this with the non-ukb subset so it needs to be added in. We can either split the age bins in the freq aggs to be  ukb-age-bin and non-ukb-age-bin  or try to determine these numbers from the current ukb and age bin freq counts. I need to think if the second approach accurately captures the age bins in the non-ukb subset.

mike-w-wilson:
I'm realizing the non-ukb subset needs to have its own grpmax and age hists and this code doesnt currently account for that. Since we are already splitting by ukb, this should be easy enough to add in but I wont until we get through one round of review.

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -772,15 +762,24 @@ def generate_faf_grpmax(ht: hl.Table) -> hl.Table:
...
+    faf_meta_expr[1] = faf_meta_expr[1].map(
+        lambda d: hl.dict(d.items().append(("subset", "non_ukb")))
+    )
     ht = ht.annotate_globals(
-        faf_meta=faf_meta_expr,
+        faf_meta=hl.flatten(faf_meta_expr),
         faf_index_dict=[
             make_faf_index_dict(hl.eval(x), label_delimiter="-") for x in faf_meta_expr
         ],
+        grp_max_meta=[
```
jkgoodrich, line 778:
```suggestion
        grpmax_meta=[
```

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -772,15 +762,24 @@ def generate_faf_grpmax(ht: hl.Table) -> hl.Table:
...
+            {"dataset": "gnomad"},
+            {"dataset": "non_ukb"},
+        ],  # TODO: These seem silly but keeps with the meta/dict theme of globals
+        grp_max_index_dict=SUBSET_DICT,
```
jkgoodrich, line 782:
```suggestion
        grpmax_index_dict=SUBSET_DICT,
```

gnomad_qc/v4/annotations/generate_freq.py
jkgoodrich, line 778: I didn't follow all the discussions on this so pinging @ch-kr to check this (L778-L782).

jkgoodrich (changes requested):
just a small change, and pinging KC to look at a few lines

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -0,0 +1,1027 @@
...
+    ht = ht.annotate_globals(
+        faf_meta=hl.flatten(faf_meta_expr),
+        faf_index_dict=[
+            make_faf_index_dict(hl.eval(x), label_delimiter="-") for x in faf_meta_expr
+        ],
+        grpmax_meta=[
+            {"dataset": "gnomad"},
+            {"dataset": "non_ukb"},
+        ],  # TODO: These seem silly but keeps with the meta/dict theme of globals
+        grpmax_index_dict=SUBSET_DICT,
```
mike-w-wilson, line 782: Sorry, I commited a suggestion on name and it looks like it removed @jkgoodrich 's other comment. @ch-kr do you have thoughts on the grpmax_meta and index_dict? L778-782

jkgoodrich (approved):
Approved pending KC's review

gnomad_qc/v4/annotations/generate_freq.py
ch-kr: hmmm. what about removing these globals and restructuring grpmax to have the same structure as the age hists? something like
```
grpmax: struct{
    gnomad: struct{
        AC: int32,
        AF: float64,
        AN: int32,
        homozygote_count: int64,
        pop: str,
        faf95: float64,
    },
    non-ukb: struct{
            AC: int32,
            AF: float64,
            AN: int32,
            homozygote_count: int64,
            pop: str,
            faf95: float64,
        }
} 
```

mike-w-wilson: I'm fine with that but just confirming we're dropping age hists so it actually wont look like the above?

ch-kr: oh yes sorry that wasn't clear. we're dropping the non-UKB age hists, so that annotation looks like this now, right?
```
age_hists: struct{
            age_hist_hom: struct {
                bin_edges: array<float64>, 
                bin_freq: array<int64>, 
                n_smaller: int64, 
                n_larger: int64
            }, 
            age_hist_het: struct {
                bin_edges: array<float64>, 
                bin_freq: array<int64>, 
                n_smaller: int64, 
                n_larger: int64
            },
        }
```

mike-w-wilson:
KC approved the structure and schema output via slack, will merge once we resolve the pylint check:https://github.com/broadinstitute/gnomad_qc/pull/416

PR #377, opened 2023-07-20
Create vrs_Dockerfile in v3 annotations

matren395:
add VRS Dockerfile for v3 to gnomad_qc/v3/annotations - should be good to go!


KoalaQin (changes requested):
Could you rename it like Dockerfile_vrs* (*for version)?

PR #378, opened 2023-07-20
Move VRS annotation batch script from tgg_methods to gnomad_qc/v3

KoalaQin:
The following changes needed for the move: 
1. since tgg_methods can't be installed / imported, we copied the init function to the beginning of the script; 
2. deleted unused imports


KoalaQin:
@matren395, it passed the checks, so you can review now. If you are looking in VS code, you should be able to see where I changed.

matren395:
lgtm - any clue what's up with no tgg_methods permissions?

matren395 (approved):

KoalaQin:
@matren395 not a permission issue, just there is no setup.py or pyproject.toml file, so we can't import.

PR #383, opened 2023-08-01
Update `get_gnomad_v4_vds` with some additional helpful parameters

jkgoodrich:


mike-w-wilson (approved):
One small thing then LGTM

PR #385, opened 2023-08-01
Suggested changes to use updated VDS loading function and modified annotate_freq

jkgoodrich:
I modified the `annotate_freq` to use the approach that you were using for adding additional stratifications. I also made it more general so it can take additional functions for adding more annotations. I think some of this should be moved to gnomad_methods, but I think you can look over this and put in a PR for that with any changes you think need to be made, then I can delete them from this PR. Some documentation still needs fixing/adding to, but I wanted to get thoughts before we put time into doing that.

I still have the second parts of the script to look over, but wanted to get this to you in the meantime.

### DIFF
--- a/gnomad_qc/v4/annotations/generate_freq.py
+++ b/gnomad_qc/v4/annotations/generate_freq.py
@@ -3,7 +3,7 @@ Script to generate the frequency data annotations across v4 exomes.
 
 This script first splits the v4 VDS into multiples VDSs based which are then densified
 and annotated with frequency data and histograms. The VDSs are then merged back together
-in a hail Table. Next the script corrects for the high AB heterzygous GATK artifact in
+in a hail Table. Next the script corrects for the high AB heterozygous GATK artifact in
 existing annotations when given the AF threshold using a high AB het array. The script
 then computes the inbreeding coefficient using the raw call stats. Finally, it computes
 the filtering allele frequency and grpmax with the AB-adjusted frequencies.
@@ -12,7 +12,7 @@ import argparse
 import itertools
 import logging
 from copy import deepcopy
-from typing import Dict, List, Optional, Union
+from typing import Callable, Dict, List, Optional, Tuple, Union
 
 import hail as hl
 from gnomad.resources.grch38.gnomad import DOWNSAMPLINGS, POPS_TO_REMOVE_FOR_POPMAX
@@ -39,7 +39,6 @@ from gnomad_qc.resource_utils import (
 from gnomad_qc.slack_creds import slack_token
 from gnomad_qc.v4.resources.annotations import get_freq
 from gnomad_qc.v4.resources.basics import get_gnomad_v4_vds
-from gnomad_qc.v4.resources.meta import meta
 
 logging.basicConfig(
     level=logging.INFO,
@@ -48,6 +47,25 @@ logging.basicConfig(
 )
 logger = logging.getLogger("gnomAD_frequency")
 logger.setLevel(logging.INFO)
+
+FREQ_ROW_FIELDS = [
+    "freq",
+    "high_ab_hets_by_group_membership",
+    "qual_hists",
+    "raw_qual_hists",
+    "age_hist_het",
+    "age_hist_hom",
+]
+"""
+List of final top level row and global annotations created from dense data that we
+want on the frequency HT before deciding on the AF cutoff.
+"""
+
+FREQ_GLOBAL_FIELDS = ["freq_meta", "age_distribution"]
+"""
+List of final global annotations created from dense data that we want on the frequency
+HT before deciding on the AF cutoff.
+"""
 
 
 def get_freq_resources(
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -101,106 +119,84 @@ def get_vds_for_freq(
...
-                meta_ht[vds.variant_data.col_key].project_meta.ukb_sample,
-                "ukb",
-                "non_ukb",
+                vds.variant_data.meta.project_meta.ukb_sample, "ukb", "non_ukb"
             ),
         ),
     )
 
+    # TODO: What above function? Also I think this can be done at a different time.
     # Downsamplings are done outside the above function as we need to annotate
```
jkgoodrich, line 167: What above function?

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -101,106 +119,84 @@ def get_vds_for_freq(
...
-        vds.reference_data,
-        vds.variant_data.select_entries("LA", "LAD", "DP", "GQ", "LGT", "_het_non_ref"),
+    vds.variant_data = vds.variant_data.select_entries(
+        "LA", "LAD", "DP", "GQ", "LGT", "_het_non_ref"
     )
 
-    return vds
-
-
-def annotate_non_ref_het(vds: hl.vds.VariantDataset) -> hl.vds.VariantDataset:
```
jkgoodrich, line 191: This is only used once and is not a complicated expression, do you think it needs to be a function?

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -315,6 +311,7 @@ def get_sample_age_bin(
     return hl.or_missing(hl.is_defined(sample_age), bin_label)
 
 
+# TODO: Does this need its own function?
 def compute_age_hist(mt: hl.MatrixTable) -> hl.MatrixTable:
```
jkgoodrich, line 315: Does this need its own function?

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -367,6 +364,22 @@ def correct_age_hists(ht: hl.Table) -> hl.Table:
...
+#        }
+#        if split_adj_and_raw:
+#            return hl.struct(
+#                raw_qual_hists=hl.struct(**qual_hists),
+#                qual_hists=hl.struct(**adj_qual_hists)
+#            )
+#        else:
+#            qual_hists.update({f"{k}_adj": v for k, v in adj_qual_hists.items()})
+#    return hl.struct(**qual_hists)
 def compute_qual_hists(mt: hl.MatrixTable) -> hl.MatrixTable:
```
jkgoodrich, line 383: I think `qual_hist_expr` in gnomad_methods should just be modified to either have an option to return them like this or to return them like this by default and just have a breaking change.

Modification could just be to add `split_adj_and_raw` parameter and then change end of function to something like this:
```
    if adj_expr is not None:
        adj_qual_hists = {
            qual_hist_name: hl.agg.filter(adj_expr, qual_hist_expr)
            for qual_hist_name, qual_hist_expr in qual_hists.items()
        }
        if split_adj_and_raw:
            return hl.struct(
                raw_qual_hists=hl.struct(**qual_hists),
                qual_hists=hl.struct(**adj_qual_hists)
            )
        else:
            qual_hists.update({f"{k}_adj": v for k, v in adj_qual_hists.items()})

    return hl.struct(**qual_hists)
```

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -461,90 +474,244 @@ def generate_faf_grpmax(ht: hl.Table) -> hl.Table:
...
-    :return: Hail Table with inbreeding coefficient annotation.
-    """
-    ht = ht.annotate(
-        InbreedingCoeff=bi_allelic_site_inbreeding_expr(callstats_expr=ht.freq[1])
-    )
-    return ht
-
-
+# TODO: Move to gnomad_methods.
 def annotate_downsamplings(
```
jkgoodrich, line 478: Move to gnomad_methods.

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -461,90 +474,244 @@ def generate_faf_grpmax(ht: hl.Table) -> hl.Table:
...
-def annotate_freq_and_high_ab_hets(
+    vmt = vds.variant_data
+    s_by_strata = vmt.aggregate_cols(
+        hl.agg.group_by(strata_expr, hl.agg.collect_as_set(vmt.s))
+    )
+
+    return [hl.vds.filter_samples(vds, list(s)) for strata, s in s_by_strata.items()]
+
+
+def compute_freq_by_strata(
```
jkgoodrich, line 559: As is, it works like previously where some groupings of samples can have 0 samples. Do we want to remove these groups instead?

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -570,274 +738,170 @@ def annotate_freq_and_high_ab_hets(
...
+        is specified within the struct, downsamplings by population is also computed.
+    :param entry_agg_funcs: When specified, additional annotations are added to the
+        output Table/MatrixTable. The keys of the dict are the names of the annotations
+        and the values are tuples of functions. The first function is used to transform
+        the `mt` entries in some way, and the second function is used to aggregate the
+        output from the first function.
+    :param annotate_mt: Whether to return the full MatrixTable with annotations added
+        instead of only a Table with `freq` and other annotations. Default is True.
+    :return: MatrixTable or Table with `freq` annotation.
     """
```
jkgoodrich, line 819: I Modified error organization a little bit to make it easier to look at.
I also feel like Not implemented seems like the wrong error here, maybe ValueError?

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -1181,8 +1221,10 @@ def main(args):  # noqa: D103
                 freq_ht = generate_freq_and_hists_ht(vds, ab_cutoff=ab_cutoff, idx=idx)
                 freq_hts[idx] = freq_ht
 
-            freq_ht = combine_freq_hts(freq_hts, final_rows_anns, final_globals_anns)
+            freq_ht = combine_freq_hts(freq_hts, FREQ_ROW_FIELDS, FREQ_GLOBAL_FIELDS)
         else:
+            # TODO: Will this only have the fields in FREQ_ROW_FIELDS and
+            #  FREQ_GLOBAL_FIELDS?
             freq_ht = generate_freq_and_hists_ht(vds, ab_cutoff=ab_cutoff)
```
jkgoodrich, line 1228: Will this only have the fields in FREQ_ROW_FIELDS and FREQ_GLOBAL_FIELDS?

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -461,90 +474,244 @@ def generate_faf_grpmax(ht: hl.Table) -> hl.Table:
     return ht
 
 
-def compute_inbreeding_coeff(ht: hl.Table) -> hl.Table:
```
jkgoodrich, line 464: I don't think this really needs its own function

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -461,90 +474,244 @@ def generate_faf_grpmax(ht: hl.Table) -> hl.Table:
...
 
-    return mt
+    return t
 
 
-def split_vds(
-    vds: hl.vds.VariantDataset, split_annotation: hl.expr.Expression
-) -> hl.ArrayExpression:
+# TODO: Move to gnomad_methods.
+def split_vds_by_strata(
```
jkgoodrich, line 541: Move to gnomad_methods

gnomad_qc/v4/annotations/generate_freq.py
mike-w-wilson: Annotation* but we can probably delete this comment.
```suggestion
    # Downsamplings are done outside the above annotation as we need to annotate
```

mike-w-wilson (approved):

PR #389, opened 2023-08-07
Suggestions to v4 frequency PR for frequency correction

jkgoodrich:


gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -466,35 +491,54 @@ def combine_freq_hts(
...
-        hl.enumerate(ht["freq_meta"], index_first=False)
-        .filter(lambda x: x[0].contains(age_group_key))
-        .map(lambda x: (x[0][age_group_key], x[1]))
+    # TODO: Comment
+    call_stats_expr = hl.map(
+        lambda f, g: hl.struct(
+            AC=hl.int32(f.AC + g),
+            AN=f.AN,
+            homozygote_count=f.homozygote_count + g,
+            AF=hl.if_else(f.AN > 0, (f.AC + g) / f.AN, hl.missing(hl.tfloat64)),
```
mike-w-wilson, line 508:
```suggestion
            AC=hl.int32(f.AC + g),
            AF=hl.if_else(f.AN > 0, (f.AC + g) / f.AN, hl.missing(hl.tfloat64)),
            AN=f.AN,
            homozygote_count=f.homozygote_count + g,
```
Changing the order so the if_else works on L529

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -466,35 +491,54 @@ def combine_freq_hts(
...
+        )
+        for x in FREQ_ROW_FIELDS
+        if "qual_hist" in x
+    }
+
+    # TODO: Comment
+    no_ab_adjusted_expr = {f"ab_adjusted_{x}": ht[x] for x in FREQ_ROW_FIELDS}
+    ht = ht.select(
+        *FREQ_ROW_FIELDS,
+        **hl.if_else(
```
mike-w-wilson, line 529: This complains about the order of the annotations in consequent and the alternate. It looks like consequent freq is `AC: int32, AN: int32, homozygote_count: int64, AF: float64` and alternate is `AC: int32, AF: float64, AN: int32, homozygote_count: int32`. I added a suggestion to L505

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -466,35 +491,54 @@ def combine_freq_hts(
...
+            ]
+        )
+        for x in FREQ_ROW_FIELDS
+        if "qual_hist" in x
+    }
+
+    # TODO: Comment
+    no_ab_adjusted_expr = {f"ab_adjusted_{x}": ht[x] for x in FREQ_ROW_FIELDS}
+    ht = ht.select(
+        *FREQ_ROW_FIELDS,
```
mike-w-wilson, line 528: I think it's worth keeping `high_ab_hets_by_group` in here so we can spot check the correction. We can use the uncorrected HT and corrected during testing too I guess but the finalize arg you mention below could just drop it. We could also put it back in FREQ_ROW_FIELDS even though it has high_ab in the name?
```suggestion
        "high_ab_hets_by_group",
        *FREQ_ROW_FIELDS,
```

mike-w-wilson (commented):
Mostly just added documentation to the new stuff where it was noted and then a couple suggestions/questions.

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -64,18 +66,23 @@
...
     "raw_qual_hists",
     "age_hists",
 ]
+ALL_FREQ_ROW_FIELDS = FREQ_ROW_FIELDS + FREQ_HIGH_AB_HET_ROW_FIELDS
 """
 List of final top level row and global annotations created from dense data that we
 want on the frequency HT before deciding on the AF cutoff.
 """
-
+# TODO: What are we doing about the sample count struct?
```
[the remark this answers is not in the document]
jkgoodrich: This was a request we got previously, so we should probably add it.

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -290,92 +319,96 @@ def generate_freq_and_hists_ht(
...
+
+    logger.info("Annotating frequencies and counting high AB het calls...")
+    freq_ht = compute_freq_by_strata(
+        mt.annotate_cols(group_membership=ht[mt.col_key].group_membership),
+        entry_agg_funcs={"high_ab_hets_by_group": (_high_ab_het, hl.agg.sum)},
     )
+    freq_ht = freq_ht.annotate_globals(**ht.index_globals())
+
     logger.info("Setting Y metrics to NA for XX groups...")
+    freq_ht = annotate_freq_index_dict(freq_ht)
```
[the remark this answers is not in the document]
jkgoodrich: The `annotate_freq_index_dict`? I replaced the call to `make_freq_index_dict_from_meta` from above with this

gnomad_qc/v4/annotations/generate_freq.py
jkgoodrich: Thank you, I forgot to change after I put in the fix to gnomad_methods last week

gnomad_qc/v4/annotations/generate_freq.py
jkgoodrich: Thank you for adding the fix!

gnomad_qc/v4/annotations/generate_freq.py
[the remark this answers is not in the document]
mike-w-wilson: Sorry, I wasnt clear. I was thinking the output of this isnt used because we recall the function in the combine step but I guess we have it so the split/combine is optional so for the code to be complete we should build this here too.

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -290,92 +322,105 @@ def generate_freq_and_hists_ht(
...
     additional_strata_expr = [
-        {"gatk_version": mt.gatk_version},
-        {"gatk_version": mt.gatk_version, "pop": mt.pop},
-        # TODO: confirm this is how we want to name this.
-        {"ukb_sample": mt.ukb_sample},
+        {"gatk_version": ht.gatk_version},
+        {"gatk_version": ht.gatk_version, "pop": ht.pop},
+        {"ukb_sample": ht.ukb_sample},
     ]
+    logger.info("Building frequency stratification list..."
```
mike-w-wilson, line 340:
```suggestion
    logger.info("Building frequency stratification list...")
```
oops! My bad, this is breaking pylint.

mike-w-wilson (approved):
LGTM

PR #392, opened 2023-08-15
process Pangolin scores from VCF to HT

KoalaQin:
Since we might have multiple scores for some variants in different genes, this code first take a maximum across splice_gain and splice_loss in each gene, then take a maximum across multiple genes. We keep the delta scores, the position information and the gene with the maximum.

### DIFF
--- a/gnomad_qc/v4/annotations/insilico_predictors.py
+++ b/gnomad_qc/v4/annotations/insilico_predictors.py
@@ -93,15 +93,21 @@ def create_pangolin_grch38_ht(vcf_path: str) -> hl.Table:
     """
     Create a Hail Table with Pangolin score for splicing for GRCh38.
 
-    .. Note::
-    The Pangolin scores were generated for variants in gene body only. The first run contains variants from gnomAD v3 genomes.
+    .. note::
+    The score was based on the splicing prediction tool Pangolin:
+    Zeng, T., Li, Y.I. Predicting RNA splicing from DNA sequence using Pangolin.
+     Genome Biol 23, 103 (2022). https://doi.org/10.1186/s13059-022-02664-4
+
+    There's no precomputed for all variants, the scores were generated for
+    gnomAD v3 variants in gene body only.
 
     :param vcf_path: path to the VCF files with Pangolin scores for splicing.
     :return: Hail Table with Pangolin score for splicing for GRCh38.
     """
-    mt = hl.import_vcf(vcf_path, min_partitions=1000, reference_genome="GRCh38")
-
-    ht = mt.rows()
+    vcf_path = (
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -89,6 +89,101 @@ def _load_cadd_raw(cadd_tsv) -> hl.Table:
...
+                hl.empty_array(hl.tint32).append(
+                    hl.int(ht.pango.split(delim=":|\\|")[1])
+                )
+            ).append(hl.int(ht.pango.split(delim=":|\\|")[3])),
+            delta_scores=(
+                hl.empty_array(hl.tfloat64).append(
+                    hl.float(ht.pango.split(delim=":|\\|")[2])
+                )
+            ).append(hl.abs(hl.float(ht.pango.split(delim=":|\\|")[4]))),
+        )
```
mike-w-wilson, line 127: A note clarifying why we are the scores are in two locations in the VCF would be helpful here. Also, you take a different approach with this tool than you did with splice_ai (where you defined variables before the annotate function and then used those variables within a hl.map call to cast them to floats within the annotate) . I think either works (though I found spliceai's approach a bit cleaner, but for readability sake, the approach should be kept consistent.

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -89,6 +89,101 @@ def _load_cadd_raw(cadd_tsv) -> hl.Table:
...
+    )
+
+    logger.info(
+        "Getting the max Pangolin score for each variant across consequences per"
```
mike-w-wilson, line 131:
```suggestion
        "Getting the max Pangolin score across consequences for each variant  per"
```

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -89,6 +89,101 @@ def _load_cadd_raw(cadd_tsv) -> hl.Table:
...
+            ),
+            position_max=hl.if_else(
+                ht.pangolin.ds_max > 0,
+                ht.pangolin.positions[
+                    ht.pangolin.delta_scores.index(ht.pangolin.ds_max)
+                ],
+                hl.missing(hl.tint32),
+            ),
+        )
+    )
```
mike-w-wilson, line 153: This is extremely similar (the same if I am not mistaken) to the splice_ai approach. I think it would make sense to create a function that is something like `process_splice_scores`.

mike-w-wilson (commented):
A lot of this is a copy and paste from splice ai. Since they live in the same module, lets create a more general function for processing the splice scores.

KoalaQin:
closing this PR because main is constantly changing, will create a new PR.

PR #396, opened 2023-08-17
Update python version 3.8 -> 3.11

jkgoodrich:
Suggested for [Hail](https://atgu.slack.com/archives/C055FKCJHAB/p1691438191032509)

Also added `gnomad` as a known third party to keep `isort` from changing the spacing


.github/workflows/ci.yml
```diff
@@ -13,10 +13,10 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: recursive
-      - name: Setup Python 3.8
+      - name: Setup Python 3.11
         uses: actions/setup-python@v2
```
ch-kr, line 17:
```suggestion
        uses: actions/setup-python@v4
```
^I think this is the current version if we want to update; https://github.com/actions/setup-python

.github/workflows/ci.yml
```diff
@@ -13,10 +13,10 @@ jobs:
...
         with:
-          python-version: 3.8
+          python-version: 3.11
       - name: Use pip cache
         uses: actions/cache@v2
```
ch-kr, line 21: I think this is v3 now as well: https://github.com/actions/cache

ch-kr (approved):
some minor comments asking if we want to update tool versions, but LGTM!

.pre-commit-config.yaml
ch-kr, line 1: as a general question for the future, should we exclude v2 code from checks? we aren't actively using it, so we shouldn't continually have to maintain it. what are your thoughts? this also came up in this PR (https://github.com/broadinstitute/gnomad_methods/pull/576); v2 still uses the deprecated `hl.cond`

ch-kr (approved):
one non-urgent question for the future, but LGTM!

also noting that I only quickly looked over the files reformatted by black

.pre-commit-config.yaml
jkgoodrich: Yeah, we do for some of them that require more than just a quick run to reformat ex: `pydocstyle --match-dir='(?!v2|cuKING).*' gnomad_qc` 
I was going to tell @KoalaQin at the meeting today that we don't need to go back and update `hl.cond` in past versions

PR #402, opened 2023-08-22
Add v4 variant QC annotation HT creation and true positive VCF export

jkgoodrich:


gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -235,14 +254,152 @@ def run_generate_sib_stats(
...
+
+    if impute_features:
+        ht = median_impute_features(ht, {"variant_type": ht.variant_type})
+
+    ht = ht.repartition(n_partitions, shuffle=False)
+    ht = ht.checkpoint(
+        hl.utils.new_temp_file("variant_qc_annotations", "ht"), overwrite=True
+    )
+
+    summary = ht.group_by(
```
matren395, line 341: would we also want to return `summary` if we want to look at it later?

gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -317,7 +495,20 @@ def main(args):
...
+    transmitted_singletons = args.transmitted_singletons
+    sibling_singletons = args.sibling_singletons
+
+    true_positive_type = None
+    if transmitted_singletons and sibling_singletons:
+        true_positive_type = "transmitted_singleton.sibling_singleton"
+    elif transmitted_singletons:
+        true_positive_type = "transmitted_singleton"
+    elif sibling_singletons:
+        true_positive_type += "sibling_singleton"
```
matren395, line 507: I'm a bit turned around, why not `=sibling_singleton` ?

matren395 (commented):
other than my small comments seems good! Might take another look at it tomorrow

gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
jkgoodrich: It's really just a quick look to make sure things look as expected. If needed we can summarize the `ht` again easily, or look up the job output

gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -316,14 +336,152 @@ def run_generate_sib_stats(
...
-    test: bool, overwrite: bool
+    test: bool, overwrite: bool, true_positive_type: Optional[str] = None
 ) -> PipelineResourceCollection:
     """
     Get PipelineResourceCollection for all resources needed in the variant QC annotation pipeline.
 
     :param test: Whether to gather all resources for testing.
     :param overwrite: Whether to overwrite resources if they exist.
+    :param true_positive_type: Type of true positive variants to use for true positive
+        VCF path resource. Default is None.
```
matren395, line 484:
```suggestion
        VCF path resource when exporting. Default is None.
```

A bit unclear abt code - does it not export the True Positive VCFs if this isn't passed? If so, may want to mention in params. If I'm just misreading it though, then nvm

matren395 (commented):
a few minor comments!

gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -316,14 +336,152 @@ def run_generate_sib_stats(
...
+    Get Tables with raw and adj true positive variants to export as a VCF for use in VQSR.
+
+    :param ht: Input Table with transmitted singleton and sibling singleton information.
+    :param transmitted_singletons: Whether to include transmitted singletons in the
+        true positive variants.
+    :param sibling_singletons: Whether to include sibling singletons in the true
+        positive variants.
+    :return: Dictionary of 'raw' and 'adj' true positive variant Tables.
+    """
+    if not transmitted_singletons and not sibling_singletons:
```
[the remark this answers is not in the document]
jkgoodrich: Just adding to clarify what we are making. Want to make two files (there is also an option for a 3rd, but we probably won't test it, so no reason to make it).
- VCF of just transmitted singletons: These are determined by looking at the trios and identifying variants that are found in one parent and the child, and those two individuals are the only two in the callset with the variant.
- VCF of transmitted singletons (defined above) and sibling singletons: sibling singletons are determined by looking at two individuals that are siblings and identifying variants that are found in both siblings, and these two individuals are the only individuals in the callset with that variant.
- We can also export just sibling singletons, but it's not really worth it.

We use these as true positives in the variant QC training so that we can see what the features we use in the model look like on rare variants that we have high confidence in being real.

This if statement is just throwing an error because if you don't pick one of these you are just asking for an empty VCF

gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
jkgoodrich: `get_variant_qc_annotation_resources` is only getting the resources and not actually performing the export

matren395: ah sorry, makes sense now

gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -519,6 +733,55 @@ def main(args):
...
+        "Arguments used to define true positive variant set.",
+    )
+    tp_vcf_args.add_argument(
+        "--export-true-positive-vcfs",
+        help=(
+            "Exports true positive variants (--transmitted-singletons and/or"
+            " --sibling-singletons) to VCF files."
+        ),
+        action="store_true",
+    )
```
matren395, line 767: Is there ever a use case where either --transmitted-singletons and/or --sibling-singletons are passed but not --export-true-positives-vcfs ? Do they, or the PipelineStepResourceCollection named 'export_true_positive_vcfs', have any other function in the code or are read in elsewhere ? 

If there isn't, you could just choose to export if either true positive types are passed - it would cut down on arguments but might be less user-friendly when passing argument. 

Think the only code change would be changing line #668 to `if true_positive_type:` , but only if this is the case

matren395 (commented):
Last note!

matren395 (approved):
Everything I said has been addressed, brought up my last question at our Variant QC meeting, all good! I think my side is still showing merge conflicts though, but i'm not sure if that's just talking abt my local version of the repo or not

PR #407, opened 2023-08-25
non_ukb downsamplings for v4 freq

mike-w-wilson:
Misnamed the branch but this adds the non_ukb subset downsamplings as well as a couple other things, like the try, finally to copy the log on OOM failures, and also the freq_meta_sample_count to the globals. I opted for doing another pass through the data instead of adding more strata to the initial pass as it would almost double our aggregations and I figured we may end up with the 137 again. This extra pass was also discussed in slack back in May before we know how much more work freqs would be so I figured it was the preferred approach. This also prevents changes to  the gnomad_methods build strata and annotate_freq functions. The way the freq_meta is updated, adding the subst key, may change in another round since I need to annotate the subset frequencies onto the main freq table still. I'll make another PR for adding in the subset specific annotations/finalize option we discussed but wanted to get this to you.

### DIFF
--- a/gnomad_qc/v4/annotations/generate_freq.py
+++ b/gnomad_qc/v4/annotations/generate_freq.py
@@ -187,7 +187,7 @@ def get_vds_for_freq(
         ukb_sample=hl.if_else(project_meta_expr.ukb_sample, "ukb", "non_ukb"),
     )
 
-    # TODO: Add comment and logger.
+    logger.info("Getting age distribution of all samples in the callset...")
     vmt = vmt.annotate_globals(
         age_distribution=vmt.aggregate_cols(hl.agg.hist(vmt.age, 30, 80, 10))
     )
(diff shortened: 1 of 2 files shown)


gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -202,166 +244,224 @@ def get_vds_for_freq(
...
+            - sex_karyotype
+            - gatk_version
+            - age
+            - ukb_sample
+
+    :param mt: Input MatrixTable.
+    :param ds_ht: Table with downsampling annotations.
+    :param meta_ht: Table with sample metadata annotations.
+    :return: Hail Table with frequency annotations.
+    """
```
jkgoodrich, line 398:
```suggestion
    """
    meta_ht = meta_ht.semi_join(mt.cols())
```

jkgoodrich (approved):
Lets merge it

PR #410, opened 2023-08-28
Add checkpoint to compute info and copy log file

jkgoodrich:


gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -211,9 +210,11 @@ def run_compute_info(mt: hl.MatrixTable) -> hl.Table:
             "release": mt.meta.release,
             "unrelated": ~mt.meta.sample_filters.relatedness_filters.related,
         },
+        n_partitions=None,
     )
+    info_ht = info_ht.checkpoint(hl.utils.new_temp_file("compute_info"), overwrite=True)
```
ch-kr, line 215: I didn't know this (`new_temp_file`) existed, that is so cool. do you need to specify "ht" here also?

gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -227,8 +228,11 @@ def run_compute_info(mt: hl.MatrixTable) -> hl.Table:
     ).rows()
 
     info_ht = info_ht.annotate(set_long_AS_missing_info=ht[info_ht.key])
+    info_ht = info_ht.checkpoint(
+        hl.utils.new_temp_file("compute_info_AS_missing"), overwrite=True
```
ch-kr, line 232:
```suggestion
        hl.utils.new_temp_file("compute_info_AS_missing", "ht"), overwrite=True
```
? (not sure if required)

gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
```diff
@@ -211,9 +210,11 @@ def run_compute_info(mt: hl.MatrixTable) -> hl.Table:
             "release": mt.meta.release,
             "unrelated": ~mt.meta.sample_filters.relatedness_filters.related,
         },
+        n_partitions=None,
```
ch-kr, line 213: did you skip the naive_coalesce here to make the join with the `set_long_AS_missing` versions of the annotations faster?

ch-kr (commented):
a couple basic questions

gnomad_qc/v4/annotations/generate_variant_qc_annotations.py
jkgoodrich: I did it because for the really long step of computation, instead of using the full ~15,000 partitions, it was splitting it into only 5,000 partitions so I was worried about things possibly being slower and taking more memory.

ch-kr (approved):
LGTM!

PR #416, opened 2023-09-08
Add cuKING to ignore-paths

mike-w-wilson:
I think this will fix our current pylint failures. I think it'sthe right behavior -- we don't want to be linting a repo we don't own.

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -7,4 +7,4 @@ disable=
   R,
   C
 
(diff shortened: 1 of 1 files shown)


ch-kr (approved):
I agree that we shouldn't lint this repo we don't control, particularly since I think both Leo and Vlad have left the CPG

PR #419, opened 2023-09-10
Add interval QC and updates to work with final annotated RF HT v4 to variant QC random forest script

jkgoodrich:


matren395 (commented):
(tentatively) lgtm! I'll give it another look over Monday morning, but looks good and I wouldn't wanna slow this down

gnomad_qc/v4/variant_qc/random_forest.py
```diff
@@ -51,50 +48,48 @@
...
-]
 LABEL_COL = "rf_label"
 PREDICTION_COL = "rf_prediction"
 TRAIN_COL = "rf_train"
-TRUTH_DATA = ["hapmap", "omni", "mills", "kgp_phase1_hc"]
+
 
 def train_rf(
     ht: hl.Table,
+    test: bool = False,
```
matren395, line 58: being able to pass Test into this method is a really good touch , big fan

gnomad_qc/v4/variant_qc/random_forest.py
```diff
@@ -104,13 +99,30 @@ def train_rf(
             for x in test_intervals
         ]
 
+    if test:
+        logger.info("Filtering to chr22 and evaluation intervals for testing...")
```
matren395, line 103: can I ask - why Chr22? Just becuase it's the smallest non-sex Chromosome and it's convention or?

gnomad_qc/v4/variant_qc/random_forest.py
```diff
@@ -104,13 +99,30 @@ def train_rf(
...
+        chr22_interval = [hl.parse_locus_interval("chr22", reference_genome="GRCh38")]
+        ht = hl.filter_intervals(ht, chr22_interval + test_intervals)
+
+    logger.info("Annotating true positives and false positives in HT...")
+    fp_expr = ht.fail_hard_filters
+    tp_expr = ht.omni | ht.mills
+    if transmitted_singletons:
```
matren395, line 110: switch of the logic from last time - I'll have to make a note to pass it in pass these in when I'm running the RF code in the future

matren395 (commented):
took another look and makes a lot more sense - thanks for the improvements to the logic and structure! Once we get it passing those PyLint checks it should be good to go

matren395 (approved):
lgtm!

gnomad_qc/v4/variant_qc/random_forest.py
jkgoodrich: I don't really have a good reason other than it's small, and not chr20 since that is being used as the test intervals

gnomad_qc/v4/variant_qc/random_forest.py
jkgoodrich: yeah, it's just personal preference, either way if fine, but I just like this way better

PR #420, opened 2023-09-12
Mw/checkpoint split vds

mike-w-wilson:
This adds a checkpoint after splitting the VDS which resolves our 137 issues. The checkpoint drops the aggs mem usage from 32G to 500 MiB. I did not use a temp storage location (gnomad-tmp is 14 days) for the split VDS given the uncertainties around frequency runtime.

### DIFF
--- a/gnomad_qc/v4/annotations/generate_freq.py
+++ b/gnomad_qc/v4/annotations/generate_freq.py
@@ -42,7 +42,7 @@ from gnomad_qc.resource_utils import (
     PipelineStepResourceCollection,
 )
 from gnomad_qc.slack_creds import slack_token
-from gnomad_qc.v4.resources.annotations import get_freq, get_split_vds_path
+from gnomad_qc.v4.resources.annotations import get_freq, get_split_vds
 from gnomad_qc.v4.resources.basics import get_gnomad_v4_vds, get_logging_path
 
 logging.basicConfig(
@@ -126,7 +126,7 @@ def get_freq_resources(
     )
     write_split_vds = PipelineStepResourceCollection(
         "--write-split-vds",
-        output_resources={"split_vds": get_split_vds_path(test=test)},
+        output_resources={"split_vds": get_split_vds(test=test)},
     )
     run_freq_and_dense_annotations = PipelineStepResourceCollection(
         "--run-freq-and-dense-annotations",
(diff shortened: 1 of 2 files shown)


gnomad_qc/v4/resources/annotations.py
```diff
@@ -398,3 +398,27 @@ def get_vrs(
...
+
+    :param version: Version of annotation path to return.
+    :param data_type: Data type of annotation resource. e.g. "exomes" or "genomes".
+           Default is "exomes".
+    :param test: Whether to use a tmp path for analysis of the test Table instead
+           of the full v4 Table.
+    :return: gnomAD v4 split VDS path.
+    """
+    return (
+        f"{_annotations_root(version, test, data_type)}/gnomad.{data_type}.v{version}.split.vds"
```
jkgoodrich, line 423: Maybe add temp just to make it obvious.
```suggestion
        f"{_annotations_root(version, test, data_type)}/temp/gnomad.{data_type}.v{version}.split.vds"
```

jkgoodrich (approved):
lgtm, small suggested change on the path name. Do you think we should change any of the other temp locations we checkpoint to?

mike-w-wilson:
Thats a good point, we probably should change those as well. I doubt it will complete in 4 days and the group membership tables may be in use? I'll update.

mike-w-wilson:
Changed the split_vds path to a more general checkpoint path that accepts a name and file type. This depends on https://github.com/broadinstitute/gnomad_methods/pull/598. Back to you @jkgoodrich !

mike-w-wilson:
Just want to confirm this looks good still after the last couple commits  if you dont mind @jkgoodrich

gnomad_qc/v4/resources/annotations.py
```diff
@@ -398,3 +398,27 @@ def get_vrs(
             for version in VERSIONS
         },
     )
+
```
jkgoodrich, line 401: I noticed that the intermediate subset files are being written to the 4day bucket because there is this `from gnomad_qc.v3.resources.basics import get_checkpoint_path` and the v3 checkpoint path goes to the 4day. 


I think we should first of all not import from that (sorry, I think it was me), and second change:

```
    if intermediate_subset:
        ht_name += f".{intermediate_subset}"
        if test:
            ht_name += ".test"
        ht_path = get_checkpoint_path(ht_name, version=CURRENT_VERSION)
```

to 
```
    if intermediate_subset:
        ht_name += f".{intermediate_subset}"
        if test:
            ht_name += ".test"
        ht_path = f"{_annotations_root(version, test)}/temp/{ht_name}.ht"
```

jkgoodrich (changes requested):
One change

gnomad_qc/v4/resources/annotations.py
mike-w-wilson: Ah good catc, fixed!

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -904,23 +913,32 @@ def main(args):
...
+    hl._set_flags(use_ssa_logs="1")
     resources = get_freq_resources(overwrite, test, chrom)
 
     try:
-        if args.run_freq_and_dense_annotations:
-            logger.info("Running dense dependent steps...")
-            res = resources.run_freq_and_dense_annotations
-            res.check_resource_existence()
-
+        if args.write_split_vds:
```
jkgoodrich, line 923: Nice, I didn't think about that, thanks for splitting it up

PR #421, opened 2023-09-12
Create load_vqsr.py for v4

matren395:
Create load_data/ and load_vqsr.py with import_vqsr() for v4 - only one import that needed changed and , helpfully, 

`from gnomad_qc.v4.resources.annotations import get_vqsr_filters `

works just fine and functionality seems the same as v3


jkgoodrich (approved):
LGTM!

PR #424, opened 2023-09-12
add function to process REVEL score: a new branch

KoalaQin:
Starting from gnomAD v4, we will use REVEL scores for MANE Select transcripts and canonical transcripts only.
We take a max of the REVEL scores if a variant falls on multiple transcripts.
Since the Ensembl ID used by REVEL was not from Ensembl 105, we also only filtered transcripts that are in Ensembl 105.
This will also avoid has_duplicate.


KoalaQin:
closing because the main branch is constantly changing.

PR #425, opened 2023-09-13
Fix frequency subset resources

jkgoodrich:
- Add downsampling resource
- Move downsampling to right after split VDS so it is only done once if using subset parallelization
- Only make resources for the subsets that will be written out

### DIFF
--- a/gnomad_qc/v4/annotations/generate_freq.py
+++ b/gnomad_qc/v4/annotations/generate_freq.py
@@ -35,14 +35,13 @@ from gnomad.utils.filtering import filter_arrays_by_meta, split_vds_by_strata
 from gnomad.utils.release import make_freq_index_dict_from_meta
 from gnomad.utils.slack import slack_notifications
 from gnomad.utils.vcf import SORT_ORDER
-from hail.utils.misc import new_temp_file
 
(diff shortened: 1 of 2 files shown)


gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -110,37 +109,56 @@
...
+    ukb: bool = False,
+    non_ukb: bool = False,
 ) -> PipelineResourceCollection:
     """
     Get frequency resources.
 
     :param overwrite: Whether to overwrite existing files.
     :param test: Whether to use test resources.
     :param chrom: Chromosome used in freq calculations.
+    :param ukb: Whether to get frequency resources for UKB subset.
```
ch-kr, line 124: basic question: do you need to create a UKB subset to make the frequency calculations complete? (asking because I thought the only v4 subset was non-ukb)

jkgoodrich: yeah, they are handled differently. The non-ukb subset actually makes a subset, they other one only makes the standard freqs and then they are added together. This subsetting is just for parallelization and smaller densify purposes

ch-kr (approved):

PR #428, opened 2023-09-13
Update split VDS path

jkgoodrich:
- Change to use path output after removing duplicate UKB samples from split VDS


mike-w-wilson (approved):
This is a better name anyways!

PR #429, opened 2023-09-13
Change v4 VDS loading function to always remove UKB samples with exact same 's'

jkgoodrich:
Noticed this problem when making the split VDS for frequency

### DIFF
--- a/gnomad_qc/v4/resources/basics.py
+++ b/gnomad_qc/v4/resources/basics.py
@@ -117,7 +117,7 @@ def get_gnomad_v4_vds(
     # https://github.com/broadinstitute/ukbb_qc/blob/70c4268ab32e9efa948fe72f3887e1b81d8acb46/ukbb_qc/resources/basics.py#L286.
     # Confirmed that the col_idx of the 27 dup samples in the original UKB MT match
     # the col_idx of the dup UKB samples in the VDS.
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/resources/basics.py
```diff
@@ -110,12 +110,43 @@ def get_gnomad_v4_vds(
...
-    # We don't need to do the UKB sample removal if we're only keeping high quality or
-    # release samples since they will not be in either of those sets.
+    # Count current number of samples in the VDS.
+    n_samples = vds.variant_data.count_cols()
+
+    # Remove 27 fully duplicated IDs (same exact name for 's' in the VDS). See:
+    # https://github.com/broadinstitute/ukbb_qc/blob/70c4268ab32e9efa948fe72f3887e1b81d8acb46/ukbb_qc/resources/basics.py#L286.
+    # Confirmed that the col_idx of the 27 dup samples in the original UKB MT match
+    # the col_idx of the dup UKB samples in the VDS.
+    dup_ids = []
```
mike-w-wilson, line 120:
```suggestion
    logger.info("Removing 27 duplicate UKB samples by column index..."
    dup_ids = []
```

mike-w-wilson (approved):
Just added one logger but then looks good! Thanks for adding that `not release` to the hard filter step too!

PR #431, opened 2023-09-15
change driver_memory to highmem in hl.init

KoalaQin:
Set driver_memory to 'highmem' to avoid out of memory error in Hail Batch run.


klaricch (approved):
lgtm

PR #432, opened 2023-09-15
VRS docker: update Hail to 0.2.122 and reorder the caching layers

KoalaQin:
I tried to build the Docker image but haven't pushed yet, I want to have Steve look over it since he taught us some great tricks. If it looks good, I will push it to the registry and run a test on VRS to make sure it's working.

### DIFF
--- a/gnomad_qc/v4/annotations/Dockerfile_vrs084_v4
+++ b/gnomad_qc/v4/annotations/Dockerfile_vrs084_v4
@@ -11,7 +11,7 @@
 # Dockerfile for creating an image that can run the VRS Annotation Script inside of it
 # includes libpqdev and pythondev, plus psycopg2, ga4gh.vrs[extras] , and seqrepo
 
-FROM 'google/cloud-sdk:slim'
+FROM 'google/cloud-sdk:446.0.1-slim'
 
 # Copy in SeqRepo from local file structure
 # Put this big folder (~12G) in the first layers so that it doesn't get rebuilt every time
--- a/gnomad_qc/v4/annotations/Dockerfile_vrs084_v4
+++ b/gnomad_qc/v4/annotations/Dockerfile_vrs084_v4
@@ -4,6 +4,8 @@
 # 1) Uses install GCS connector from broad.io
 # 2) Installs bgzip
 # 3) Puts gnomad & gnomad_qc dependencies inside the docker image
+# 4) Reorder the installation of packages to optimize caching
+# 5) Update hail version to 0.2.122 from 0.2.113 for a few fixes to run Hail Query-on-Batch
 ############################################################################################################
 # Dockerfile for creating an image that can run the VRS Annotation Script inside of it
 # includes libpqdev and pythondev, plus psycopg2, ga4gh.vrs[extras] , and seqrepo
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/annotations/Dockerfile_vrs084_v4
```diff
@@ -34,14 +36,17 @@ RUN apt-get -y update && apt-get install -y --no-install-recommends \
 	# clean up apt cache
     rm -rf /var/lib/apt/lists/*
 
+# Copy in SeqRepo from local file structure
+# Put this big folder (~12G) in the first layers so that it doesn't get rebuilt every time
```
sjahl, line 40: If the seqrepo never really changes, it might actually be better to put this directly after the `FROM 'google/cloud-sdk:slim'` line. The packages that get installed with `apt-get` change with some regularity, probably every few weeks or so.

gnomad_qc/v4/annotations/Dockerfile_vrs084_v4
```diff
@@ -53,6 +58,17 @@ RUN wget https://github.com/vcftools/vcftools/releases/download/v${VCFTOOLS_VERS
 	&& make install \
 	&& make clean
 
+# Install bgzip
```
sjahl, line 61: It might be better to have this comment say `# Install bgzip from htslib`

As someone with no familiarity with htslib, this looked like a typo, since there wasn't anything in the install commands that mentioned bgzip!

gnomad_qc/v4/annotations/Dockerfile_vrs084_v4
```diff
@@ -53,6 +58,17 @@ RUN wget https://github.com/vcftools/vcftools/releases/download/v${VCFTOOLS_VERS
...
+ENV BGZIP_VERSION="1.17"
+RUN wget https://github.com/samtools/htslib/releases/download/1.17/htslib-1.17.tar.bz2 \
+    && tar -xjf htslib-1.17.tar.bz2 \
+    && rm htslib-1.17.tar.bz2 \
+    && cd htslib-1.17 \
+    && ./configure \
+    && make \
+    && make install \
+    && make clean \
+    && export PATH=htslib-1.17/bin:$PATH
```
sjahl, line 71: since we set the `BGZIP_VERSION` environment variable, we should use it in all the lines that reference that version, so that we only have to update that one line of code when there's a new version. So, replace all the `1.17`s in the `RUN` command with `${BGZIP_VERSION}`

sjahl (changes requested):
Overall looks good -- couple of quick things:

- if seqrepo never changes, we can probably move it even earlier in the dockerfile, so that it'll stay cached if there's any updated apt packages
- suggested a more detailed comment, and some better variable usage in the bgzip install step

sjahl (changes requested):
i made an inline suggestion for the bgzip version changes. Otherwise this looks good to me. You may want a quick review from someone on the methods team since I'm not typically a reviewer in gnomad_qc!

KoalaQin:
Yeah, I'm going to build an image with this one and tested it on Hail Batch, if it's successful, then I will ask Kristen to review it for me! Thanks a lot!

KoalaQin:
Steve already looked over the order of layers and gave some suggestions. I built and pushed the docker image, I ran the test of 200 partitions with it and it worked well! Could you have another look for me?

sjahl (approved):

PR #434, opened 2023-09-15
Move rf_runs loading so multiple models can be trained at the same time

jkgoodrich:


gnomad_qc/v4/variant_qc/random_forest.py
```diff
@@ -321,7 +319,7 @@ def main(args):
         ht = ht.checkpoint(res.rf_training_ht.path, overwrite=overwrite)
 
         logger.info("Adding run to RF run list")
-        add_model_to_run_list(ht, model_id, rf_runs, rf_run_path)
+        add_model_to_run_list(ht, model_id, get_rf_runs(rf_run_path), rf_run_path)
```
matren395, line 322: looks very straight-forward!

matren395 (approved):
lgtm!

PR #435, opened 2023-09-15
Function to process Pangolin score

KoalaQin:
Based on review of #392 and what we've done for SpliceAI, I simplified the code to get a larger gain or loss in each gene, then get a larger across genes, and it will keep the DS for gain if the absolute values are equal. 
I also make it ready to merge scores for both v4 genomes and exomes variants.


gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -159,6 +159,114 @@ def import_spliceai_vcf(path: str) -> hl.Table:
...
+            )
+        )
+    )
+    logger.info(
+        "Number of variants indicating splice gain: %s;"
+        "Number of variants indicating splice loss: %s; "
+        "Number of variants with no splicing consequence: %s",
+        ht.filter(ht.pangolin.largest_ds > 0).count(),
+        ht.filter(ht.pangolin.largest_ds < 0).count(),
+        ht.filter(ht.pangolin.largest_ds == 0).count(),
```
mike-w-wilson, line 265: This should always be zero given the explode after import right?

mike-w-wilson, line 265: An single agg may be faster, I imaigne 3 filters + count are fairly slow?

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -159,6 +159,114 @@ def import_spliceai_vcf(path: str) -> hl.Table:
...
+    # splice gain
+    ht = ht.annotate(
+        largest_ds_gene=hl.if_else(
+            hl.abs(hl.min(ht.pangolin.delta_scores))
+            > hl.abs(hl.max(ht.pangolin.delta_scores)),
+            hl.min(ht.pangolin.delta_scores),
+            hl.max(ht.pangolin.delta_scores),
+        )
+    )
+    # TODO: tried hl.zip to make it one-step, but it's not working
```
mike-w-wilson, line 243: Maybe remove this todo if you dont plan on trying to get the zip working

mike-w-wilson (commented):
Back to you @KoalaQin ! Mostly just readability again. One additional thing not within the review is that depending on which in silico module youre in, I notice you switch back and forth between using strings and expressions in functions like select, e.g. `'score' `vs `ht.score`. I think either is fine but sticking to one just makes the code feel a bit cleaner.

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -159,6 +159,114 @@ def import_spliceai_vcf(path: str) -> hl.Table:
...
+    ht = ht_g.union(ht_e)
+
+    logger.info("Exploding Pangolin scores...")
+    # `explode` will eliminate rows with empty array
+    # The VCF INFO of Pangolin is a string with the following format:
+    # gene1|pos_splice_gain:largest_increase|pos_splice_loss:largest_decrease|gene2...
+    # for example:
+    # Pangolin=ENSG00000121005.9|-86:0.25|38:-0.49|Warnings:||ENSG00000254238.1|-40:0.01|30:-0.17|Warnings:
+
+    ht = ht.annotate(pango=ht.info.Pangolin[0].split(delim="\|\|"))
```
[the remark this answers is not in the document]
KoalaQin: Because of error: 'ArrayExpression' object has no attribute 'split', if removing [0]. It's an array of strings, and there's only one string element before splitting. I will write a note.

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -159,6 +159,114 @@ def import_spliceai_vcf(path: str) -> hl.Table:
...
+    ht = ht.explode(ht.pango)
+    logger.info("Number of rows in exploded Pangolin Hail Table: %s", ht.count())
+
+    logger.info("Annotating Pangolin scores...")
+    ht = ht.annotate(
+        pangolin=hl.struct(
+            delta_scores=(
+                hl.empty_array(hl.tfloat64).append(
```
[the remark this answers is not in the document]
KoalaQin: I wanted to take a max in one gene first, that's why I put the two in one array.

gnomad_qc/v4/annotations/insilico_predictors.py
KoalaQin: yeah, this was just to show you if you have better solution.

gnomad_qc/v4/annotations/insilico_predictors.py
KoalaQin: No, explode is just for the case when a variant doesn't have Pangolin prediction and when one variant has prediction multiple genes. Zero is still a valid prediction.

KoalaQin: It's not very slow, but what agg do you mean? I found count_where can also only do one each time.

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -159,6 +159,116 @@ def import_spliceai_vcf(path: str) -> hl.Table:
...
+            )
+        )
+    )
+    logger.info(
+        "\nNumber of variants indicating splice gain: %s;\n"
+        "Number of variants indicating splice loss: %s; \n"
+        "Number of variants with no splicing consequence: %s \n",
+        ht.filter(ht.pangolin.largest_ds > 0).count(),
+        ht.filter(ht.pangolin.largest_ds < 0).count(),
+        ht.filter(ht.pangolin.largest_ds == 0).count(),
```
mike-w-wilson, line 267:
```suggestion
        hl.agg.count_where(ht.pangolin.largest_ds > 0),
        hl.agg.count_where(ht.pangolin.largest_ds < 0),
        hl.agg.count_where(ht.pangolin.largest_ds == 0),
```
Yeah thats what I was thinking, I thought hail had an optimization on aggs over a filter(...).count(), like it may actually do all three of these at the same time but I'm not actually certain.

mike-w-wilson (commented):
So so close, this will be the last one!

mike-w-wilson (approved):

PR #439, opened 2023-09-18
Variant QC evaluation -- extract and merge truth samples

jkgoodrich:


gnomad_qc/v4/resources/basics.py
```diff
@@ -147,7 +149,7 @@ def _remove_ukb_dup_by_index(
     # We don't need to do the UKB withdrawn and pharma remove list sample removal if
     # we're only keeping high quality or release samples since they will not be in
     # either of those sets.
-    if not high_quality_only and not release_only:
+    if not (high_quality_only or release_only or controls_only):
```
matren395, line 152: I actually didn't realize Python allowed triple logical statements like this, v cool !

gnomad_qc/v4/resources/basics.py
```diff
@@ -253,6 +255,10 @@ def _remove_ukb_dup_by_index(
             filter_expr |= hl.literal(TRUTH_SAMPLES_S).contains(meta_ht.s)
         vds = hl.vds.filter_samples(vds, meta_ht.filter(filter_expr))
 
+    if controls_only:
```
matren395, line 258: everything in this file makes sense

gnomad_qc/v4/resources/variant_qc.py
```diff
@@ -70,24 +70,26 @@ def _variant_qc_root(version: str = CURRENT_VERSION, test: bool = False) -> str:
...
+    :param test: Whether to use a tmp path for variant QC tests.
+    :return: Path to callset truth sample MT.
     """
     if mt:
         return VersionedMatrixTableResource(
             CURRENT_VERSION,
             {
                 version: MatrixTableResource(
-                    f"{_variant_qc_root(version)}/truth_samples/gnomad.exomes.v{version}.{truth_sample}.mt"
+                    f"{_variant_qc_root(version, test=test)}/truth_samples/gnomad.exomes.v{version}.{truth_sample}.mt"
```
matren395, line 92: QQ: are these populated yet ?

gnomad_qc/v4/variant_qc/evaluation.py
```diff
@@ -154,8 +158,80 @@ def create_aggregated_bin_ht(model_id: str) -> hl.Table:
...
+    :return: PipelineResourceCollection containing resources for all steps of the
+        variant QC evaluation pipeline.
+    """
+    # Initialize variant QC evaluation pipeline resource collection.
+    evaluation_pipeline = PipelineResourceCollection(
+        pipeline_name="variant_qc_evaluation",
+        overwrite=overwrite,
+        pipeline_resources={
+            "annotations/generate_variant_qc_annotations.py --compute-info --split-info": {
+                "info_ht": get_info(split=True)
```
matren395, line 179: would there ever be a case here where test=True for get_info() ?

gnomad_qc/v4/variant_qc/evaluation.py
```diff
@@ -193,55 +269,44 @@ def main(args):  # noqa: D103
...
-            # Filter to variants in truth data
-            truth_sample_mt = truth_sample_mt.filter_rows(
-                hl.agg.any(truth_sample_mt.GT.is_non_ref())
-            )
-            truth_sample_mt.naive_coalesce(args.n_partitions).write(
-                get_callset_truth_data(truth_sample).path,
-                overwrite=args.overwrite,
-            )
+        # Checkpoint to prevent needing to go through the large VDS multiple times.
+        vds = vds.checkpoint(new_temp_file("truth_samples", "vds"), overwrite=True)
```
matren395, line 279: do you need overwrite=True if you're using new_temp_file ? I thought it was a new name each time so an overwrite wasn't needed

matren395 (commented):
left a few questions , but looks good for the most part

gnomad_qc/v4/resources/variant_qc.py
jkgoodrich: No, they are created by the --extract-truth-samples and --merge-with-truth-data in the other script

gnomad_qc/v4/variant_qc/evaluation.py
jkgoodrich: Not in this case, we are only using the full info HT even when testing because it's already created and only annotated onto a MT to exclude AS_lowqual variants

gnomad_qc/v4/variant_qc/evaluation.py
jkgoodrich: yeah, I guess you don't really, I can remove

matren395 (approved):
makes sense , lgtm!

PR #442, opened 2023-09-20
change overwrite in checkpoint

KoalaQin:


mike-w-wilson (commented):
I think you need to pass overwrite to the revel function itself. Or can you access args from here?

KoalaQin:
Yeah, I used this: 
`hailctl dataproc submit qh-insilico gnomad_qc/v4/annotations/insilico_predictors.py --revel --overwrite`
but there was a copy of that file already.

mike-w-wilson:
Yes, I shouldnt have approved the last PR, I missed that the `_read_if_exists` was changed to an `overwrite`. `overwrite= not arg.overwrite` should not be a thing. My question is can you access the args array from inside the revel function since you are not passing it?

KoalaQin:
I could access the args, the run was successful, but the merging checks didn't pass: https://github.com/broadinstitute/gnomad_qc/actions/runs/6251213810/job/16971846070
What do we do?

mike-w-wilson:
You could try rerunning the job

KoalaQin:
Is it okay to use overwrite instead of _read_if_exists? I can also change it back to:
`_read_if_exists=not args.overwrite`?

mike-w-wilson:
It should be overwrite. If the job is fast though do you even need the checkpoint?

KoalaQin:
Maybe not, I just don't like Hail output temporary files in logger.

PR #403, opened 2023-08-24
Script to compute combined FAF of v4 genomes and exomes

jkgoodrich:

PR #446, opened 2023-09-25
Script to compute combined FAF of v4 genomes and exomes

KoalaQin:
This is based on Julia's PR #403 with some changes to use the freq_ht instead of release_ht, calculate faf and grpmax when it's not present in the prefixed freq_ht, rename the popmax to grpmax for v4, and test on one gene: PCSK9 (chr1:55039447-55064852).

### DIFF
--- a/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
+++ b/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
@@ -27,6 +27,7 @@ from gnomad_qc.resource_utils import (
 from gnomad_qc.slack_creds import slack_token
 from gnomad_qc.v4.resources.annotations import (
     get_combined_frequency,
+    get_freq,
     get_freq_comparison,
 )
 from gnomad_qc.v4.resources.basics import get_logging_path
@@ -260,7 +261,7 @@ def perform_cmh_test(
 
 
 def get_combine_faf_resources(
-    overwrite: bool, test: bool, public: bool = False
+        overwrite: bool, test: bool, public: bool = False
 ) -> PipelineResourceCollection:
     """
     Get PipelineResourceCollection for all resources needed in the combined FAF resource creation pipeline.
@@ -283,12 +284,11 @@ def get_combine_faf_resources(
         output_resources={"freq_ht": get_combined_frequency(test=test)},
         input_resources={
             # TODO: rename when release scripts are finalized.
-            "create_release.py": {
-                "exomes_ht": release_sites("exomes"),
-            },
-            "create_genome_release.py": {
-                "genomes_ht": release_sites("genomes"),
-            },
+            "generate_freq.py": {
+                "exomes_ht": get_freq(test=test, finalized=False)},
+            "create_release_sites_ht_genomes.py": {
+                # TODO: "genomes_freq_ht": get_freq(test=test, data_type='genomes', finalized=True)},
+                "genomes_ht": release_sites(public=True).versions["3.1.2"]},
         },
     )
     contingency_table_test = PipelineStepResourceCollection(
@@ -325,6 +325,16 @@ def get_combine_faf_resources(
 
     return combine_faf_pipeline
 
+def filter_gene_to_test(ht: hl.Table) -> hl.Table:
+    """
+    Filter to PCSK9 1:55039447-55064852 for testing
+
+    :param ht: Table with frequency and FAF information.
+    :return: Table with frequency and FAF information of the filtered interval of a gene
+    """
+    return hl.filter_intervals(ht,
+        [hl.parse_locus_interval("1:55039447-55064852", reference_genome='GRCh38')])
+
 
 def main(args):
     """Create combined FAF resource."""
@@ -339,9 +349,14 @@ def main(args):
         if args.create_combined_frequency_table:
             res = combine_faf_resources.combined_frequency
             res.check_resource_existence()
+            exomes_ht = res.exomes_ht.ht()
+            genomes_ht = res.genomes_ht.ht()
             if test:
-                exomes_ht = res.exomes_ht.ht()._filter_partitions(range(20))
-                genomes_ht = res.genomes_ht.ht()._filter_partitions(range(20))
+                # exomes_ht = res.exomes_freq_ht.ht()._filter_partitions(range(20))
+                # genomes_ht = res.genomes_freq_ht.ht()._filter_partitions(range(20))
+                # filter to PCSK9 1:55039447-55064852 for testing
+                exomes_ht = filter_gene_to_test(exomes_ht)
+                genomes_ht = filter_gene_to_test(genomes_ht)
 
             exomes_ht = extract_freq_info(exomes_ht, pops, faf_pops, "genomes")
             genomes_ht = extract_freq_info(genomes_ht, pops, faf_pops, "exomes")
--- a/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
+++ b/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
@@ -31,7 +31,9 @@ from gnomad_qc.v4.resources.annotations import (
     get_freq_comparison,
 )
 from gnomad_qc.v4.resources.basics import get_logging_path
-from gnomad_qc.v4.resources.release import get_combined_faf_release, release_sites
+from gnomad_qc.v4.resources.release import get_combined_faf_release
+# TODO: change to freq_ht when the freq_ht is finalized
+from gnomad_qc.v3.resources.release import release_sites
 
 logging.basicConfig(format="%(levelname)s (%(name)s %(lineno)s): %(message)s")
 logger = logging.getLogger("compute_combined_faf")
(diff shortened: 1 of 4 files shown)


gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -25,17 +25,94 @@
...
 )
 from gnomad_qc.v4.resources.basics import get_logging_path
-from gnomad_qc.v4.resources.release import get_combined_faf_release, release_sites
+from gnomad_qc.v4.resources.release import get_combined_faf_release
 
 logging.basicConfig(format="%(levelname)s (%(name)s %(lineno)s): %(message)s")
 logger = logging.getLogger("compute_combined_faf")
 logger.setLevel(logging.INFO)
 
+faf_pops_to_exclude = POPS_TO_REMOVE_FOR_POPMAX
```
mike-w-wilson, line 46: Lets move this into the function that is using it, so if a parameter exists, default to `POPS_TO_REMOVE_FOR_POPMAX`, if it doesn't yet, create it.

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -25,17 +25,94 @@
...
+
+
+# TODO: not sure if we should put this function in gnomad.utils.annotations
```
mike-w-wilson, line 49: Lets do that since it seems like Kristen is already using some form of this too. Please put me on the PR.

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -81,23 +159,55 @@ def _get_pop_meta_indices(
...
         """
         meta = hl.eval(meta)
+        # for adj and raw, but raw is not in faf_meta for genomes release HT,
+        # so we will remove the duplicate idx with list(set(idx))
         idx = [0, 1]
         idx.extend(
             [i for i, m in enumerate(meta) if m.get("pop") in pop_list and len(m) == 2]
         )
-        meta = [meta[i] for i in idx]
+        meta = [meta[i] for i in list(set(idx))]
```
mike-w-wilson, line 168: revert once drop the 1

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -81,23 +159,55 @@ def _get_pop_meta_indices(
...
+    # TODO: remove this once v4 genomes freq_ht is finalized
+    if "popmax" in ht.row.keys():
+        ht = ht.transmute(grpmax=ht.popmax)
+
     freq_idx, freq_meta = _get_pop_meta_indices(ht.freq_meta, pops)
     faf_idx, faf_meta = _get_pop_meta_indices(ht.faf_meta, faf_pops)
 
+    # Compute FAF max (fafmax)
+    fafmax = pop_max_for_faf_expr(ht.faf, ht.faf_meta)
+    ht = ht.annotate(fafmax=fafmax)
```
mike-w-wilson, line 203: Just confirming this was discussed and needed in the combined FAF? If so maybe just shorten it to a single line, no need to assign fafmax before the annotation.

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -143,17 +253,24 @@ def get_joint_freq_and_faf(
...
+    grpmax = grpmax.annotate(faf95=faf[faf_meta_by_pop.get(grpmax.pop)].faf95)
 
     # Annotate Table with all joint exomes + genomes computations.
     ht = ht.annotate(joint_freq=freq, joint_faf=faf, joint_grpmax=grpmax)
+
     ht = ht.annotate_globals(
         joint_freq_meta=freq_meta,
-        joint_freq_index_dict=make_freq_index_dict(freq_meta, label_delimiter="-"),
+        joint_freq_index_dict=make_freq_index_dict_from_meta(
+            freq_meta, label_delimiter="-"
```
mike-w-wilson, line 264: should be underscore, which is the default

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -143,17 +253,24 @@ def get_joint_freq_and_faf(
...
+        ),
         joint_faf_meta=faf_meta,
         joint_faf_index_dict=make_faf_index_dict(faf_meta, label_delimiter="-"),
     )
 
+    # Compute FAF max (fafmax) on the merged exomes + genomes frequencies.
+    fafmax = pop_max_for_faf_expr(ht.joint_faf, ht.joint_faf_meta)
```
mike-w-wilson, line 271: If were keeping this in, move to L260 and then select it with everything else

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -326,25 +446,46 @@ def get_combine_faf_resources(
...
+    )
+
+
 def main(args):
     """Create combined FAF resource."""
     hl.init(log="/compute_combined_faf.log")
     test = args.test
     overwrite = args.overwrite
-    pops = list(POPS["v3"] & POPS["v4"])
+    pops = POPS["v3"] + POPS["v4"]
```
mike-w-wilson, line 467:
```suggestion
    pops = list(set(POPS["v3"] + POPS["v4"]))
```
The original code assumed POPS[version] returned a set, this creates a list of unique values.

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -326,25 +446,46 @@ def get_combine_faf_resources(
...
-            exomes_ht = extract_freq_info(exomes_ht, pops, faf_pops, "genomes")
-            genomes_ht = extract_freq_info(genomes_ht, pops, faf_pops, "exomes")
+            # replace oth with remaining in freq_meta and freq_index_dict for genomes_ht
+            genomes_ht = replace_oth_with_remaining(genomes_ht)
+            if test:
+                # exomes_ht = res.exomes_freq_ht.ht()._filter_partitions(range(20))
+                # genomes_ht = res.genomes_freq_ht.ht()._filter_partitions(range(20))
+                # filter to PCSK9 1:55039447-55064852 for testing
+                exomes_ht = filter_gene_to_test(exomes_ht)
+                genomes_ht = filter_gene_to_test(genomes_ht)
```
mike-w-wilson, line 485: Generally, the test should come before any other work is done so you reduce the amount of data your handling immediately. I know the replace_oth_with_remaining is working on freq globals but its best practice to filter to test things once you read your tables in. It would be better to keep the test and filter partitions and then add in another command line arg that will filter to a gene so we can test in different ways.

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -353,6 +494,7 @@ def main(args):
             res.check_resource_existence()
             ht = res.freq_ht.ht()
             ht = ht.select(
+                # TODO: this step is very slow, need to optimize
```
mike-w-wilson, line 497: Maybe include how slow so if you're not the one to pick up the work others have a reference.

gnomad_qc/v4/resources/release.py
```diff
@@ -79,13 +79,13 @@ def combined_faf_release_ht_path(
...
-        else:
-            return f"gs://gnomad-public-requester-pays/release/{release_version}/ht/joint/gnomad.joint.v{release_version}.faf.ht"
-    else:
-        return f"gs://gnomad/release/{release_version}/ht/joint/gnomad.joint.v{release_version}.faf.ht"
+    # if public:
+    #     if file_exists(combined_faf().versions[release_version].path):
+    #         return combined_faf().versions[release_version].path
+    #     else:
+    #         return f"gs://gnomad-public-requester-pays/release/{release_version}/ht/joint/gnomad.joint.v{release_version}.faf.ht"
+    # else:
```
mike-w-wilson, line 87: This should be entirely deleted if we aren't using it but why is it being removed?

mike-w-wilson (commented):
Back to you Qin!

gnomad_qc/v4/resources/release.py
KoalaQin: because Julia said we will put combined_faf resources later in gnomAD_methods. Can I just quote it for later use?

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
jkgoodrich: Also just note that it depends significantly on the `min_cell_count` value

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
KoalaQin: I had problem to move it to L260 with others, I will test it again once the fafmax PS is merged in gnomad_methods.

KoalaQin: I found out why, faf_meta from `faf_expr()` is a List, not ArrayExpression, I added hl.literal(faf_meta).

PR #447, opened 2023-09-26
Finalize freq HT

mike-w-wilson:
This finalizes the freq HT. It renames the original and  adjusted freq annotations. It also drops any 'gatk_version' index from freq_meta, freq, high_ab_hets_by_group, pre_adjusted_freq, and freq_meta_sample_count. For now this stores the pre_adjusted freq but I'm wondering if we need to here since we currently have the preadjustment HT. If we want all the information in one place, we would need to store the preadjustment histograms as well so I will need to add that in. Alternatively, if we plan on keeping the full preadjustment HT around, I'm not sure we need the predajustment/high AB hets on this table but it was in the functions docstring todo placeholder so I left it for now.

### DIFF
--- a/gnomad_qc/v4/annotations/generate_freq.py
+++ b/gnomad_qc/v4/annotations/generate_freq.py
@@ -909,10 +909,52 @@ def create_final_freq_ht(ht: hl.Table) -> hl.Table:
     Filter the following from the 'freq' field:
         - gatk versions
 
+    .. note::
+
+        `pre_adjustment_freq` is the frequency annotation before any high AB het
+        adjustment. This value will be the same as freq for sites that have an AF below
+        the AF adjustment threshold.
+
     :param ht: Hail Table containing all annotations.
     :return: Hail Table with only desired annotations.
     """
-    raise NotImplementedError("Creation of final freq Table is not implemented yet.")
+    logger.info("Dropping gatk_version from freq ht array annotations...")
+    freq_meta, array_exprs = filter_arrays_by_meta(
+        ht.freq_meta,
+        {
+            "freq": ht.ab_adjusted_freq,
+            "freq_meta_sample_count": ht.index_globals().freq_meta_sample_count,
+        },
+        items_to_filter=["gatk_version"],
+        keep=False,
+    )
+
+    ht = ht.select(
+        freq=array_exprs["freq"],
+        faf=ht.faf,
+        grpmax=ht.grpmax,
+        inbreeding_coeff=ht.InbreedingCoeff,
+        histograms=hl.struct(
+            **{
+                "qual_hists": ht.ab_adjusted_qual_hists,
+                "raw_qual_hists": ht.ab_adjusted_raw_qual_hists,
+                "age_hists": ht.ab_adjusted_age_hists,
+            },
+        ),
+        pre_adjustment_freq=ht.freq,
+        high_ab_hets_by_group=ht.high_ab_hets_by_group,
+    )
+
+    ht = ht.select_globals(
+        downsamplings=ht.index_globals().downsamplings,
+        freq_meta=freq_meta,
+        freq_index_dict=make_freq_index_dict_from_meta(freq_meta),
+        freq_meta_sample_count=array_exprs["freq_meta_sample_count"],
+        faf_meta=ht.index_globals().faf_meta,
+        faf_index_dict=ht.index_globals().faf_index_dict,
+    )
+
+    return ht
 
 
 def main(args):
--- a/gnomad_qc/v4/annotations/generate_freq.py
+++ b/gnomad_qc/v4/annotations/generate_freq.py
@@ -899,16 +899,6 @@ def create_final_freq_ht(ht: hl.Table) -> hl.Table:
     """
     Create final freq Table with only desired annotations.
 
-    Drop the following annotations:
-        - 'age_high_ab_hist', all annotations from the split VDSs....
-
-    Rename the following annotations:
-        - 'ab_adjusted_freq' -> 'freq'
-        - decide if we want to store uncorrected? Probably,just rename it.
-
-    Filter the following from the 'freq' field:
-        - gatk versions
-
     .. note::
 
         `pre_adjustment_freq` is the frequency annotation before any high AB het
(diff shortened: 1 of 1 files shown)


mike-w-wilson:
I didnt change `pop` -> `ga_grp` here because in our slack conversation with KC we decided we should do a single large conversion and we probably still need to run this change past a larger group

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -899,20 +889,55 @@ def create_final_freq_ht(ht: hl.Table) -> hl.Table:
...
+            "pre_adjustment_freq": ht.freq,
+        },
+        items_to_filter=["gatk_version"],
+        keep=False,
+    )
+
+    ht = ht.select(
+        freq=array_exprs["freq"],
+        faf=ht.faf,
+        grpmax=ht.grpmax,
```
jkgoodrich, line 917: Just confirming that we are not changing `pop` here:
```
    'grpmax': struct {
        gnomad: struct {
            AC: int32, 
            AF: float64, 
            AN: int32, 
            homozygote_count: int64, 
            pop: str, 
            faf95: float64
        }, 
        non_ukb: struct {
            AC: int32, 
            AF: float64, 
            AN: int32, 
            homozygote_count: int64, 
            pop: str, 
            faf95: float64
        }
    } 
```
Can you ping in the gnomad channel about changing `pop` -> `ga_grp` and maybe make a poll with a few that we came up with as options?

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -899,20 +889,55 @@ def create_final_freq_ht(ht: hl.Table) -> hl.Table:
...
-    Rename the following annotations:
-        - 'ab_adjusted_freq' -> 'freq'
-        - decide if we want to store uncorrected? Probably,just rename it.
+    .. note::
 
-    Filter the following from the 'freq' field:
-        - gatk versions
+        `pre_adjustment_freq` is the frequency annotation before any high AB het
+        adjustment. This value will be the same as freq for sites that have an AF below
+        the AF adjustment threshold.
```
jkgoodrich, line 896: Don't need this if we drop that from this table
```suggestion
```

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -899,20 +889,55 @@ def create_final_freq_ht(ht: hl.Table) -> hl.Table:
     """
     Create final freq Table with only desired annotations.
 
-    Drop the following annotations:
-        - 'age_high_ab_hist', all annotations from the split VDSs....
-
-    Rename the following annotations:
-        - 'ab_adjusted_freq' -> 'freq'
-        - decide if we want to store uncorrected? Probably,just rename it.
+    .. note::
```
jkgoodrich, line 893: Don't need this if we drop that from this table
```suggestion
```

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -899,20 +889,55 @@ def create_final_freq_ht(ht: hl.Table) -> hl.Table:
...
+        inbreeding_coeff=ht.InbreedingCoeff,
+        histograms=hl.struct(
+            **{
+                "qual_hists": ht.ab_adjusted_qual_hists,
+                "raw_qual_hists": ht.ab_adjusted_raw_qual_hists,
+                "age_hists": ht.ab_adjusted_age_hists,
+            },
+        ),
+        pre_adjustment_freq=array_exprs["pre_adjustment_freq"],
+        high_ab_hets_by_group=array_exprs["high_ab_hets_by_group"],
```
jkgoodrich, line 927: I agree we should just drop this, I would like to keep both tables around

```suggestion
```

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -899,20 +889,55 @@ def create_final_freq_ht(ht: hl.Table) -> hl.Table:
...
+
+    ht = ht.select_globals(
+        downsamplings=ht.index_globals().downsamplings,
+        freq_meta=freq_meta,
+        freq_index_dict=make_freq_index_dict_from_meta(freq_meta),
+        freq_meta_sample_count=array_exprs["freq_meta_sample_count"],
+        faf_meta=ht.index_globals().faf_meta,
+        faf_index_dict=ht.index_globals().faf_index_dict,
+        age_distribution=ht.index_globals().age_distribution,
+    )
```
jkgoodrich, line 938: Slight preference for this, but don't really care:

```suggestion
    g = ht.index_globals()
    ht = ht.select_globals(
        downsamplings=g.downsamplings,
        freq_meta=freq_meta,
        freq_index_dict=make_freq_index_dict_from_meta(freq_meta),
        freq_meta_sample_count=array_exprs["freq_meta_sample_count"],
        faf_meta=g.faf_meta,
        faf_index_dict=g.faf_index_dict,
        age_distribution=g.age_distribution,
    )
```

gnomad_qc/v4/annotations/generate_freq.py
```diff
@@ -1039,6 +1064,9 @@ def main(args):
 
         if args.finalize_freq_ht:
             logger.info("Writing final frequency Table...")
+            res = resources.finalize_freq_ht
+            res.check_resource_existence()
+            ht = res.corrected_freq_ht.ht()
             ht = create_final_freq_ht(ht)
```
jkgoodrich, line 1070:
```suggestion
            ht = create_final_freq_ht(res.corrected_freq_ht.ht())
```

gnomad_qc/v4/annotations/generate_freq.py
mike-w-wilson: I wasn't going to originally but I think this is the only place I can think of that will have "pop" in the main release HT outside of the globals so now I'm leaning towards just changing it here. I polled the group here and `gen_anc` won.

mike-w-wilson:
Back to you @jkgoodrich ! I updated `pop`-> `gen_anc` in `grpmax` on L849 in the original annotation. This was way cleaner than messing with the nested structs in the finalize bit but I'll need to rerun the correct step (which runs in minutes). I'll just run both the correction and finalize whenever were ready though.

jkgoodrich (approved):
LGTM!

PR #452, opened 2023-10-03
Fix the NA bug in frequency adjustments

mike-w-wilson:
For any site with an adj AF of NA, the script was setting all fields to NA. This came down to the hl.if_else statement on the AF threshold -- If the conditional in hl.if_else is missing (AF[0] is often NA), everything got set to NA which was very unexpected but I guess NA is just that. I'll rerun the combine, adjustment, and finalize step of freq once this and the [merge_histogram bug](https://github.com/broadinstitute/gnomad_methods/pull/614) go in.


jkgoodrich (commented):
would `missing_false=True` work?

mike-w-wilson:
Ah-- yes it does. Thank you!

jkgoodrich (approved):
LGTM!

PR #453, opened 2023-10-04
Final variant QC evaluation PR

jkgoodrich:
This is just the branch that we were merging the other PRs into, but now it is for the final merge of the full script into main. Just take a final quick scan over everything to make sure it looks OK


matren395 (approved):
if this was what I'd already reviewed then lgtm!

PR #454, opened 2023-10-04
Add UKB/platform stratification option to v4 coverage

jkgoodrich:
Depends on gnomad_methods PR https://github.com/broadinstitute/gnomad_methods/pull/615

Should probably change the output somehow so we don't overwrite the one we already have until we confirm everything looks OK. 

Thoughts on the best thing to do for that?

### DIFF
--- a/gnomad_qc/v4/annotations/compute_coverage.py
+++ b/gnomad_qc/v4/annotations/compute_coverage.py
@@ -123,7 +123,7 @@ def main(args):
                 meta_expr = vds.variant_data.meta
                 ukb_expr = meta_expr.project_meta.ukb_sample
                 strata = [
-                    {"subset": hl.if_else(ukb_expr, "ukb", "non_ukb")},
+                    {"ukb_strata": hl.if_else(ukb_expr, "ukb", "non_ukb")},
                     {
                         "subset": hl.or_missing(~ukb_expr, "non_ukb"),
                         "platform": meta_expr.platform_inference.qc_platform,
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/annotations/compute_coverage.py
```diff
@@ -116,12 +116,43 @@ def main(args):
...
+            )
 
             # Checkpoint Table.
+            if args.stratify_by_ukb_and_platform:
+                coverage_ht = coverage_ht.annotate_globals(
+                    coverage_stats_meta=coverage_ht.coverage_stats_meta.map(
+                        lambda x: hl.dict(
+                            x.items().map(
+                                lambda m: hl.if_else(
+                                    m[0] == "ukb_strata", ("subset", m[1]), m
```
mike-w-wilson, line 150: Any reason to not name this subset originally?

mike-w-wilson (approved):
One question but LGTM! For writing out the stratified elsewhere, can you do something like we did with the freq resources for split  nonukb and ukb freq hts? IMO it can overwrite the exisiting table since even the stratified  option will have the coverage annotations for the entire callset, just in a different schema.

klaricch:
the coverage ht shouldn't be that big, so we can also put a copy of it in the 30 day temp bucket before overwritting?

gnomad_qc/v4/annotations/compute_coverage.py
```diff
@@ -116,12 +116,43 @@ def main(args):
...
+                                )
+                            )
+                        )
+                    )
+                )
             coverage_ht = coverage_ht.checkpoint(
                 hl.utils.new_temp_file("coverage", extension="ht")
             )
```
klaricch, line 158: should we remove all the stratifications for the export_coverage_tsv step so that it doesn't get too large, and if anyone wants them they can just use the ht?

gnomad_qc/v4/annotations/compute_coverage.py
jkgoodrich: yeah, the way the group annotation code works is that is expects the strata with the same name to have the same values, but my first one has "ukb" and "non_ukb", but the second one only has non_ukb, this was a fast way to get the results needed

jkgoodrich:
> the coverage ht shouldn't be that big, so we can also put a copy of it in the 30 day temp bucket before overwritting?

I like this idea, I will go with that

klaricch (approved):
lgtm

mike-w-wilson (approved):

PR #455, opened 2023-10-05
Add script to create the final variant QC HT

jkgoodrich:
Still a few TODOs relating to VQSR models, but ready for the first pass of review in the meantime.

There is not really any difficult logic in this script, it's mostly just rearranging/renaming annotations and adding the filters and a couple flags.

### DIFF
--- a/gnomad_qc/v4/variant_qc/final_filter.py
+++ b/gnomad_qc/v4/variant_qc/final_filter.py
@@ -42,7 +42,7 @@ TRAINING_INFO_FIELDS = {
         "positive_train_site",
         "negative_train_site",
     ],
-    "AS_VQSR": ["POSITIVE_TRAIN_SITE", "NEGATIVE_TRAIN_SITE"],
+    "AS_VQSR": ["positive_train_site", "negative_train_site"],
 }
 """Annotations to keep in the 'training_info' field of the final filter Table."""
 VARIANT_QC_RESULT_FIELDS = {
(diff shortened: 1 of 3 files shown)


gnomad_qc/v4/variant_qc/final_filter.py
```diff
@@ -0,0 +1,588 @@
...
+        "rf_label",
+        "positive_train_site",
+        "negative_train_site",
+    ],
+    "AS_VQSR": ["positive_train_site", "negative_train_site"],
+}
+"""Annotations to keep in the 'training_info' field of the final filter Table."""
+VARIANT_QC_RESULT_FIELDS = {
+    "RF": ["rf_tp_probability", "rf_prediction"],
+    "AS_VQSR": ["AS_VQSLOD", "AS_culprit"],
```
matren395, line 50: do we want to add anything for IF_VQSR?

gnomad_qc/v4/variant_qc/final_filter.py
```diff
@@ -0,0 +1,588 @@
...
+        variant_qc_globals = variant_qc_globals.annotate(
+            feature_medians=variant_qc_globals.feature_medians.map_values(
+                lambda x: x[coumpute_info_method]
+            )
+        )
+        variant_qc_globals = variant_qc_globals.select(
+            rf_globals=variant_qc_globals.select(*VARIANT_QC_GLOBAL_FIELDS["RF"])
+        )
+    else:
+        # TODO: Need to add a compute_info_method global in the vqsr loading script,
```
matren395, line 298: will that be a part of this PR, or a separate one ?

matren395 (commented):
will give a closer review later! looks like this needs something more than a quick glance and approving , so I will come back to it when I'm a bit more free

matren395 (approved):
lgtm! tests worked and outputs look correct

PR #460, opened 2023-10-06
Small change to work with updated `generate_freq_group_membership_array`

jkgoodrich:
I made `generate_freq_group_membership_array` return a place holder for the raw genotype because it was too confusing having it be a different length than the globals. I know this might not be run again ever, but wanted to make sure it works


mike-w-wilson (approved):
Never say never!

mike-w-wilson:
errr...ever?

PR #461, opened 2023-10-06
Change return type of vep in silico predictors

mike-w-wilson:
The tuple expression this returned was difficult to use on any table. This simplifies it by just returning a table with the max scores. I think this behavior is more in line with what we expect out of these modules.


KoalaQin (approved):
LGTM!

PR #462, opened 2023-10-06
add function to create Phylop score

KoalaQin:
This was already run, so I don't have the raw bigwig or bedgraph anymore, I can redownload and convert it later, so with this code, you can't run a test, but the result file is already transferred to the bucket as other in silico predictors.

### DIFF
--- a/gnomad_qc/v4/annotations/insilico_predictors.py
+++ b/gnomad_qc/v4/annotations/insilico_predictors.py
@@ -439,6 +439,40 @@ def create_revel_grch38_ht() -> hl.Table:
     return final_ht
 
 
+def create_phylop_grch38_ht() -> hl.Table:
+    """
+    Convert PhyloP scores to Hail Table.
+
+    .. note::
+       BigWig format of Phylop from here:
+       https://cgl.gi.ucsc.edu/data/cactus/241-mammalian-2020v2-hub/Homo_sapiens/241-mammalian-2020v2.bigWig
+       And convert
+       Converted it to bedGraph format with bigWigToBedGraph from the kent packages
+       of UCSC (https://hgdownload.cse.ucsc.edu/admin/exe/) with the following command:
+       `./bigWigToBedGraph ~/Downloads/241-mammalian-2020v2.bigWig ~/Downloads/241-mammalian-2020v2.bedGraph`
+       The bedGraph file is bigzipped before import to Hail.
+       Different to other in silico predictors, Phylop HT is not keyed by locus &
+       alleles, we have a score per base, so we won't have any score for deletions.
+
+    :return: Hail Table with Phylop Scores for GRCh38
+    """
+    bg_path = "gs://gnomad-insilico/phylop/Human-GRCh38-Phylop-241-mammalian-2020v2.bedGraph.bgz"
+    columns = ["chr", "start", "end", "phylop"]
+    ht = hl.import_table(
+        bg_path,
+        min_partitions=1000,
+        impute=True,
+        no_header=True,
+    ).rename({f"f{i}": c for i, c in enumerate(columns)})
+
+    ht = ht.annotate(pos=hl.range(ht.start + 1, ht.end + 1))
+    ht = ht.explode("pos")
+    ht = ht.annotate(locus=hl.locus(ht.chr, ht.pos, reference_genome="GRCh38"))
+    ht = ht.select("locus", "phylop")
+
+    return ht
+
+
 def main(args):
     """Generate Hail Tables with in silico predictors."""
     hl.init(
@@ -485,6 +519,13 @@ def main(args):
             overwrite=args.overwrite,
         )
         logger.info("REVEL Hail Table for GRCh38 created.")
+    if args.phylop:
+        logger.info("Creating PhyloP Hail Table for GRCh38...")
+        ht = create_phylop_grch38_ht()
+        ht.write(
+            get_insilico_predictors(predictor="phylop").path,
+            overwrite=args.overwrite,
+        )
 
 
 if __name__ == "__main__":
@@ -497,6 +538,7 @@ if __name__ == "__main__":
     parser.add_argument("--spliceai", help="Create SpliceAI HT", action="store_true")
     parser.add_argument("--pangolin", help="Create Pangolin HT", action="store_true")
     parser.add_argument("--revel", help="Create REVEL HT.", action="store_true")
+    parser.add_argument("--phylop", help="Create PhyloP HT.", action="store_true")
     args = parser.parse_args()
     if args.slack_channel:
         with slack_notifications(slack_token, args.slack_channel):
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -439,6 +439,39 @@ def create_revel_grch38_ht() -> hl.Table:
...
+    bg_path = "gs://gnomad-insilico/phylop/Human-GRCh38-Phylop-241-mammalian-2020v2.bedGraph.bgz"
+    columns = ["chr", "start", "end", "phylop"]
+    ht = hl.import_table(
+        bg_path,
+        min_partitions=1000,
+        impute=True,
+        no_header=True,
+    ).rename({f"f{i}": c for i, c in enumerate(columns)})
+
+    ht = ht.annotate(pos=hl.range(ht.start + 1, ht.end + 1))
```
mike-w-wilson, line 467: Why is 1 added to start?

KoalaQin: Because this bed format is 0-based instead of 1-based, I figured that for our locus, we need start and end. Should I put a note?

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -439,6 +439,39 @@ def create_revel_grch38_ht() -> hl.Table:
...
+    ht = ht.explode("pos")
+    ht = ht.annotate(locus=hl.locus(ht.chr, ht.pos, reference_genome="GRCh38"))
+    ht = ht.select("locus", "phylop")
```
mike-w-wilson, line 470: This whole block code be 
```suggestion
    ht = hl.import_bed(bg_path, min_partitions=1000)
    ht = ht.rename({"target":"phylop"})
```
This will key by interval. Since we only have locus anyways, we will need to join by ht.locus, not ht.key. Hail will handle the logic of if a locus is in an interval so we can just keep the file as intervals. This is from [their docs on import_bed](https://hail.is/docs/0.2/methods/impex.html#hail.methods.import_bed).
```
bed = hl.import_bed('data/file2.bed')
result = dataset.annotate_rows(cnv_id = bed[dataset.locus].target)
```

mike-w-wilson (commented):
I think this needs to be regenerated because of hl.range(ht.start+1, ht.end+1), hl.range will is end exclusive so I think we're losing sites with the current logic.

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -439,6 +439,39 @@ def create_revel_grch38_ht() -> hl.Table:
...
+    `./bigWigToBedGraph ~/Downloads/241-mammalian-2020v2.bigWig ~/Downloads/241-mammalian-2020v2.bedGraph`
+    The bedGraph file is bigzipped before importing to Hail.
+    Different to other in silico predictors, Phylop HT is not keyed by locus
+    & alleles, we have a score per base pair, so we won't have any score for deletions.
+
+    :return: Hail Table with Phylop Scores for GRCh38
+    """
+    bg_path = "gs://gnomad-insilico/phylop/Human-GRCh38-Phylop-241-mammalian-2020v2.bedGraph.bgz"
+    columns = ["chr", "start", "end", "phylop"]
+    ht = hl.import_table(
```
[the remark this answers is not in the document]
KoalaQin: oh, I didn't link bedgraph to the bed in Hail. I'm redownloading the raw file and process it.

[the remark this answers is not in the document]
KoalaQin: I will put the raw file in that path.

gnomad_qc/v4/annotations/insilico_predictors.py
mike-w-wilson: I see that the file intervals are also end-exclusive so this is fine.

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -439,6 +439,39 @@ def create_revel_grch38_ht() -> hl.Table:
...
+
+    .. note::
+    BigWig format of Phylop was download from here:
+    https://cgl.gi.ucsc.edu/data/cactus/241-mammalian-2020v2-hub/Homo_sapiens/241-mammalian-2020v2.bigWig
+    and converted it to bedGraph format with bigWigToBedGraph from the kent packages
+    of UCSC (https://hgdownload.cse.ucsc.edu/admin/exe/) with the following command:
+    `./bigWigToBedGraph ~/Downloads/241-mammalian-2020v2.bigWig ~/Downloads/241-mammalian-2020v2.bedGraph`
+    The bedGraph file is bigzipped before importing to Hail.
+    Different to other in silico predictors, Phylop HT is not keyed by locus
+    & alleles, we have a score per base pair, so we won't have any score for deletions.
```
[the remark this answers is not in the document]
mike-w-wilson: Responded

gnomad_qc/v4/annotations/insilico_predictors.py
mike-w-wilson, line 467: Yeah maybe just add a note saying its 0 indexed and end exclusive.

gnomad_qc/v4/annotations/insilico_predictors.py
mike-w-wilson: Qin shared a slack thread with me where KC and her decided to explode the intervals.

gnomad_qc/v4/annotations/insilico_predictors.py
mike-w-wilson, line 454:
```suggestion
    Different than other in silico predictors, the Phylop HT is keyed by locus only. The HT has a 
    score per position and there are no scores for deletions.
```

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -439,6 +439,42 @@ def create_revel_grch38_ht() -> hl.Table:
...
+    columns = ["chr", "start", "end", "phylop"]
+    ht = hl.import_table(
+        bg_path,
+        min_partitions=1000,
+        impute=True,
+        no_header=True,
+    ).rename({f"f{i}": c for i, c in enumerate(columns)})
+
+    # We add 1 to both start and end because input bedGraph is 0-indexed interval and
+    # hl.range is end exclusive
```
mike-w-wilson, line 470:
```suggestion
    # We add 1 to both start and end because input bedGraph is 0-indexed interval and
    # the interval is end exclusive
```
Sorry I meant the interval

mike-w-wilson (approved):
Just one clarifying comment

gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -452,7 +452,8 @@ def create_phylop_grch38_ht() -> hl.Table:
     The bedGraph file is bigzipped before importing to Hail.
     Different to other in silico predictors, the Phylop HT is keyed by locus only. Since
      the PhyloP scores have one value per position, we exploded the interval to store
-     the HT by locus.
+     the HT by locus. In result, we have Phylop scores for 2,852,623,265 locus from
```
mike-w-wilson, line 455: 👍

KoalaQin:
BTW, I regenerated the raw bedGraph and transferred it to the path, like for other in silico predictors, in case we need them in the future.

PR #390, opened 2023-08-09
update HGDP+TGP AF in v4 release HT with updated sample QC and pop names

KoalaQin:
This script includes the following functions:
1. add updated sample QC annotations; 
2. update the HGDP pop labels;
3. filter `samples_to_add` and `samples_to_subtract`; 
4. calculate AF with the old & new pop labels, for the whole set or subsets; 
5. concatenate subset frequencies; 
Along with the changes in some resources path.

PR #465, opened 2023-10-08
HGDP + 1KG get call stats for updated samples for the v4 genomes sites HT

jkgoodrich:
Functions copied from this PR: https://github.com/broadinstitute/gnomad_qc/pull/390

This contains the functions required to identify samples that need call stats computed for an update to the v4 genomes sites HT

The first commit is basically a copy from the above PR, and any commits after that are my suggested changes.


gnomad_qc/v4/resources/annotations.py
```diff
@@ -407,6 +413,32 @@ def get_vrs(
...
+
+    :param test: If True, will return the annotation resource for testing purposes.
+    :param subset: The subset of the HGDP + 1KG/TGP release to return,
+       must be "added", "subtracted", "pop_diff", or "final"
+    :return: MatrixTableResource for specified subset.
+    """
+    # if subset not in ["added", "subtracted", "pop_diff", "final"]:
+    #    raise ValueError(
+    #        "Operation must be one of 'added', 'subtracted', 'pop_diff', or 'final'"
+    #    )
```
jkgoodrich, line 430: Sorry, just noticed I have this commented out 
```suggestion
    if subset not in ["added", "subtracted", "pop_diff", "final"]:
        raise ValueError(
            "Operation must be one of 'added', 'subtracted', 'pop_diff', or 'final'"
        )
```

KoalaQin (changes requested):
I only spot a small formatting issue. I went through this twice and tested on drd2 and X gene. Both worked perfectly.

gnomad_qc/v4/resources/annotations.py
```diff
@@ -407,6 +413,32 @@ def get_vrs(
...
+
+    :param test: If True, will return the annotation resource for testing purposes.
+    :param subset: The subset of the HGDP + 1KG/TGP release to return,
+       must be "added", "subtracted", "pop_diff", or "final"
+    :return: MatrixTableResource for specified subset.
+    """
+    if subset not in ["added", "subtracted", "pop_diff", "final"]:
+        raise ValueError(
+            "Operation must be one of 'added', 'subtracted', 'pop_diff', or 'final'"
+        )
```
KoalaQin, line 430:
```suggestion
    if subset not in ["added", "subtracted", "pop_diff", "final"]
        raise ValueError(
            "Operation must be one of 'added', 'subtracted', 'pop_diff', or 'join'"
        )
```
The `final` has to be changed to using `get_freq`, right?

KoalaQin (changes requested):
one more thing for the path of the joint freq_ht.

gnomad_qc/v4/resources/annotations.py
jkgoodrich: yup, thanks!

KoalaQin (approved):
LGTM!

PR #467, opened 2023-10-09
Improve CADD run by adding new_temp_file

KoalaQin:
The initial run failed, so I added this to make it faster.

### DIFF
--- a/gnomad_qc/v4/annotations/insilico_predictors.py
+++ b/gnomad_qc/v4/annotations/insilico_predictors.py
@@ -76,9 +76,9 @@ def create_cadd_grch38_ht() -> hl.Table:
           that are new in gnomAD v4 genomes because of the addition of HGDP/TGP samples.
 
          .. note::
-         1,972,208 indels were duplicated in gnomAD v3.0 and v4.0 or in gnomAD
-         v3.1 and v4.0. However, CADD only generates a score per loci.
-         We keep only the latest prediction, v4.0, for these loci.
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/annotations/insilico_predictors.py
```diff
@@ -143,7 +144,6 @@ def _load_cadd_raw(cadd_tsv) -> hl.Table:
     logger.info("Number of indels in v3 and not in v4: %s.", indel3.count())
```
jkgoodrich, line 144: You removed the other loggers, do you still want this one? Also, why this one, but not the others?

jkgoodrich (commented):
Looks good, just a quick question

gnomad_qc/v4/annotations/insilico_predictors.py
KoalaQin: I didn't know logging wasn't working, I removed others to make it faster, and this number is the only one I cared about because I put other numbers in the dosstring and I could add them up.

jkgoodrich: Do you still want the logger in there now that it's in the docstring?

KoalaQin: This one is not, I want it but it's not working and it's already run, I will just remove it.

KoalaQin: I kept 2 loggers because I kind of need them to monitor the duplicates.

jkgoodrich (approved):
LGTM!

PR #469, opened 2023-10-09
Suggestions to v4 exomes release HT branch

jkgoodrich:


gnomad_qc/v4/create_release/create_release_sites_ht_exomes.py
```diff
@@ -492,8 +502,8 @@ def main(args):
         version=args.version,
     )
 
-    # The dbsnp table does not have a global field for dbsnp_versions, same
-    # with sift/polyphen and vrs (still need these)
+    # The dbsnp table does not have a global field for dbsnp_versions, same with
+    # sift/polyphen and vrs (still need these).
     ht = ht.annotate_globals(
```
jkgoodrich, line 507: Should we add the Gencode version that is associated with the vep version?

gnomad_qc/v4/create_release/create_release_sites_ht_exomes.py
```diff
@@ -38,55 +38,53 @@
...
-        #     "select": ["filters", "vqsr"],
-        #     "custom_select": custom_filters_select,
-        #     "select_globals": ["filtering_model", "inbreeding_coeff_cutoff"],
-        # },
+        "filters": {
+            "ht": final_filter().ht(),
+            "path": final_filter().path,
+            "select": ["filters"],
+            "custom_select": custom_filters_select,
+            "select_globals": ["filtering_model", "inbreeding_coeff_cutoff"],
```
jkgoodrich, line 86: filtering_model contains 'model_id' in it, but I think that is only useful internally so we should drop it from the release HT.

gnomad_qc/v4/create_release/create_release_sites_ht_exomes.py
KoalaQin: I think we should, it was even confusing to me at the beginning.

gnomad_qc/v4/create_release/create_release_sites_ht_exomes.py
```diff
@@ -492,8 +502,8 @@ def main(args):
         version=args.version,
     )
 
-    # The dbsnp table does not have a global field for dbsnp_versions, same
-    # with sift/polyphen and vrs (still need these)
+    # The dbsnp table does not have a global field for dbsnp_versions, same with
+    # sift/polyphen and vrs (still need these).
```
KoalaQin, line 506:
```suggestion
            sift_version="5.2.2",
            polyphen_version="2.2.2",
            vrs_version=hl.struct(
                **{
                    "vrs_schema_version": "1.3.0"
                    "vrs_python_version": "0.8.4",
                    "seqrepo_version": "2018-11-26",
                },
            ),
        )
    )
```
It's for the lines below. 
How can I keep the indent without using suggestion?

gnomad_qc/v4/create_release/create_release_sites_ht_exomes.py
KoalaQin: I also want to add mane_select and mane_pluls_clinical version inside vep_globals. I tried very hard to convince Matt to use the same version as VEP 105 on the browser.

mike-w-wilson: Works for me, vep has its own global struct, do we think it belongs in there or down here?

gnomad_qc/v4/create_release/create_release_sites_ht_exomes.py
```diff
@@ -147,13 +145,12 @@ def get_config(release_exists: bool = False):
         },
         "vep": {
             "ht": get_vep().ht(),
-            # TODO: drop 100% missing? Can add to the custom vep select function
             "path": get_vep().path,
             "select": ["vep"],
-            # Add in "custom_select" that drops insilicos
+            # Add in "custom_select" that drops in silicos and 100% missing VEP fields.
             "custom_select": custom_vep_select,
             # TODO: Update to have vep_csq_header -- should this be on the vep table
```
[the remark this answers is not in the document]
mike-w-wilson: Should we just add it in the VCF release code instead of here then?

gnomad_qc/v4/create_release/create_release_sites_ht_exomes.py
mike-w-wilson, line 86:
```suggestion
            # TODO: Drop model_id from filtering_model struct
            "select_globals": ["filtering_model", "inbreeding_coeff_cutoff"],
```

gnomad_qc/v4/create_release/create_release_sites_ht_exomes.py
```diff
@@ -147,13 +145,12 @@ def get_config(release_exists: bool = False):
         },
         "vep": {
             "ht": get_vep().ht(),
-            # TODO: drop 100% missing? Can add to the custom vep select function
             "path": get_vep().path,
             "select": ["vep"],
-            # Add in "custom_select" that drops insilicos
+            # Add in "custom_select" that drops in silicos and 100% missing VEP fields.
```
mike-w-wilson, line 150:
```suggestion
```
Since this is complete

gnomad_qc/v4/create_release/create_release_sites_ht_exomes.py
```diff
@@ -492,8 +502,8 @@ def main(args):
         version=args.version,
     )
 
-    # The dbsnp table does not have a global field for dbsnp_versions, same
-    # with sift/polyphen and vrs (still need these)
+    # The dbsnp table does not have a global field for dbsnp_versions, same with
+    # sift/polyphen and vrs (still need these).
     ht = ht.annotate_globals(
         tool_versions=ht.tool_versions.annotate(
```
mike-w-wilson, line 508:
```suggestion
        # TODO: Add sift_version="5.2.2",  polyphen_version="2.2.2", vrs_version=hl.struct("vrs_schema_version": "1.3.0", "vrs_python_version": "0.8.4", "seqrepo_version": "2018-11-26" to globals at FILL THIS IS points
        tool_versions=ht.tool_versions.annotate(
```

mike-w-wilson (approved):
I just have a couple questions and added TODOs for the things I'll tackle once this goes in. Thank you for fixing return types and updating docs.

mike-w-wilson:
> Are the annotations in the order we want them in?
> For instance:
> 'rsid': set
> 'filters': set
> 'was_split': bool
> 'a_index': int32
> are randomly in the middle I think.

I havent thought about the order. I feel in the past when asking about order for VCF, people generally havent had a strong opinion but we can certainly change this if we want to move things up. Maybe @ch-kr could weigh in?

> We need to decide on what interval info that we want to add to the release and where it should go? Under region_flags? We have calling intervals (both UKB and DSP) and we do also have interval QC that we can add if wanted.
Region flags or a new struct `intervals` sounds good to me. Probably also a @ch-kr question.

> I understand why we would want these internally, but I'm not positive I like having them on a release HT:
> 
> datasets=Struct(dbsnp='gs://gcp-public-data--gnomad/resources/grch38/dbsnp/dbsnp_b156_grch38_all_20221116.ht', filters='gs://gnomad/v4.0/variant_qc/exomes/gnomad.exomes.v4.0.final_filter.ht', in_silico=['gs://gnomad/v4.0/annotations/in_silico_predictors/gnomad.v4.0.spliceai.grch38.ht', 'gs://gnomad/v4.0/annotations/in_silico_predictors/gnomad.v4.0.pangolin.grch38.ht', 'gs://gnomad/v4.0/annotations/in_silico_predictors/gnomad.v4.0.revel.grch38.ht'], info='gs://gnomad/v4.0/annotations/exomes/gnomad.exomes.v4.0.info.split.ht', freq='gs://gnomad/v4.0/annotations/exomes/gnomad.exomes.v4.0.frequencies.final.ht', vep='gs://gnomad/v4.0/annotations/exomes/gnomad.exomes.v4.0.vep.ht', region_flags='gs://gnomad/v4.0/annotations/exomes/gnomad.exomes.v4.0.frequencies.final.ht')
> 
> What about having a JSON output that includes this info so we have it internally, but don't need to have it on the release HT?

That works for me.

gnomad_qc/v4/create_release/create_release_sites_ht_exomes.py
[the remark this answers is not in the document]
jkgoodrich: I think that's probably fine, but maybe check with @ch-kr to make sure she agrees

gnomad_qc/v4/create_release/create_release_sites_ht_exomes.py
mike-w-wilson: @KoalaQin, are you able to provide the versions for the tools mentioned in this thread?

KoalaQin: Yes, gencode_version="Release 39", mane_select_version="v0.95".

KoalaQin: They are coming from the same table, I think one is enough. only 56 are transcripts of 18640 are MANE Plus Clinical, it means that 56 genes have both MANE Select and MANE Plus Clinical. It seems we are not showing 'mane_pluls_clinical' on the browser, and for other in silico predictors, we only get score for the MANE Select transcripts.

KoalaQin: Or should we just use "mane_version"?

KoalaQin: The Ensembl people told me this: 
> The MANE Plus Clinical set includes additional transcripts for genes where
MANE Select alone is not sufficient to report all "Pathogenic (P)" or "Likely
Pathogenic (LP)" clinical variants available in public resources.

mike-w-wilson: hmm, I think mane_select is sufficient then. Thank you for all the context, this is helpful!

gnomad_qc/v4/create_release/create_release_sites_ht_exomes.py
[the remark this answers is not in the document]
ch-kr: this function drops fields from the VEP struct, correct? I'd be inclined to leave this here if that is the case, just to note that those are the fields we dropped in this release. however, if I read that function wrong, and this function is only related to VCF export, then I agree we don't need to keep this code here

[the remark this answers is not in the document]
mike-w-wilson: It drops the 100% missing fields, and the sift and polyphen annotations. The struct is kept and put onto the release HT.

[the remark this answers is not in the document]
mike-w-wilson: The vep_csq_header is only used in the VCF header. It's kinda unnecessary for the HT since you can do a ht.describe() to see hat fields are included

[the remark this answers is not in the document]
ch-kr: oh no, I must have misread this -- I thought you were asking whether to keep the custom vep select here. agree that vep_csq_header doesn't need to be in the release HT code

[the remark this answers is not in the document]
mike-w-wilson: Sorry misread your comment. I guess I dont understand fully what people use the header for. I would think people would want it to reflect what is in the data?

mike-w-wilson:
>I'm not super concerned about the VCF ordering, though I agree that these fields feel like they should likely be earlier in the struct. One small question since I haven't looked at the code yet: are rsid and filters going to get unnested from the info struct so that they can exported into the ID and FILTER fields?
Yes they are unnested. At the moment, this produces the schema within the v4 release doc.

gnomad_qc/v4/create_release/create_release_sites_ht_exomes.py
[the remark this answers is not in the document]
ch-kr: not at all, sorry, I misunderstood which line this comment was asking about. the vep csq header just describes which fields are in the vep field in the vcf export (and the order of the fields), so I agree that it doesn't need to get added to the release HT. for ukb, we added this field only when we were preparing to export data to vcf (so I agree it makes sense to move this to the VCF release code)

mike-w-wilson:
I'm going to merge this into my branch and make TODOs for the region flags, gencode and mane select vep globals, dropping the model ID, datsets to json, schema order, and validity checking the output.

PR #470, opened 2023-10-10
Make VRS batch code more general on input and output

KoalaQin:
I added the options to launch the batch code on either exomes or genomes vep_ht input, but as of Hail 0.2.122, Hail Batch seems still flaky to run on >100k partitions, I tried on genomes but it failed, so I changed the code a bit to use input and output paths as we want. 
I also put the versions in globals.

### DIFF
--- a/gnomad_qc/v4/annotations/vrs_annotation_batch.py
+++ b/gnomad_qc/v4/annotations/vrs_annotation_batch.py
@@ -167,18 +167,16 @@ def main(args):
 
     working_bucket = args.working_bucket
     data_type = args.data_type
-
-    input_path = v4_input_ht(data_type=data_type).path or args.input_path
-    output_vrs_path = (
-        v4_vrs_annotations(test=args.test, data_type=data_type).path
-        or args.output_vrs_path
-    )
-    output_vrs_anno_ori_path = (
-        v4_vrs_annotations(
+    if args.input_path is not None:
+        input_path = args.input_path
+        output_vrs_path = args.output_vrs_path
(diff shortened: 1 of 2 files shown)


gnomad_qc/v4/annotations/vrs_annotation_batch.py
```diff
@@ -163,8 +166,17 @@ def main(args):
     ht_example.show()
 
     working_bucket = args.working_bucket
-
-    input_path = v4_input_ht().path
+    data_type = args.data_type
+    if args.input_path is not None:
+        input_path = args.input_path
```
klaricch, line 171: should raise a ValueError here if input_path is not None but output_vrs_path or output_vrs_anno_ori_path have not been specified

gnomad_qc/v4/annotations/vrs_annotation_batch.py
```diff
@@ -401,6 +415,31 @@ def main(args):
         default=None,
         type=int,
     )
+    input_args = parser.add_mutually_exclusive_group(required=True)
+    input_args.add_argument(
+        "--data-type",
+        help="Data type to annotate (exomes or genomes).",
+        default="exomes",
```
klaricch, line 422: Since data-type is now part of a mutually exclusive group and needs to be specified to run, I don't think there needs to be a default anymore

klaricch (changes requested):

KoalaQin:
Back to you!

klaricch (approved):
lgtm

PR #488, opened 2023-10-16
Runtime change to add a checkpoint v4 genomes AN computation

jkgoodrich:


gnomad_qc/v4/annotations/generate_freq_genomes.py
```diff
@@ -808,6 +808,9 @@ def compute_an_by_group_membership(
     # Confirm that all variants in variant_filter_ht are present in the
     # vds.variant_dataset and throw an error if this is not the case since all samples
     # added to v4.0 genomes should be in this vds.
+    # TODO: This can be improved by setting split=False in get_gnomad_v3_vds and doing
```
KoalaQin, line 811: Are we going to rerun or just change this in the future?

jkgoodrich: Not going to rerun, but if others look at this code I want them to know it can be faster

KoalaQin: Okay, how about add a note at L1171 something like this?

> Please refer to 'TODO' in compute_an_by_group_membership function to set `split=True` for optimization.

KoalaQin (approved):

PR #490, opened 2023-10-17
Add data_type for combined FAF

KoalaQin:
Question: why do we use "outer" when it says "inner join"?

### DIFF
--- a/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
+++ b/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
@@ -141,7 +141,18 @@ def get_joint_freq_and_faf(
     :return: Table with joint genomes and exomes frequency and FAF information.
     """
     logger.info("Performing an inner join on frequency HTs...")
-    ht = genomes_ht.join(exomes_ht, how="outer")
+    ht = genomes_ht.join(exomes_ht, how="inner")
+
+    # Use a temp SORT_ORDER for v4 release
+    SORT_ORDER = [
+        "subset",
+        "downsampling",
+        "popmax",
+        "gen_anc",
+        "subpop",
+        "sex",
+        "group",
+    ]
 
     # Merge exomes and genomes frequencies.
     freq, freq_meta, count_arrays_dict = merge_freq_arrays(
@@ -190,9 +201,13 @@ def get_joint_freq_and_faf(
 
     ht = ht.annotate_globals(
         joint_freq_meta=freq_meta,
-        joint_freq_index_dict=make_freq_index_dict_from_meta(freq_meta),
+        joint_freq_index_dict=make_freq_index_dict_from_meta(
+            freq_meta, sort_order=SORT_ORDER
+        ),
         joint_faf_meta=faf_meta,
-        joint_faf_index_dict=make_freq_index_dict_from_meta(hl.literal(faf_meta)),
+        joint_faf_index_dict=make_freq_index_dict_from_meta(
+            hl.literal(faf_meta), sort_order=SORT_ORDER
+        ),
         joint_freq_meta_sample_count=count_arrays_dict["counts"],
     )
 
--- a/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
+++ b/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
@@ -113,6 +113,9 @@ def extract_freq_info(
     grpmax_expr = ht.grpmax
     fafmax_expr = ht.gen_anc_faf_max
     if prefix == "exomes":
+        # Note: The `grpmax` and `fafmax` structs in the exomes freq HT have two nested structs:
+        # `gnomad` and `non_ukb`. This section selects only the `gnomad` values (values across full 
+        # v4 exomes release)
         grpmax_expr = grpmax_expr.gnomad
         fafmax_expr = fafmax_expr.gnomad
 
--- a/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
+++ b/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
@@ -143,17 +143,6 @@ def get_joint_freq_and_faf(
     logger.info("Performing an outer join on frequency HTs...")
     ht = genomes_ht.join(exomes_ht, how="outer")
 
-    # Use a temp SORT_ORDER for v4 release
-    SORT_ORDER = [
-        "subset",
-        "downsampling",
-        "popmax",
-        "gen_anc",
-        "subpop",
-        "sex",
-        "group",
-    ]
-
     # Merge exomes and genomes frequencies.
     freq, freq_meta, count_arrays_dict = merge_freq_arrays(
         farrays=[ht.genomes_freq, ht.exomes_freq],
@@ -201,13 +190,9 @@ def get_joint_freq_and_faf(
 
     ht = ht.annotate_globals(
         joint_freq_meta=freq_meta,
-        joint_freq_index_dict=make_freq_index_dict_from_meta(
-            freq_meta, sort_order=SORT_ORDER
-        ),
+        joint_freq_index_dict=make_freq_index_dict_from_meta(freq_meta),
         joint_faf_meta=faf_meta,
-        joint_faf_index_dict=make_freq_index_dict_from_meta(
-            hl.literal(faf_meta), sort_order=SORT_ORDER
-        ),
+        joint_faf_index_dict=make_freq_index_dict_from_meta(hl.literal(faf_meta)),
         joint_freq_meta_sample_count=count_arrays_dict["counts"],
     )
 
--- a/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
+++ b/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
@@ -114,7 +114,7 @@ def extract_freq_info(
     fafmax_expr = ht.gen_anc_faf_max
     if prefix == "exomes":
         # Note: The `grpmax` and `fafmax` structs in the exomes freq HT have two nested structs:
-        # `gnomad` and `non_ukb`. This section selects only the `gnomad` values (values across full 
+        # `gnomad` and `non_ukb`. This section selects only the `gnomad` values (values across full
         # v4 exomes release)
         grpmax_expr = grpmax_expr.gnomad
         fafmax_expr = fafmax_expr.gnomad
--- a/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
+++ b/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
@@ -67,6 +67,8 @@ def extract_freq_info(
     The following annotations are filtered and renamed:
         - freq: {prefix}_freq
         - faf: {prefix}_faf
+        - grpmax: {prefix}_grpmax
+        - fafmax: {prefix}_fafmax
 
     The following global annotations are filtered and renamed:
         - freq_meta: {prefix}_freq_meta
@@ -112,6 +114,8 @@ def extract_freq_info(
         **{
             f"{prefix}_freq": array_exprs["freq"],
             f"{prefix}_faf": faf["faf"],
+            f"{prefix}_grpmax": ht.grpmax,
+            f"{prefix}_fafmax": ht.gen_anc_faf_max,
         }
     )
     ht = ht.select_globals(
@@ -458,10 +462,26 @@ def main(args):
                 cochran_mantel_haenszel_test=res.cmh_ht.ht()[
                     ht.key
                 ].cochran_mantel_haenszel_test,
-                data_type=hl.case()
-                .when(~hl.is_defined(ht.genomes_freq), "exomes")
-                .when(~hl.is_defined(ht.exomes_freq), "genomes")
-                .default("both"),
+                join_metric_data_type=hl.case()
+                .when(
+                    (hl.is_defined(ht.genomes_grpmax.AC))
+                    & hl.is_defined(ht.exomes_grpmax.gnomad.AC),
+                    "both",
+                )
+                .when(hl.is_defined(ht.genomes_grpmax.AC), "genomes")
+                .when(hl.is_defined(ht.exomes_grpmax.gnomad.AC), "exomes")
+                .default(hl.missing(hl.tstr)),
+                joint_fafmax=ht.joint_fafmax.annotate(
+                    join_metric_data_type=hl.case()
+                    .when(
+                        (hl.is_defined(ht.genomes_fafmax.faf95_max))
+                        & hl.is_defined(ht.exomes_fafmax.gnomad.faf95_max),
+                        "both",
+                    )
+                    .when(hl.is_defined(ht.genomes_fafmax.faf95_max), "genomes")
+                    .when(hl.is_defined(ht.exomes_fafmax.gnomad.faf95_max), "exomes")
+                    .default(hl.missing(hl.tstr)),
+                ),
             )
             ht.describe()
             ht.write(res.final_combined_faf_ht.path, overwrite=overwrite)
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -141,7 +141,18 @@ def get_joint_freq_and_faf(
     :return: Table with joint genomes and exomes frequency and FAF information.
     """
     logger.info("Performing an inner join on frequency HTs...")
-    ht = genomes_ht.join(exomes_ht, how="outer")
+    ht = genomes_ht.join(exomes_ht, how="inner")
```
mike-w-wilson, line 144: I think the logger is wrong, not the code.

KoalaQin: Why don't we get freq for variants in both exomes and genomes?

mike-w-wilson: An outer join will grab any row in either exomes or gneomes. An inner join only takes rows that appear in both.

KoalaQin: Btw, could you check my slack message? I want to find why we're getting negative values again in `merge_freq_arrays()`, Thanks!

mike-w-wilson: No worries at all -- I think it was never officially settled on so the logger being in conflict with the code is very confusing. I shouldve caught this in my original review. I'm discussing this with KC now

ch-kr: thank you both for flagging this -- Mike is right in that this wasn't completely settled. from old conversations with Grace, it sounds like we want to have an outer join here for maximum utility. 

we want to have a joint FAF number displayed for variants present in both the exomes and genomes so that users can use a single value (rather than having to decide whether to use the exomes FAF or the genomes FAF, or figure out how to combine these on their own). variants that are only present in one data type do not need a joint FAF, but for ease of use, it would make sense to present them within the same field, so that users don't have to go through this process: is joint FAF defined? -> which one of exomes or genomes FAF is defined?

also adding a note that, we should add a `data_type` field to our joint FAF structs
image

mike-w-wilson (changes requested):
Lets go with Julia suggestion in gnomad_methods

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
mike-w-wilson: We might as well since its only a logger change right now?

KoalaQin: Okay, I will send it in.

mike-w-wilson: Ok, I can also work on this if you don't have the bandwidth right now

mike-w-wilson: @ch-kr -- joint FAF is an array or structs. Do we want dataype on every entry in FAF?  Or are we really after joint_fafmax here?

KoalaQin: I made the change but I may have underestimate the complicacity?

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -458,6 +458,10 @@ def main(args):
                 cochran_mantel_haenszel_test=res.cmh_ht.ht()[
                     ht.key
                 ].cochran_mantel_haenszel_test,
+                data_type=hl.case()
+                .when(~hl.is_defined(ht.genomes_freq), "exomes")
+                .when(~hl.is_defined(ht.exomes_freq), "genomes")
```
mike-w-wilson, line 463: So we also need to make sure that the variant is not exclusive to one of the groups in `POPS_TO_REMOVE_FOR_FAF`. I think we can do this by just checking if `genomes_faf` or `exomes_faf` is defined instead of `*_freq`...I'm not totally clear where we want this flag though so double checking with @ch-kr

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
mike-w-wilson: Sorry see -- https://github.com/broadinstitute/gnomad_qc/pull/490/files#r1362611232

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
mike-w-wilson: This also makes it a top level annotation -- I would think wed want the `data_type` annotation nested somewhere for the release, or I guess we could renamed it there to something like `joint_metric_data_type` if we want it top level, that just feels confusing though.

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
ch-kr: > @ch-kr -- joint FAF is an array or structs. Do we want dataype on every entry in FAF? Or are we really after joint_fafmax here?

good point. I think just joint_fafmax needs the data type -- this is the field I expect users would want to use

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
ch-kr: >So we also need to make sure that the variant is not exclusive to one of the groups in POPS_TO_REMOVE_FOR_FAF.  I think we can do this by just checking if genomes_faf or exomes_faf is defined instead of *_freq

yes, please update to checking if faf is defined rather than just the frequency

>I'm not totally clear where we want this flag though so double checking with @ch-kr

I agree this should be nested on the release HTs (probably in the fafmax struct). for the joint FAF resource, I actually kind of like this as a top level annotation if renamed to `joint_metric_data_type`, since it explains which data type is relevant for each joint metric on that row. that said, I don't have strong feelings about this

mike-w-wilson: That makes sense to me too -- top level here as joint_metric_data_type and I'll nest it in the release joint_fafmax as `data_type`

KoalaQin: so, we want to check if genomes_faf and exomes_faf are defined, or they have non-0 values?

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -458,6 +458,10 @@ def main(args):
...
+                .default("both"),
```
mike-w-wilson, line 464: @KoalaQin, faf is not always defined even when freq is (variant is see only in remaining or a bottleneck population) so we'll want an extra case statement for `both` and then I think the default should be missing.

mike-w-wilson (changes requested):

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
jkgoodrich: sorry, just seeing this now. The inner join in the logger is because I wrote this script originally just for analysis purposes and we filtered to only variants that PASS in both datasets. I modified to be what we actually want, but in the rush I didn't update the logger

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
mike-w-wilson: This statement ^^^ is incorrect is defined but 0 for both rare variants and groups with AC =0.

KoalaQin: I just changed the code and tested it on the test gene, could you both check the naming and struct? @mike-w-wilson @ch-kr

KoalaQin:
@jkgoodrich sorry, I mis-assigned you. I need the reviews from them.

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -112,6 +114,8 @@ def extract_freq_info(
         **{
             f"{prefix}_freq": array_exprs["freq"],
             f"{prefix}_faf": faf["faf"],
+            f"{prefix}_grpmax": ht.grpmax,
+            f"{prefix}_fafmax": ht.gen_anc_faf_max,
```
ch-kr, line 118: you shouldn't need to select the subset grpmax or faf values here, right? maybe the code should have an additional if here to unnest the release-wide values from the exomes, something like
```
grpmax_expr = ht.grpmax
fafmax_expr = ht.gen_anc_faf_max
if prefix == "exomes":
    grpmax_expr = grpmax_expr.gnomad
    fafmax_expr = fafmax_expr.gnomad
```

though you shouldn't need the extra `if` if the genome freq HT has similar stratifications as the exomes HT (I was going off the schema of the current public v3 release)

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -458,6 +462,26 @@ def main(args):
...
+                joint_fafmax=ht.joint_fafmax.annotate(
+                    join_metric_data_type=hl.case()
+                    .when(
+                        (hl.is_defined(ht.genomes_fafmax.faf95_max))
+                        & hl.is_defined(ht.exomes_fafmax.gnomad.faf95_max),
+                        "both",
+                    )
+                    .when(hl.is_defined(ht.genomes_fafmax.faf95_max), "genomes")
+                    .when(hl.is_defined(ht.exomes_fafmax.gnomad.faf95_max), "exomes")
+                    .default(hl.missing(hl.tstr)),
```
ch-kr, line 483: also same question about validity checks

ch-kr (commented):
a couple questions and a name suggestion

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -458,6 +462,26 @@ def main(args):
                 cochran_mantel_haenszel_test=res.cmh_ht.ht()[
                     ht.key
                 ].cochran_mantel_haenszel_test,
+                join_metric_data_type=hl.case()
```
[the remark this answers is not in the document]
KoalaQin: yes, thanks!

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
KoalaQin: I don't have them in the joint HT if I don't select them.

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -458,6 +462,26 @@ def main(args):
...
+                .when(
+                    (hl.is_defined(ht.genomes_grpmax.AC))
+                    & hl.is_defined(ht.exomes_grpmax.gnomad.AC),
+                    "both",
+                )
+                .when(hl.is_defined(ht.genomes_grpmax.AC), "genomes")
+                .when(hl.is_defined(ht.exomes_grpmax.gnomad.AC), "exomes")
+                .default(hl.missing(hl.tstr)),
```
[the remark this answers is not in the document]
KoalaQin: There are a lot of NA already, even in my test gene. A lot of grpmax.AC is missing.

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -458,6 +462,26 @@ def main(args):
...
+                joint_fafmax=ht.joint_fafmax.annotate(
+                    join_metric_data_type=hl.case()
```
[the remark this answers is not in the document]
KoalaQin: sounds good!

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
KoalaQin: even more for fafmax, do we really a validity check?

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
KoalaQin: I thought from the release code, we will just select only the joint_*** & top level data_type from combined_faf_ht.

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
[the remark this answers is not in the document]
mike-w-wilson: `grpmax` is missing for any variant exclusive to bottleneck pop.

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
mike-w-wilson: So the `grpmax` and `fafmax` structs for exomes also contain values for the `non_ukb` subset. 
```
    'grpmax': struct {
        gnomad: struct {
            AC: int32,
            AF: float64,
            AN: int32,
            homozygote_count: int64,
            gen_anc: str
        },
        non_ukb: struct {
            AC: int32,
            AF: float64,
            AN: int32,
            homozygote_count: int64,
            gen_anc: str
        }
    }
...
    'fafmax': struct {
        gnomad: struct {
            faf95_max: float64,
            faf95_max_gen_anc: str,
            faf99_max: float64,
            faf99_max_gen_anc: str
        },
        non_ukb: struct {
            faf95_max: float64,
            faf95_max_gen_anc: str,
            faf99_max: float64,
            faf99_max_gen_anc: str
        }
    }
```
I believe KC is suggesting that you only pull `gnomad` in stead of both `gnomad` and `non_ukb` which her code example will do. As written, you will get an retain the `non_ukb` struct which we dont need on the combined FAF reference

KoalaQin: I changed it.

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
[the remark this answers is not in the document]
mike-w-wilson: It will be part of the existing `missingness` checks which just return the percent of missing values for an annotation. @ch-kr, are you suggesting instead of defaulting to missing, have a when statement that checks and then assigns/keeps missingness? and then the last bit would be `or_error`?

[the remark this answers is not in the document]
ch-kr: nope, just wanted to double check why there was an explicitly set missing case. makes sense to keep this as is given this will already be checked with the rest of the missingness checks (so we should know if there is an error)

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
ch-kr: thanks! yes, I was suggesting only pulling the `gnomad` values (values for the entire release) rather than keeping both nested structs within the grpmax and fafmax structs

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
ch-kr: the point of validity checks is to make sure nothing went awry, so we would want to know if this annotation is missing more often than expected given the missing rate of the faf annotations (and it sounds like this is taken care of based on Mike's comment in the other thread)

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -107,11 +109,20 @@ def extract_freq_info(
         combine_operator="or",
     )
 
+    # Select grpmax and fafmax
+    grpmax_expr = ht.grpmax
+    fafmax_expr = ht.gen_anc_faf_max
+    if prefix == "exomes":
+        grpmax_expr = grpmax_expr.gnomad
```
ch-kr, line 116:
```suggestion
        # Note: The `grpmax` and `fafmax` structs in the exomes freq HT have two nested structs:
        # `gnomad` and `non_ukb`. This section selects only the `gnomad` values (values across full 
        # v4 exomes release)
        grpmax_expr = grpmax_expr.gnomad
```

ch-kr (approved):
one small suggestion but LGTM

KoalaQin:
@mike-w-wilson did you request other changes? I can't see, but merging is blocked.

mike-w-wilson (approved):
LGTMT

PR #494, opened 2023-10-19
Suggestions for create genomes release HT

jkgoodrich:


mike-w-wilson (approved):
These looks good to me!

PR #497, opened 2023-10-20
Mw/custom global select

mike-w-wilson:
This adds custom global select functionality to the release code. This allows us to drop fields from global structs or make any other desired modifications, similar to custom_select_fields.

### DIFF
--- a/gnomad_qc/v4/create_release/create_release_sites_ht.py
+++ b/gnomad_qc/v4/create_release/create_release_sites_ht.py
@@ -5,7 +5,7 @@ import logging
 from copy import deepcopy
 from datetime import datetime
 from functools import reduce
-from typing import Dict, List, Optional, Union
+from typing import Dict, List, Union
 
 import hail as hl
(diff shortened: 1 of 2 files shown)


jkgoodrich (approved):
LGTM, only question is, should we do something about the temp output path? Or add a TODO if it's not worth doing now

    output_path = (
        f"{qc_temp_prefix()}release/gnomad.{data_type}.sites.test.updated_102023.ht"
        if args.test
        else release_sites(data_type=data_type).path
    )

KoalaQin:
Maybe we can put a current date in the test output name?

KoalaQin (commented):
Maybe we can put a current date in the test output name? I mean use `datetime`?

mike-w-wilson:
@jkgoodrich @KoalaQin , Yeah the temp path was something I had in the back of my mind, good chance we'll rerun this so I added in datetime (thanks for the suggestion Qin) and a datatype to the basic qc temp path.

KoalaQin (approved):
LGTM!

PR #499, opened 2023-10-20
Add datatype to initial release_path use

mike-w-wilson:
Adds datatype to the initial release run path.


jkgoodrich (approved):
thank you! lgtm

PR #500, opened 2023-10-21
Code to update v4 genome sample_qc metadata

KoalaQin:
This script is used to create the sample QC metadata HT for genomes. The main updates from v3.1 to v4.0 are:
Sample QC annotations for HGDP + TGP subset, this script directly merges the updated annotations from the subset meta HT to the full meta HT.

### DIFF
--- a/gnomad_qc/v4/resources/basics.py
+++ b/gnomad_qc/v4/resources/basics.py
@@ -385,14 +385,16 @@ def get_logging_path(name: str, version: str = CURRENT_VERSION) -> str:
     return f"{qc_temp_prefix(version)}{name}.log"
 
 
-def add_meta(mt: hl.MatrixTable, meta_name: str = "meta") -> hl.MatrixTable:
+def add_meta(
+    mt: hl.MatrixTable, version: str = CURRENT_VERSION, meta_name: str = "meta"
+) -> hl.MatrixTable:
     """
     Add metadata to MT in 'meta_name' column.
 
     :param mt: MatrixTable to which 'meta_name' annotation should be added
     :return: MatrixTable with metadata added in a 'meta' column
     """
-    mt = mt.annotate_cols(**{meta_name: meta().ht()[mt.col_key]})
+    mt = mt.annotate_cols(**{meta_name: meta(version=version).ht()[mt.col_key]})
 
     return mt
 
--- a/gnomad_qc/v4/resources/basics.py
+++ b/gnomad_qc/v4/resources/basics.py
@@ -249,7 +249,7 @@ def get_gnomad_v4_vds(
         if test:
             meta_ht = gnomad_v4_testset_meta.ht()
         else:
-            meta_ht = meta.ht()
+            meta_ht = meta().ht()
         filter_expr = meta_ht.release
         if keep_controls:
             filter_expr |= hl.literal(TRUTH_SAMPLES_S).contains(meta_ht.s)
@@ -261,7 +261,7 @@ def get_gnomad_v4_vds(
 
     if annotate_meta:
         logger.info("Annotating VDS variant_data with metadata...")
-        meta_ht = meta.ht()
+        meta_ht = meta().ht()
         vds = hl.vds.VariantDataset(
             vds.reference_data,
             vds.variant_data.annotate_cols(meta=meta_ht[vds.variant_data.col_key]),
@@ -395,7 +395,7 @@ def add_meta(
     :param version: Version of metadata ht to use for annotations
     :return: MatrixTable with metadata added in a 'meta' column
     """
-    mt = mt.annotate_cols(**{meta_name: meta.versions[version].ht()[mt.col_key]})
+    mt = mt.annotate_cols(**{meta_name: meta().versions[version].ht()[mt.col_key]})
 
     return mt
 
--- a/gnomad_qc/v4/resources/basics.py
+++ b/gnomad_qc/v4/resources/basics.py
@@ -385,17 +385,14 @@ def get_logging_path(name: str, version: str = CURRENT_VERSION) -> str:
     return f"{qc_temp_prefix(version)}{name}.log"
 
 
-def add_meta(
-    mt: hl.MatrixTable, version: str = CURRENT_VERSION, meta_name: str = "meta"
-) -> hl.MatrixTable:
+def add_meta(mt: hl.MatrixTable, meta_name: str = "meta") -> hl.MatrixTable:
     """
     Add metadata to MT in 'meta_name' column.
 
     :param mt: MatrixTable to which 'meta_name' annotation should be added
-    :param version: Version of metadata ht to use for annotations
     :return: MatrixTable with metadata added in a 'meta' column
     """
-    mt = mt.annotate_cols(**{meta_name: meta().versions[version].ht()[mt.col_key]})
+    mt = mt.annotate_cols(**{meta_name: meta().ht()[mt.col_key]})
 
     return mt
 
--- a/gnomad_qc/v4/resources/meta.py
+++ b/gnomad_qc/v4/resources/meta.py
@@ -61,8 +61,8 @@ def meta(
     )
 
 
-# This is used in `meta()` function anymore but is kept here to quickly get the current
-# existing versions of sample metadata.
+# This isn't used in `meta()` function anymore but is kept here to quickly get the
+# currently existing versions of sample metadata.
 _meta_versions = {
     "exomes": {
         "4.0": TableResource(
--- a/gnomad_qc/v4/resources/meta.py
+++ b/gnomad_qc/v4/resources/meta.py
@@ -1,7 +1,8 @@
 """Script containing metadata related resources."""
 from gnomad.resources.resource_utils import TableResource, VersionedTableResource
 
-from gnomad_qc.v4.resources.constants import CURRENT_VERSION, VERSIONS
+from gnomad_qc.resource_utils import check_resource_existence
+from gnomad_qc.v4.resources.constants import CURRENT_VERSION
 
 
 def _meta_root_path(version: str = CURRENT_VERSION, data_type: str = "exomes") -> str:
@@ -29,23 +30,31 @@ def meta_tsv_path(version: str = CURRENT_VERSION, data_type: str = "exomes") -> 
     )
 
 
-def meta(data_type: str = "exomes") -> VersionedTableResource:
+def meta(
+    version: str = CURRENT_VERSION, data_type: str = "exomes", overwrite: bool = False
+) -> TableResource:
     """
     Get the gnomAD v4 sample QC meta VersionedTableResource.
 
+    :param version: Version of meta that corresponds to its data type.
     :param data_type: Data type ("exomes" or "genomes"). Default is "exomes".
+    :param overwrite: Whether to overwrite existing resources.
     :return: gnomAD v4 sample meta VersionedTableResource.
     """
-    return VersionedTableResource(
-        CURRENT_VERSION,
-        {
-            version: TableResource(
-                path=(
-                    f"{_meta_root_path(version, data_type)}/gnomad.{data_type}.v{version}.sample_qc_metadata.ht"
-                )
-            )
-            for version in VERSIONS
+    # Check if the meta Table exists
+    check_resource_existence(
+        output_step_resources={
+            f"{data_type}_{version}": [
+                f"{_meta_root_path(version, data_type)}/gnomad.{data_type}.v{version}.sample_qc_metadata.ht"
+            ],
         },
+        overwrite=overwrite,
+    )
+
+    return TableResource(
+        path=(
+            f"{_meta_root_path(version, data_type)}/gnomad.{data_type}.v{version}.sample_qc_metadata.ht"
+        )
     )
 
 
--- a/gnomad_qc/v4/resources/meta.py
+++ b/gnomad_qc/v4/resources/meta.py
@@ -9,7 +9,7 @@ def _meta_root_path(version: str = CURRENT_VERSION, data_type: str = "exomes") -
     Retrieve the path to the root metadata directory.
 
     :param version: gnomAD version.
-    :param data_type: Data type (exomes or genomes), defaults to exomes.
+    :param data_type: Data type ("exomes" or "genomes"). Default is "exomes".
     :return: String representation of the path to the root metadata directory.
     """
     return f"gs://gnomad/v{version}/metadata/{data_type}"
@@ -20,8 +20,8 @@ def meta_tsv_path(version: str = CURRENT_VERSION, data_type: str = "exomes") -> 
     Get the path to the finalized sample metadata information after sample QC.
 
     :param version: gnomAD version.
-    :param data_type: Data type (exomes or genomes), defaults to exomes.
-    :return: String path to the finalized metadata.
+    :param data_type: Data type ("exomes" or "genomes"). Default is "exomes".
+    :return: String path to the finalized metadata TSV.
     """
     return (
         f"{_meta_root_path(version, data_type)}/gnomad.{data_type}."
@@ -33,9 +33,8 @@ def meta(data_type: str = "exomes") -> VersionedTableResource:
     """
     Get the gnomAD v4 sample QC meta VersionedTableResource.
 
-    :param data_type: Data type of annotation resource. e.g. "exomes" or "genomes".
-       Default is "exomes".
-    :return: gnomAD v4 meta VersionedTableResource.
+    :param data_type: Data type ("exomes" or "genomes"). Default is "exomes".
+    :return: gnomAD v4 sample meta VersionedTableResource.
     """
     return VersionedTableResource(
         CURRENT_VERSION,
@@ -50,6 +49,7 @@ def meta(data_type: str = "exomes") -> VersionedTableResource:
     )
 
 
+# The following variables only apply to "exomes" data.
 _project_meta_versions = {
     "4.0": TableResource(
         path="gs://gnomad/v4.0/metadata/exomes/gnomad.exomes.v4.0.project_meta.ht"
--- a/gnomad_qc/v4/resources/meta.py
+++ b/gnomad_qc/v4/resources/meta.py
@@ -23,7 +23,10 @@ def meta_tsv_path(version: str = CURRENT_VERSION, data_type: str = "exomes") -> 
     :param data_type: Data type (exomes or genomes), defaults to exomes.
     :return: String path to the finalized metadata.
     """
-    return f"{_meta_root_path(version)}/gnomad.{data_type}.v{version}.metadata.tsv.gz"
+    return (
+        f"{_meta_root_path(version, data_type)}/gnomad.{data_type}."
+        f"v{version}.metadata.tsv.gz"
+    )
 
 
 def meta(data_type: str = "exomes") -> VersionedTableResource:
--- a/gnomad_qc/v4/resources/meta.py
+++ b/gnomad_qc/v4/resources/meta.py
@@ -61,6 +61,21 @@ def meta(
     )
 
 
+# This is used in `meta()` function anymore but is kept here to quickly get the current
+# existing versions of sample metadata.
+_meta_versions = {
+    "exomes": {
+        "4.0": TableResource(
+            path="gs://gnomad/v4.0/metadata/exomes/gnomad.exomes.v4.0.sample_qc_metadata.ht"
+        ),
+    },
+    "genomes": {
+        "4.0": TableResource(
+            path="gs://gnomad/v4.0/metadata/genomes/gnomad.genomes.v4.0.sample_qc_metadata.ht"
+        ),
+    },
+}
+
 # The following variables only apply to "exomes" data.
 _project_meta_versions = {
     "4.0": TableResource(
--- a/gnomad_qc/v4/resources/meta.py
+++ b/gnomad_qc/v4/resources/meta.py
@@ -31,24 +31,27 @@ def meta_tsv_path(version: str = CURRENT_VERSION, data_type: str = "exomes") -> 
 
 
 def meta(
-    version: str = CURRENT_VERSION, data_type: str = "exomes", overwrite: bool = False
+    version: str = CURRENT_VERSION,
+    data_type: str = "exomes",
 ) -> TableResource:
     """
     Get the gnomAD v4 sample QC meta VersionedTableResource.
 
-    :param version: Version of meta that corresponds to its data type.
+    Function will check that metadata exists for the requested gnomAD version and data type
+    and will throw an error if it already exists.
+
+    :param version: gnomAD version.
     :param data_type: Data type ("exomes" or "genomes"). Default is "exomes".
-    :param overwrite: Whether to overwrite existing resources.
     :return: gnomAD v4 sample meta VersionedTableResource.
     """
-    # Check if the meta Table exists
+    # Check if the meta Table exists for specified version and data type combination
     check_resource_existence(
         output_step_resources={
             f"{data_type}_{version}": [
                 f"{_meta_root_path(version, data_type)}/gnomad.{data_type}.v{version}.sample_qc_metadata.ht"
             ],
         },
-        overwrite=overwrite,
+        overwrite=False,
     )
 
     return TableResource(
--- a/gnomad_qc/v4/resources/meta.py
+++ b/gnomad_qc/v4/resources/meta.py
@@ -38,7 +38,7 @@ def meta(
     Get the gnomAD v4 sample QC meta VersionedTableResource.
 
     Function will check that metadata exists for the requested gnomAD version and data type
-    and will throw an error if it already exists.
+    and will throw an error if it doesn't exist.
 
     :param version: gnomAD version.
     :param data_type: Data type ("exomes" or "genomes"). Default is "exomes".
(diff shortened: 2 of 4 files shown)


KoalaQin:
I put both of you as reviewer, but feel free to do it once you have time.

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
```diff
@@ -0,0 +1,129 @@
...
+    :param subset_ht: Updated meta HT of HGDP + TGP subset
+    :return: Updated v4 genomes meta HT
+    """
+    # Some HGDP/TGP samples in v3.1 meta HT have a prefix of "v3.1::" in the key 's',
+    # but the project_meta.sample_id field does not have this prefix. This is used as
+    # a temp key to join the subset meta HT.
+    release_v3 = ht.filter(ht.release).count()
+    ht1 = ht.filter(ht.subsets.hgdp | ht.subsets.tgp)
+    ht1 = ht1.naive_coalesce(5).checkpoint(new_temp_file("v4.0_genomes_meta", "ht"))
+    ht2 = ht.filter(~(ht.subsets.hgdp | ht.subsets.tgp))
```
ch-kr, line 49: you don't use the full HT (`ht`) again after this, right? to avoid cluttering the variable namespace, you could make this
```
ht = ht.filter(~(ht.subsets.hgdp | ht.subsets.tgp))
```

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
```diff
@@ -0,0 +1,129 @@
...
+    ht2 = ht2.checkpoint(new_temp_file("v4.0_genomes_meta", "ht"))
```
ch-kr, line 50: I haven't used Hail's `new_temp_file` before, so just double checking: is it OK that this has the same file name as the HT checkpointed two lines above?

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
```diff
@@ -0,0 +1,129 @@
...
+    """
+    Import updated annotations from subset meta HT to full meta HT.
+
+    :param ht: v3.1 genomes meta HT of genomes
+    :param subset_ht: Updated meta HT of HGDP + TGP subset
+    :return: Updated v4 genomes meta HT
+    """
+    # Some HGDP/TGP samples in v3.1 meta HT have a prefix of "v3.1::" in the key 's',
+    # but the project_meta.sample_id field does not have this prefix. This is used as
+    # a temp key to join the subset meta HT.
```
ch-kr, line 45: and to confirm, the subset meta HT also doesn't have the  "v3.1::" prefix?

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
```diff
@@ -0,0 +1,129 @@
...
+    release_v3 = ht.filter(ht.release).count()
+    ht1 = ht.filter(ht.subsets.hgdp | ht.subsets.tgp)
+    ht1 = ht1.naive_coalesce(5).checkpoint(new_temp_file("v4.0_genomes_meta", "ht"))
+    ht2 = ht.filter(~(ht.subsets.hgdp | ht.subsets.tgp))
+    ht2 = ht2.checkpoint(new_temp_file("v4.0_genomes_meta", "ht"))
+
+    ht1 = ht1.key_by(ht1.project_meta.sample_id)
+    annot = subset_ht[ht1.key]
+    subpop = annot.hgdp_tgp_meta.population
+    freemix = annot.bam_metrics.freemix
```
ch-kr, line 55: did this field get updated? I thought some samples were newly removed due to contamination estimates from CHARR, not freemix

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
```diff
@@ -0,0 +1,129 @@
...
+        "Number of genome release samples will change from %s in v3.1 to %s in v4.0.",
+        release_v3,
+        release_v4,
+    )
+
+    return ht
+
+
+def main(args):
+    """Create v4.0 genomes sample QC metadata HT."""
```
ch-kr, line 102:
```suggestion
    """Create updated v4 genomes sample QC metadata HT."""
```

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
```diff
@@ -0,0 +1,129 @@
...
+    hl.init(
+        log="/create_v4.0_genomes_meta.log",
```
ch-kr, line 104:
```suggestion
        log="/create_v4_genomes_meta.log",
```

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
```diff
@@ -0,0 +1,129 @@
+"""
+Script to create sample QC metadata HT for genomes.
+
+This script is used to create the sample QC metadata HT for genomes. The main updates
+from v3.1 to v4.0 are:
+Sample QC annotations for HGDP + TGP subset, this script directly merges the updated
+annotations from the subset meta HT to the full meta HT.
```
ch-kr, line 7:
```suggestion
The main updates from v3.1 to v4.0 are are due to the following updates in the HGDP + TGP subset meta:
- 169 samples added
- 110 samples removed

The 169 samples added were diverse HGDP/TGP samples in v3.1 that were unintentionally removed during sample outlier detection.

The 110 samples removed were removed due to the following reasons:
- Filtered due to contamination flagged by CHARR
- Removed due to updated relatedness results

All of the samples impacted (169 added, 110 released) are releasable; the updates we made here were to their sample QC status (high quality vs filtered).
```

it might also be helpful to add this code snippet you sent me:
```
import hail as hl

from gnomad_qc.v4.resources.sample_qc import hgdp_tgp_meta_updated

from gnomad_qc.v4.annotations.generate_freq_genomes import get_updated_release_samples

pop_diff, added, subtracted = get_updated_release_samples(hgdp_tgp_meta_updated.ht())
```

also, please update the text if I have gotten any of the details wrong

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
```diff
@@ -0,0 +1,129 @@
...
+            pop=hl.if_else(
+                ht.population_inference.pop == "oth",
+                "remaining",
+                ht.population_inference.pop,
+            )
+        )
+    )
+    release_v4 = ht.filter(ht.release).count()
+    logger.info(
+        "Number of genome release samples will change from %s in v3.1 to %s in v4.0.",
```
ch-kr, line 93:
```suggestion
        "Number of genome release samples will change from %s in v3.1 to %s in v4",
```

ch-kr: it might also be helpful to note as a comment what you are expecting the new number to be. could you also add a check here to make sure that `release_v4` is greater than `release_v3` and throw an error if not?

ch-kr (changes requested):
some questions and suggestions (many are rewording suggestions)

gnomad_qc/v4/resources/meta.py
```diff
@@ -1,34 +1,54 @@
...
+        f"v{version}.metadata.tsv.gz"
+    )
 
 
-_meta_versions = {
-    "4.0": TableResource(
-        path="gs://gnomad/v4.0/metadata/exomes/gnomad.exomes.v4.0.sample_qc_metadata.ht"
-    ),
-}
+def meta(data_type: str = "exomes") -> VersionedTableResource:
```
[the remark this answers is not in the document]
KoalaQin: Good idea! So we don't have to write everytime.

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
```diff
@@ -0,0 +1,129 @@
...
+        tmp_dir="gs://gnomad-tmp-4day",
+    )
+
+    logger.info("Importing v3.1 genomes meta HT...")
+    ht = v3_meta.ht()
+    logger.info("Importing updated HGDP + TGP meta HT...")
+    subset_ht = hgdp_tgp_meta_updated.ht()
+    logger.info("Updating annotations...")
+    ht = import_updated_annotations(ht, subset_ht)
+    ht = ht.naive_coalesce(160)
```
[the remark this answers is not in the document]
KoalaQin: The v3 meta HT has 159 partitions, if I don't put this, when I use `union`, it became 313 partitions for v4 meta HT, but when I use 160, the final partitions are 83, I don't know if it matters, I can also remove it.

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
```diff
@@ -0,0 +1,129 @@
...
+    :param ht: v3.1 genomes meta HT of genomes
+    :param subset_ht: Updated meta HT of HGDP + TGP subset
+    :return: Updated v4 genomes meta HT
+    """
+    # Some HGDP/TGP samples in v3.1 meta HT have a prefix of "v3.1::" in the key 's',
+    # but the project_meta.sample_id field does not have this prefix. This is used as
+    # a temp key to join the subset meta HT.
+    release_v3 = ht.filter(ht.release).count()
+    ht1 = ht.filter(ht.subsets.hgdp | ht.subsets.tgp)
+    ht1 = ht1.naive_coalesce(5).checkpoint(new_temp_file("v4.0_genomes_meta", "ht"))
```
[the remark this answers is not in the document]
KoalaQin: I was trying to speed up the run, but it took maximum 3 minutes for the whole thing, I will just remove all the `naive_coalesce()`

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
KoalaQin: Yes, that's just's the prefix, it will get some random text in the name.

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
KoalaQin: Yes, I confirmed with anti_join, if we use `sample_id` in full meta and `s` in the subset meta, it's the same.

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
KoalaQin: It is CHARR, but we converted CHARR value and stored in freemix.

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
KoalaQin: Why do you change all my v4.0 to v4? I had this exchange with Julia that we made everything v4.0 in the freq_ht code.

KoalaQin:
@ch-kr Test run is here: [0677fc8311b04852b9661913b6278822](https://console.cloud.google.com/dataproc/jobs/0677fc8311b04852b9661913b6278822?region=us-central1&project=broad-mpg-gnomad) if you want to see the naming of new_temp_file. Runtime is fine without repartitioning for this one.

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
ch-kr: my personal preference is to not hard code v4.0, since that will just mean it is one more thing to update if this code ever needs to be rerun.

that said, Julia has been much closer to the code and what goes into it than I have, so I would defer to her preference here

gnomad_qc/v4/resources/meta.py
```diff
@@ -1,35 +1,55 @@
...
+    """
+    Get the gnomAD v4 sample QC meta VersionedTableResource.
+
+    :param data_type: Data type ("exomes" or "genomes"). Default is "exomes".
+    :return: gnomAD v4 sample meta VersionedTableResource.
+    """
+    return VersionedTableResource(
+        CURRENT_VERSION,
```
ch-kr, line 40: I just realized that this function assumes we will have a sample meta for both exomes and genomes for each version in `VERSIONS`, but what if we update the exome meta in a minor release and don't update the genomes?

I realize this makes the function longer, but it's probably a good idea to have this function take in version as an argument (with the default set to `CURRENT_VERSION`), re-add a meta_versions dictionary that tracks the available sample meta versions for both data types, and does a check at the top of the function verifying that sample meta exists for the specified combination of version and data_type

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
```diff
@@ -0,0 +1,144 @@
...
+Script to create sample QC metadata HT for genomes.
+
+The main updates from v3.1 to v4.0 are due to the following updates in the HGDP + TGP subset meta:
+- 169 samples added
+- 110 samples removed
+
+The 169 samples added were diverse HGDP/TGP samples in v3.1 that were unintentionally removed during sample outlier detection.
+
+The 110 samples were removed due to the following reasons:
+- Contamination flagged by CHARR.
```
ch-kr, line 11:
```suggestion
- Contamination flagged by CHARR (CHARR score stored in `freemix` column).
```

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
```diff
@@ -0,0 +1,144 @@
...
+            )
+        )
+    )
+    release_v4 = ht.filter(ht.release).count()
+    logger.info(
+        "Number of genome release samples will change from %s in v3.1 to %s in v4.0",
+        release_v3,
+        release_v4,
+    )
+    if release_v4 <= release_v3 or release_v4 != release_v3 + 169 - 110:
```
ch-kr, line 102: just a note that this second part (`release_V3 + 169 - 110`) would need to get updated if we ever had to run this again. what about making this a constant at the top of the script? something like:
```
N_DIFF_SAMPLES = 169 - 110
"""
The number of samples that have been updated between v3.1 and v4.0.
"""
```
and then
```
if release_v4 <= release_v3 or release_v4 != release_v3 + N_DIFF_SAMPLES:
```

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
```diff
@@ -1,10 +1,20 @@
...
+- 110 samples removed
+
+The 169 samples added were diverse HGDP/TGP samples in v3.1 that were unintentionally removed during sample outlier detection.
+
+The 110 samples were removed due to the following reasons:
+- Contamination flagged by CHARR.
+- Subcontinental PCA outliers within the subset.
+- Updated relatedness results within the subset, re-evaluation of relatedness
+  inference of the subset to the rest of gnomAD release samples, and removal of any
+  that are duplicates of v4.0 exomes release samples.
```
ch-kr, line 15:
```suggestion
  samples that are duplicates of v4.0 exomes release samples.
```
also basic question: there were duplicates between the HGDP/1KG samples and v4 samples?

ch-kr (commented):
one thought about `meta()` and suggestion to add constant to `create_sample_qc_metadata_ht_genomes.py`

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
KoalaQin: Yes, there are a few: https://github.com/broadinstitute/gnomad_qc/blob/8e55307cfef330371174f03b40089a3975df7e00/gnomad_qc/v4/annotations/generate_freq_genomes.py#L190

ch-kr: follow-up question: why weren't these caught by cuKING? also, this means that our documentation
```
After running cuKING, we used the [maximal_independent_set method in Hail](https://hail.is/docs/0.2/methods/misc.html#hail.methods.maximal_independent_set) to get the largest set of samples with no pair of samples related at the 2nd degree or closer. When we had to select a sample amongst multiple possibilities, we used the same scheme to favor a sample over another:
Genomes had priority over exomes in gnomAD v4.0
Within genomes
- PCR free had priority over PCR+
- In a parent/child pair, the parent was kept
- Ties broken by the sample with highest mean coverage
Within exomes
- In a parent/child pair, the parent was kept
- Ties broken by the sample with the higher mean coverage on chromosome 20
```
is no longer fully correct, right?

KoalaQin: It's caught in cuKing, that's the results we're using to get them. Oh, I think here we put 'exomes' as priority.. @jkgoodrich could you confirm this?

KoalaQin: I meant we didn't want to change the number of release samples in exomes, this one came after, almost the last step when I realized we should check the duplicates.

ch-kr: follow-up question: if these samples were flagged by cuKING as being duplicates, wouldn't the v4 exomes results have been filtered such that this wouldn't be an issue? as in, shouldn't the HGDP/1KG samples been prioritized in the pruning given the logic here: https://github.com/broadinstitute/gnomad_qc/blob/main/gnomad_qc/v4/sample_qc/relatedness.py#L275? or were these samples not releasable in v3 to begin with?

gnomad_qc/v4/resources/basics.py
```diff
@@ -385,17 +385,14 @@ def get_logging_path(name: str, version: str = CURRENT_VERSION) -> str:
     return f"{qc_temp_prefix(version)}{name}.log"
 
 
-def add_meta(
-    mt: hl.MatrixTable, version: str = CURRENT_VERSION, meta_name: str = "meta"
-) -> hl.MatrixTable:
+def add_meta(mt: hl.MatrixTable, meta_name: str = "meta") -> hl.MatrixTable:
```
ch-kr, line 388: it might still be helpful to be able to specify which meta version requested here, right?

gnomad_qc/v4/resources/meta.py
```diff
@@ -29,23 +30,31 @@ def meta_tsv_path(version: str = CURRENT_VERSION, data_type: str = "exomes") ->
     )
 
 
-def meta(data_type: str = "exomes") -> VersionedTableResource:
+def meta(
+    version: str = CURRENT_VERSION, data_type: str = "exomes", overwrite: bool = False
```
ch-kr, line 34: this is contradictory to my other comments, but I don't think this overwrite needs to be an argument -- you wouldn't ever want to set this to True, since this is only used to check if the meta already exists. it also cleans up calls to this function; e.g.
```
    ht.write(
        v4_meta(data_type="genomes", overwrite=args.overwrite).path,
        overwrite=args.overwrite,
    )
```
could be reverted to
```
    ht.write(v4_meta(data_type="genomes").path, overwrite=args.overwrite)
```

gnomad_qc/v4/resources/meta.py
```diff
@@ -29,23 +30,31 @@ def meta_tsv_path(version: str = CURRENT_VERSION, data_type: str = "exomes") ->
...
+) -> TableResource:
     """
     Get the gnomAD v4 sample QC meta VersionedTableResource.
 
+    :param version: Version of meta that corresponds to its data type.
```
ch-kr, line 39:
```suggestion
    :param version: gnomAD version.
```

gnomad_qc/v4/resources/meta.py
```diff
@@ -29,23 +30,31 @@ def meta_tsv_path(version: str = CURRENT_VERSION, data_type: str = "exomes") ->
     )
 
 
-def meta(data_type: str = "exomes") -> VersionedTableResource:
+def meta(
+    version: str = CURRENT_VERSION, data_type: str = "exomes", overwrite: bool = False
+) -> TableResource:
     """
     Get the gnomAD v4 sample QC meta VersionedTableResource.
```
ch-kr, line 37:
```suggestion
    Get the gnomAD v4 sample QC meta VersionedTableResource.
    
    Function will check that metadata exists for the requested gnomAD version and data type
    and will throw an error if not.
```

gnomad_qc/v4/resources/meta.py
```diff
@@ -29,23 +30,31 @@ def meta_tsv_path(version: str = CURRENT_VERSION, data_type: str = "exomes") ->
...
-    return VersionedTableResource(
-        CURRENT_VERSION,
-        {
-            version: TableResource(
-                path=(
-                    f"{_meta_root_path(version, data_type)}/gnomad.{data_type}.v{version}.sample_qc_metadata.ht"
-                )
-            )
-            for version in VERSIONS
+    # Check if the meta Table exists
```
ch-kr, line 44:
```suggestion
    # Check if the meta Table exists for specified version and data type combination
```

gnomad_qc/v4/resources/meta.py
```diff
@@ -29,23 +30,31 @@ def meta_tsv_path(version: str = CURRENT_VERSION, data_type: str = "exomes") ->
...
+    check_resource_existence(
+        output_step_resources={
+            f"{data_type}_{version}": [
+                f"{_meta_root_path(version, data_type)}/gnomad.{data_type}.v{version}.sample_qc_metadata.ht"
+            ],
         },
+        overwrite=overwrite,
```
ch-kr, line 51:
```suggestion
        overwrite=False,
```

ch-kr (changes requested):
two minor suggestions for `add_meta` and `meta`

gnomad_qc/v4/resources/meta.py
KoalaQin: I think because I need to test the script, `check_resource_existence` here will throw an error when it exists. I will revert it.

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
KoalaQin: Yes, because these samples weren't in v3 release, so they weren't considered before. Now I used them as in v3 (not in v3 release) to get the duplicates.

gnomad_qc/v4/resources/meta.py
```diff
@@ -31,24 +31,27 @@ def meta_tsv_path(version: str = CURRENT_VERSION, data_type: str = "exomes") ->
...
+
+    :param version: gnomAD version.
     :param data_type: Data type ("exomes" or "genomes"). Default is "exomes".
-    :param overwrite: Whether to overwrite existing resources.
     :return: gnomAD v4 sample meta VersionedTableResource.
     """
-    # Check if the meta Table exists
+    # Check if the meta Table exists for specified version and data type combination
     check_resource_existence(
         output_step_resources={
```
ch-kr, line 49: shouldn't the behavior of this error out if the HT doesn't exist? as in, shouldn't this be `input_step_resources` rather than `output_step_resources`? it should be going through this logic in this function https://github.com/broadinstitute/gnomad_qc/blob/main/gnomad_qc/resource_utils.py#L36

ch-kr (commented):
question about `check_resource_existence`

gnomad_qc/v4/resources/meta.py
KoalaQin: Oops, sorry, I misread the usage. I changed it.

gnomad_qc/v4/sample_qc/create_sample_qc_metadata_ht_genomes.py
ch-kr: hmmm. OK. I think this is fine then, and consistent with our documentation

gnomad_qc/v4/resources/meta.py
```diff
@@ -1,35 +1,67 @@
...
+    :return: String path to the finalized metadata TSV.
     """
-    return f"{_meta_root_path(version)}/gnomad.exomes.v{version}.metadata.tsv.gz"
+    return (
+        f"{_meta_root_path(version, data_type)}/gnomad.{data_type}."
+        f"v{version}.metadata.tsv.gz"
+    )
 
 
-_meta_versions = {
```
ch-kr, line 27: should we revive this dictionary so that it's easy to tell which metadata versions exist from a quick scan through this code? something like
```
_meta_versions = {
    "exomes": {
        "4.0": TableResource(
            path="gs://gnomad/v4.0/metadata/exomes/gnomad.exomes.v4.0.sample_qc_metadata.ht"
        ),
    },
    "genomes": {
        "4.0": TableResource(
            path="gs://gnomad/v4.0/metadata/genomes/gnomad.genomes.v4.0.sample_qc_metadata.ht"
        ),
    }
}
```
though this wouldn't get used based on the current structure of `meta()`. what do you think?

ch-kr (approved):
one question but LGTM!

PR #501, opened 2023-10-23
Add summary stats script

jkgoodrich:
I have not yet run or tested `--generate-gene-lof-matrix` or `--summarize-gene-lof-matrix` since it requires a densify


gnomad_qc/v4/assessment/summary_stats.py
```diff
@@ -0,0 +1,230 @@
...
+                filter_name = filter_name + ".100k_downsampling"
+
+            logger.info("Getting summary counts per variant category...")
+            ht = get_summary_counts(
+                ht.select_globals(),
+                canonical_only=False,
+                mane_select_only=True,
+                index=freq_index,
+            )
+            meta_ht = meta.ht()
```
matren395, line 96:
```suggestion
            meta_ht = meta(data_type=data_type).ht()
```

gnomad_qc/v4/assessment/summary_stats.py
```diff
@@ -0,0 +1,230 @@
...
+            meta_ht = meta_ht.filter(meta_ht.release)
+
+            logger.info(f"Number of release samples: {meta_ht.count()}")
+            ht = ht.annotate_globals(num_release_samples=meta_ht.count())
+            ht.write(
+                release_summary_stats(
+                    test=test, data_type="exomes", filter_name=filter_name
```
matren395, line 103:
```suggestion
                    test=test, data_type=data_type, filter_name=filter_name
```

gnomad_qc/v4/assessment/summary_stats.py
```diff
@@ -0,0 +1,230 @@
...
+                mt=mt,
+                tx_ht=None,
+                additional_grps=[
+                    "broad_ukb_union_intervals",
+                    "broad_ukb_intersection_intervals",
+                    "fail_interval_qc",
+                ],
+            )
+            mt.write(
+                release_lof(test=test, data_type="exomes", mt=True).path,
```
matren395, line 146:
```suggestion
                release_lof(test=test, data_type=data_type, mt=True).path,
```

gnomad_qc/v4/assessment/summary_stats.py
```diff
@@ -0,0 +1,230 @@
...
+                overwrite,
+            )
+
+        if args.summarize_gene_lof_matrix:
+            mt = release_lof(test=test, data_type="exomes", mt=True).mt()
```
matren395, line 151:
```suggestion
            mt = release_lof(test=test, data_type=data_type, mt=True).mt()
```

gnomad_qc/v4/assessment/summary_stats.py
```diff
@@ -0,0 +1,230 @@
...
+            mt = mt.annotate_cols(
+                meta=mt.meta.annotate(pop=mt.meta.population_inference.pop)
+            )
+            ht = default_generate_gene_lof_summary(mt)
+            ht.write(
+                release_lof(test=test, data_type="exomes").path,
```
matren395, line 157:
```suggestion
                release_lof(test=test, data_type=data_type).path,
```

gnomad_qc/v4/assessment/summary_stats.py
```diff
@@ -0,0 +1,230 @@
...
+    parser.add_argument(
+        "--generate-gene-lof-matrix",
+        help="Generate gene LoF matrix.",
+        action="store_true",
+    )
+    parser.add_argument(
+        "--summarize-gene-lof-matrix",
+        help="Creates gene LoF matrix summary Table.",
+        action="store_true",
+    )
```
matren395, line 223: parser.add_argument(
        "--data-type",
        default="exomes",
        choices=["exomes","genomes"],
        help="Data type (exomes or genomes) to produce summary stats for."
    )

matren395 (commented):
There have been a few requests for summary stats for genomes - possible in a notebook but could be good to productionize/add an argument for it in this script ?

gnomad_qc/v4/assessment/summary_stats.py
matren395: wait it doesn't work like this in this branch - did meta change from main?

KoalaQin:
Should we get this PR in too? I can help review if needed. Want to wrap up all the stats tickets.

matren395:
Hmm so this code was instead ran in a notebook and essentially this runs [get_summary_stats()](https://github.com/broadinstitute/gnomad_methods/blob/c09762282a71bc0d341c224288fa4c865f10c2e2/gnomad/assessment/summary_stats.py#L166) to get per-callset stats (not per sample) of just about exactly what we're doing. 

Following this goal, I think we should maybe calculate the TOTALS of everything we're calculating per sample as sums instead from the intermediate file ? I can add the code for it - if we want per-callset instead of per-sample - stats

matren395:
> Should we get this PR in too? I can help review if needed. Want to wrap up all the stats tickets.

so I took a run at this in #611 - just getting those stats from the existing aggregated table, while they're already there. LMK your thoughts.

matren395:
After another look, this actually looks like it's doing something a bit more complex and productionized and better versionized than the simple code I put into #611 . Running the existing summary stats method =/= calculating things from the intermediate per-sample table , though we can return 99% of the same results from it. @jkgoodrich is this worth pursuing, or is calculating plenty of per-callset stats via #611 good enough ?

matren395 (approved):
okay looks good to me !

PR #503, opened 2023-10-23
Runtime combined FAF changes

jkgoodrich:
- Fix to getting the correct index for the grpmax.faf95
- Fix to prevent mem errors when trying to convert to a pandas dataframe from Hail
- Add options for what stats to include on the release combined FAF Table

### DIFF
--- a/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
+++ b/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
@@ -315,6 +315,9 @@ def perform_cmh_test(
     df = _ht.to_spark().write.option("compression", "zstd")
     df.mode("overwrite").parquet(tmp_path)
     df = pd.read_parquet(tmp_path)
+
+    # Perform CMH test on the list of 2x2 matrices (list of lists of 2 lists with 2
+    # elements each).
     df["cmh"] = df.apply(
         lambda x: StratifiedTable(
             [list([list(a) for a in p]) for p in x.an]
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -303,30 +306,52 @@ def _get_freq_by_pop(
...
+    # Export one compressed Parquet file per partition.
+    df = _ht.to_spark().write.option("compression", "zstd")
+    df.mode("overwrite").parquet(tmp_path)
+    df = pd.read_parquet(tmp_path)
+    df["cmh"] = df.apply(
+        lambda x: StratifiedTable(
+            [list([list(a) for a in p]) for p in x.an]
+        ).test_null_odds(),
+        axis=1,
+    )
```
mike-w-wilson, line 323: Maybe just a note what the `[list([list...` is doing, from conversations before this and docs I think its just building that 2x2 matrix but just want to confirm and note for future reference.

gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
```diff
@@ -494,7 +537,9 @@ def main(args):
                 ),
             )
             ht.describe()
-            ht.write(res.final_combined_faf_ht.path, overwrite=overwrite)
+            ht.naive_coalesce(10000).write(
```
mike-w-wilson, line 540: parameterize?

mike-w-wilson (approved):
Two minor things but then looks good to me!

PR #504, opened 2023-10-23
set set_negatives_to_zero=True for the final v4 genomes freq run

KoalaQin:
The final run was only successful when we changed this to True.


gnomad_qc/v4/annotations/generate_freq_genomes.py
```diff
@@ -955,7 +955,7 @@ def generate_v4_genomes_callstats(ht: hl.Table, an_ht: hl.Table) -> hl.Table:
         [ht.freq, ht.ann_array[3].freq],
         [freq_meta, fmeta[4]],
         operation="diff",
-        set_negatives_to_zero=False,
+        set_negatives_to_zero=True,
```
jkgoodrich, line 958: Can you add a note for why we are adding this, and that we advise keeping it as False unless it's understood why negative values are happening

KoalaQin: Is it clear like this?

jkgoodrich (approved):
LGTM!

PR #508, opened 2023-10-24
Flatten the gnomad/non_ukb structs in fafmax and grpmax

mike-w-wilson:
These werent being flattened properly nor were we grabbing all the data for the VCF. They arent the prettiest implementation but grpmax has two renames within it and in order to keep the subset specific annotations together I added the for loop for the subsets.


jkgoodrich (approved):
lgtm

PR #511, opened 2023-10-24
Combined freq set Y to None earlier before faf is made

jkgoodrich:
Expects a freq struct not faf, so move up to apply to freq before faf is made


mike-w-wilson (approved):
Makes sense

PR #514, opened 2023-10-24
suggestions to use DOWNSAMPLINGS in the v3 freq_meta

KoalaQin:
I copied the changes you made for the merge_arrays and update how we use get `downsamplings`, test worked.

### DIFF
--- a/gnomad_qc/v4/annotations/generate_freq_genomes.py
+++ b/gnomad_qc/v4/annotations/generate_freq_genomes.py
@@ -93,39 +93,6 @@ JOIN_FREQS = ["release", "pop_diff", "added", "subtracted"]
 FREQ_GLOBALS = ("freq_meta", "freq_meta_sample_count")
 """Global annotations on the frequency Table."""
 
-DOWNSAMPLINGS = [
-    10,
-    20,
-    50,
-    100,
-    158,
-    200,
-    456,
-    500,
-    1000,
-    1047,
-    1736,
-    2000,
-    2419,
-    2604,
-    5000,
-    5316,
-    7647,
-    10000,
-    15000,
-    20000,
-    20744,
-    25000,
-    30000,
-    34029,
-    40000,
-    50000,
-    60000,
-    70000,
-    75000,
-]
-"""Downsamplings used in v3.1 release"""
-
 
 def filter_to_test(
     t: Union[hl.Table, hl.MatrixTable, hl.vds.VariantDataset],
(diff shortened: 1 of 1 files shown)


jkgoodrich:
Why though? Are the ones one the release HT wrong?

jkgoodrich:
If they are not wrong, I would much prefer take them from there, just like we do the hists

KoalaQin:
They are not present in v3 release HT.

KoalaQin:
I saw the error in your test run.

jkgoodrich:
So odd, I'm not sure why they are not there, thanks for putting in

gnomad_qc/v4/annotations/generate_freq_genomes.py
```diff
@@ -97,6 +93,39 @@
 FREQ_GLOBALS = ("freq_meta", "freq_meta_sample_count")
 """Global annotations on the frequency Table."""
 
+DOWNSAMPLINGS = [
```
jkgoodrich, line 96: Just so I know, where did this list come from? Was it from the freq_meta?

KoalaQin: Yes, I got the from get_freq() for genomes and then freq_meta.collect(), Mike also pasted in our Slack thread.

gnomad_qc/v4/annotations/generate_freq_genomes.py
```diff
@@ -1485,7 +1516,7 @@ def main(args):
         )
         ht = get_histograms(ht, v3_sites_ht)
         ht = ht.annotate_globals(
```
jkgoodrich, line 1518: What about something like this?
```suggestion
        # NOTE: The v3.1 release HT doesn't have a downsamplings global. Since
        #  non-standard downsampling values are created in the frequency script
        #  corresponding to population totals, so this needs to be determined from the
        #  freq_meta.
        downsamplings = list(
            {
                str(x["downsampling"]) 
                for x in hl.eval(ht.freq_meta) if "downsampling" in x
            }
        )
        downsamplings.sort()
        
        ht = ht.annotate_globals(
            downsamplings=downsamplings,
```

gnomad_qc/v4/annotations/generate_freq_genomes.py
```diff
@@ -1485,7 +1516,7 @@ def main(args):
...
-            downsamplings=v3_sites_ht.downsamplings,
+            downsamplings=hl.literal(DOWNSAMPLINGS),
```
jkgoodrich, line 1519: goes with suggestion above
```suggestion
```

KoalaQin:
Yes, it looks better, I only changed it to int() instead of str().

jkgoodrich:
Ah yes, good point

jkgoodrich (approved):
lgtm!

PR #518, opened 2023-10-25
Get rid of duplicates in revel

mike-w-wilson:
The current implementation leads to duplicate rows because the annotations and the scores are identical in mane and canonical tables. This gets around that issue by naming the annotation by table so as to not duplicate rows and then selects mane over canonical for revel_max


jkgoodrich (approved):
Oh, I get it now, thanks for figuring it out and changing!

KoalaQin (approved):
LGTM!

KoalaQin:
Could you rerun the code and record it here? https://app.zenhub.com/workspaces/gnomad-5f4d127ea61afc001d6be50b/issues/gh/broadinstitute/gnomad_production/1170

PR #519, opened 2023-10-25
Move duplication check in release so it is only performed once

jkgoodrich:


KoalaQin (commented):
I'm running the test, I found this makes the run a bit slow at the beginning but I think we only need it to run once for exomes once for genomes, it should be fine.

KoalaQin (approved):

PR #520, opened 2023-10-25
using high_quality instead of gnomad_high_quality from hgdp_tgp meta

KoalaQin:
In v3.1, `high_quality` means means NOT `hard_filtered` and NOT `gnomad_sample_filters.qc_metrics_filters`, and it was kept in hgdp_tgp_meta as `gnomad_high_quality`, in v4.0, `high_quality` means NOT `hard_filtered` and NOT `subcontinental_pca.outlier` from the subset. So, `gnomad_high_quality` in the subset meta shouldn't be copied to the full meta. Here, I corrected this.


KoalaQin:
@ch-kr, I have the small PR to correct the `high_quality`, so we will get correct count of high_quality samples.

ch-kr (approved):
thanks for adding! sorry didn't realize this was a single line change -- LGTM!

PR #523, opened 2023-10-27
Make HGDP + 1KG use the exact same methods as those used for the v3.1.2 release HT frequency

jkgoodrich:
We need to rerun the hotfix because the current v3_1_GT was already adjusted for adj

### DIFF
--- a/gnomad_qc/v4/annotations/generate_freq_genomes.py
+++ b/gnomad_qc/v4/annotations/generate_freq_genomes.py
@@ -1276,6 +1276,7 @@ def get_v4_genomes_release_resources(
             update_annotations,
             join_callstats_for_update,
             compute_an_for_new_variants,
+            compute_an_for_pop_diff,
         ],
         add_input_resources={
             "v3.1 metadata": {"v3_meta_ht": v3_meta},
--- a/gnomad_qc/v4/annotations/generate_freq_genomes.py
+++ b/gnomad_qc/v4/annotations/generate_freq_genomes.py
@@ -1370,7 +1370,7 @@ def main(args):
     )
     v3_hgdp_tgp_dense_mt = v3_hgdp_tgp_dense_mt.annotate_entries(
         GT=v3_hgdp_tgp_dense_mt.hom_alt_fix_investigation.sex_ploidy_adjusted_GT,
-        adj=v3_hgdp_tgp_dense_mt.unadjusted_adj,
+        adj=v3_hgdp_tgp_dense_mt.hom_alt_fix_investigation.unadjusted_adj,
     )
 
     # Temporary hotfix for depletion of homozygous alternate genotypes
--- a/gnomad_qc/v4/annotations/generate_freq_genomes.py
+++ b/gnomad_qc/v4/annotations/generate_freq_genomes.py
@@ -1365,8 +1365,8 @@ def main(args):
     #  v3.1.2 release sites HT.
 
     logger.info(
-        "Annotating GT with sex ploidy adjusted genotypes and adj calculated on genotypes 
-        without sex ploidy adjustment..."
+        "Annotating GT with sex ploidy adjusted genotypes and adj calculated on "
+        "genotypes without sex ploidy adjustment..."
     )
     v3_hgdp_tgp_dense_mt = v3_hgdp_tgp_dense_mt.annotate_entries(
         GT=v3_hgdp_tgp_dense_mt.hom_alt_fix_investigation.sex_ploidy_adjusted_GT,
--- a/gnomad_qc/v4/annotations/generate_freq_genomes.py
+++ b/gnomad_qc/v4/annotations/generate_freq_genomes.py
@@ -1280,7 +1280,6 @@ def patch_v4_genomes_callstats(freq_ht: hl.Table) -> hl.Table:
     freq_groups1 = hl.set(
         [
             {"group": "adj", "subset": "hgdp", "gen_anc": "han"},
-            {"group": "adj", "subset": "hgdp", "sex": "XX"},
             {"group": "adj", "subset": "hgdp", "gen_anc": "han", "sex": "XX"},
         ]
     )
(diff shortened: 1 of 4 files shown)


ch-kr (approved):
two documentation suggestions but LGTM!

KoalaQin:
The last 2 commits LGTM!

gnomad_qc/v4/annotations/generate_freq_genomes.py
```diff
@@ -1168,16 +1203,211 @@ def get_age_distribution(
...
+
+    pop_diff_patch_ht = compute_an_by_group_membership(
+        v3_vds,
+        pop_diff_group_membership_ht,
+        pop_diff_patch_ht,
+        include_full_call_stats=True,
+    ).checkpoint(new_temp_file("pop_diff_89834177", "ht"))
+
+    # Frequency groups that need a patch for variant chr9-93596925-ATTTTTTTTTTTTTT-A.
+    # In the original v3.1 a single Han sample had a missing GT at this variant. In the
```
KoalaQin, line 1432:
```suggestion
    # In the original v3.1, a single Han sample had a missing GT at this variant. In the
```

gnomad_qc/v4/annotations/generate_freq_genomes.py
```diff
@@ -1168,16 +1203,211 @@ def get_age_distribution(
...
+    # computation for v4.0, the _het_non_ref of True caused it to be given a 0/1 (the
+    # hl.if_else statement gets a False, so it defaults to the input GT. The original v3
+    # code didn't include the _het_non_ref check, and therefore it would have evaluated
+    # to missing.
```
KoalaQin: Could you add more details here? It's not very clear. 
What do you mean by the original v3 code, v3.1?

gnomad_qc/v4/annotations/generate_freq_genomes.py
```diff
@@ -1168,16 +1203,211 @@ def get_age_distribution(
...
+            ),
+            pop_diff_patch_ht.freq,
+            pop_diff_patch_ht.freq_meta,
+        )
+    )
+
+    # Frequency groups that need a patch for variant chr13-20656122-T-TAAAAAAAAGAA.
+    # This variant has an AC raw of 1 and AC adj of 2 because one of the TGP samples
+    # being removed has a raw GT 0/1, but adj GT of None because it fails adj. In v3.1
+    # it was None for both and therefore shouldn't be subtracted from raw either.
```
KoalaQin, line 1469:
```suggestion
    # being removed has a raw of GT 0/1 but adj GT of None because it fails adj. In v3.1 it was None for both and therefore shouldn't be subtracted from raw either.
```

KoalaQin (changes requested):

gnomad_qc/v4/annotations/generate_freq_genomes.py
```diff
@@ -1168,16 +1218,310 @@ def get_age_distribution(
...
+    :param freq_ht: Table with the call stats for the v4.0 genomes release.
+    :return: Table with the patched call stats for the v4.0 genomes release.
+    """
+    # For most of the variants, there is a single Han sample that was 0/1 in v3.1, but
+    # from the dense MT in v4.0 frequencies it was given a missing genotype. These
+    # variants are not in v3, which is why they get a missing genotype in the v4.0
+    # freq. We don’t understand why they were 0/1 in v3.1, we would expect them to
+    # have been missing. It should be 0/1 if missing v3 frequency had been handled
+    # correctly in the fix for high AB hets.
+    freq_groups1 = hl.set(
```
KoalaQin, line 1280: I know we're short of time, it took me some time to understand why there are so many groups, now I understand it's a combination of problem, sample's sex and sample's population. 
Since patch1_expr ~ patch4_expr are identical, can we make a small function like this? 
```commandline
def patch_add(freq_groups: hl.set_expr, freq: hl.expr) -> hl.expr:
    return hl.map(
        lambda x, m: hl.if_else(
            freq_groups.contains(m),
            hl.struct(
                AC=x.AC + 1,
                AF=(x.AC + 1) / (x.AN + 2),
                AN=x.AN + 2,
                homozygote_count=x.homozygote_count,
            ),
            x,
        ),
        freq_ht.freq,
        freq_ht.freq_meta,
    )
```

KoalaQin (commented):
It's only a suggestion, but I think the long version is understandable.

gnomad_qc/v4/annotations/generate_freq_genomes.py
jkgoodrich: not at this time please

jkgoodrich: we can refactor after

gnomad_qc/v4/annotations/generate_freq_genomes.py
```diff
@@ -1479,24 +1854,17 @@ def main(args):
...
-        #  freq_meta.
-        downsamplings = sorted(
-            {
-                int(x["downsampling"])
-                for x in hl.eval(ht.freq_meta)
-                if "downsampling" in x
-            }
+        ht = finalize_v4_genomes_callstats(
+            ht, v3_sites_ht, res.v3_meta_ht.ht(), res.updated_meta_ht.ht()
         )
```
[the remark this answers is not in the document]
jkgoodrich: No, but we already did on the other one, so I was trying to keep the code true to what was actually done

gnomad_qc/v4/annotations/generate_freq_genomes.py
KoalaQin: okay, sounds good.

gnomad_qc/v4/annotations/generate_freq_genomes.py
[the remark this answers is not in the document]
KoalaQin: Okay, if we ever need to rerun that step, it will be written there.

KoalaQin (approved):
LGTM!

PR #534, opened 2023-11-09
Fix stats on combined exomes/genomes frequencies

jkgoodrich:
There were some errors for the stats on the combined freq/faf HT that need to be fixed before we can determine if we should release these stats and/or add a flag indicating a significant deviation between the two

### DIFF
--- a/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
+++ b/gnomad_qc/v4/create_release/create_combined_faf_release_ht.py
@@ -277,12 +277,13 @@ def perform_contingency_table_test(
         Default is 5.
     :return: ArrayExpression for contingency table test results.
     """
+    # Using the joint_meta_expr to get the indexes of the two frequency expressions.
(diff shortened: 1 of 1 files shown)


KoalaQin (changes requested):
The code is working, and I only have a few minor suggestions, it seems also much faster to run it with the new non-zero AC filtering, are we regenerating the combined faf release Table and replacing the current one in the browser too?

jkgoodrich:
Yeah we will rerun it. We didn't actually release the stats at all, or the final table this outputs, so we don't need to change anything that is released. I think the browser team uses the values on the exomes and genomes release HTs for the joint numbers, but even if they don't, they were not currently using these stats for anything and that is all that will change

KoalaQin (approved):
Thanks! LGTM!
I double checked a few infinity ORs, I think your logic of getting at least one non-zero AC is great, the infinity OR will indicate one of them being 0.

PR #535, opened 2023-11-21
Remove sift and polyphen from VCF export

mike-w-wilson:
A user found that the number of v4's VCF VEP fields(48) do not match the header(46). This is a result of using default parameters in the function that creates a string out of the vep struct. The default call uses the default VEP_CSQ_FIELDS  list which contains SIFT and Polyphen. It then inserts missing values  in the "validated_ht"  after not finding the fields in the VEP annotation. This "validated_ht" is then used for VCF export. 

This PR fixes the exported VCF fields by removing SIFT and Polyphen from the `csq_fields` passed to the `vep_struct_to_csq` function, resulting in the correct number of vep fields in the VCF.

### DIFF
--- a/gnomad_qc/v4/create_release/validate_and_export_vcf.py
+++ b/gnomad_qc/v4/create_release/validate_and_export_vcf.py
@@ -503,6 +503,8 @@ def prepare_ht_for_validation(
 
     logger.info("Constructing INFO field")
     # Remove SIFT and Polyphen from CSQ fields or they will be inserted with
+    # missing values by vep_struct_to_csq. These fields are processed separately
+    # as in silico annotations.
     csq_fields = "|".join(
         [
             c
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/create_release/validate_and_export_vcf.py
```diff
@@ -497,12 +502,22 @@ def prepare_ht_for_validation(
     )
 
     logger.info("Constructing INFO field")
+    # Remove SIFT and Polyphen from csq fields or they will be inserted with
+    # missing values by vep_struct_to_csq
```
KoalaQin, line 506:
```suggestion
    # Remove SIFT and Polyphen from CSQ fields or they will be inserted with
```

KoalaQin, line 506: Do you want to add a note stating they're extracted from VEP fields and processed separately?

KoalaQin (approved):
LGTM!

PR #537, opened 2023-11-30
Move PCA files to under ancestry_inference

mike-w-wilson:
This change will give access to the pca scores, loadings, and eigenvalue files which will move to under the ancestry_inference subdirectory.


klaricch (approved):
lgtm

PR #539, opened 2023-12-06
v4 subset update to include variant QC

jkgoodrich:


gnomad_qc/v4/subset.py
```diff
@@ -280,11 +341,18 @@ def main(args):
                 overwrite=args.overwrite,
             )
             ht = adjust_vcf_incompatible_types(ht)
-            mt = mt.annotate_rows(info=ht[mt.row_key].info)
+
+            info_expr = ht[mt.row_key].info
```
mike-w-wilson, line 345: So I hit this when I made a subset back in October but VRS is a struct within the info and hails export VCF does not like that. Given the timeline and us being in the middle of production, KC and I decided to drop the annotation. I think we could drop or remove the struct when we are delivering a VCF?

PR #543, opened 2024-01-17
Create variant_statistics.py, per_sample_datatype

matren395:
Create a script in variant_qc/ for variant statistics, first populated with the method per_sample_datatype() to find the number of variants per data type (exomes or genomes)

(does this count as Sample QC or Variant QC? if it's variants per sample? or something else?)

### DIFF
--- a/gnomad_qc/v4/variant_qc/variant_statistics.py
+++ b/gnomad_qc/v4/variant_qc/variant_statistics.py
@@ -1,0 +1,147 @@
+"""Produce per-exome stats, per-genome stats, and variant type counts for v4."""
+import argparse
+import logging
+
+import hail as hl
+from gnomad.assessment.summary_stats import (
+    default_generate_gene_lof_matrix,
+    default_generate_gene_lof_summary,
+    get_summary_counts,
+)
+from gnomad.utils.filtering import filter_to_adj
+from gnomad.utils.slack import slack_notifications
+
+from gnomad_qc.slack_creds import slack_token
+from gnomad_qc.v4.resources.basics import get_gnomad_v4_vds, get_logging_path
+import gnomad_qc.v4.resources.meta as meta
+from gnomad_qc.v4.resources.release import (
+    # release_lof,
+    release_sites,
+    # release_summary_stats
+)
+
+logging.basicConfig(
+    format="%(asctime)s (%(name)s %(lineno)s): %(message)s",
+    datefmt="%m/%d/%Y %I:%M:%S %p",
+)
+logger = logging.getLogger("summary_stats")
+logger.setLevel(logging.INFO)
+
+
+def per_sample_datatype(
+    data_type: str = "exomes",
+    test: bool = False,
+    overwrite: bool = False,
+) -> None:
+    # Read in release data and metadata and VDS
+    ht_release_data = release_sites(data_type).ht()
+    if test:
+        ht_release_data = ht_release_data.filter(
+            ht_release_data.locus.contig == "chr22"
+        )
+    meta_ht = meta(data_type=data_type).ht()
+    vds_data = get_gnomad_v4_vds(test=test, release_only=True)
+    vds_data_mt = vds_data.variant_data
+
+    # Filter to only variants and samples in release data
+    vds_data_filtered = vds_data_mt.filter_rows(
+        ~hl.is_missing(ht_release_data[vds_data_mt.row_key])
+    )
+
+    vds_data_filtered = vds_data_filtered.filter_cols(
+        meta_ht[vds_data_filtered.s].release
+    )
+
+    # Select data down to: only variant locus & alleles and only if samples do or don't have it
+    # Create then select down to GT
+    vds_data_filtered = vds_data_filtered.annotate_entries(GT=hl.vds.lgt_to_gt(vds_data_filtered.LGT, vds_data_filtered.LA))
+    vds_data_filtered = vds_data_filtered.select_entries("GT")
+    vds_data_filtered = vds_data_filtered.select_rows()
+    vds_data_filtered = (
+        vds_data_filtered.select_cols()
+    )
+
+    # Perform Hail's Sample QC module
+    sample_qc_ht = hl.sample_qc(vds_data_filtered).cols().key_by("s")
+
+    sample_qc_ht = sample_qc_ht.checkpoint(
+        f"gs://gnomad-tmp-4day/sample_qc_mt_per_{data_type}_sample.ht", overwrite=overwrite
+    )
+
+    mean_called_per = sample_qc_ht.aggregate(
+        hl.agg.mean(sample_qc_ht.sample_qc.n_called)
+    )
+
+    print("mean called per exome: ", mean_called_per)
+
+
+def variant_types():
+    # what is needed: stats from your notebook for: 
+    # Indels and Snps per data type
+    # How many (number and proportion) pass VQSR and pass All Filters
+    pass
+
+def vep_consequences():
+    # what is needed: stats from notebook for: 
+    # VEP consequence breakdown , in those two different returns
+    # let users specify: pass all filters, pass VQSR, or all variants
+    pass
+
+
+def main(args):
+    """Collect summary stats of gnomAD v4 data."""
+    hl.init(
+        log="/per_exome.log",
+        default_reference="GRCh38",
+        tmp_dir="gs://gnomad-tmp-30day",
+    )
+    # SSA Logs are easier to troubleshoot with.
+    hl._set_flags(use_ssa_logs="1")
+
+    data_type = args.data_type
+    test = args.test
+    overwrite = args.overwrite
+
+    if args.get_per_sample:
+        per_sample_datatype(data_type, test, overwrite)
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument(
+        "-o",
+        "--overwrite",
+        help="Overwrite data (default: False)",
+        action="store_true",
+    )
+    parser.add_argument(
+        "--slack-channel", help="Slack channel to post results and notifications to."
+    )
+    parser.add_argument(
+        "--test",
+        help="Test on a small number of variants",
+        action="store_true",
+    )
+    parser.add_argument(
+        "--get-summary-counts",
+        help="Get summary counts per variant category.",
+        action="store_true",
+    )
+    parser.add_argument(
+        "--get-per-sample",
+        help="Get variants per sample data type.",
+        action="store_true",
+    )
+    parser.add_argument(
+        "--data-type",
+        default="exomes",
+        choices=["exomes", "genomes"],
+        help="Data type (exomes or genomes) to produce summary stats for.",
+    )
+    args = parser.parse_args()
+
+    if args.slack_channel:
+        with slack_notifications(slack_token, args.slack_channel):
+            main(args)
+    else:
+        main(args)


matren395:
the method is currently working for Chr22 with the release test VDS, giving me a result of ~ 2683 variants per 467 samples in Chr22. The world's quickest google search tells me most people have ~4-5million SNPs in their genome , so I'd like to think that my results aren't orders of magnitude off

gnomad_qc/v4/create_release/variant_statistics.py
```diff
@@ -0,0 +1,241 @@
...
+
+    :param data_type: String of either "exomes" or "genomes" for the sample type.
+    :param test: Boolean for if you would like to use a small test set.
+    :param overwrite: Boolean to overwrite checkpoint files if requested.
+    :param filter_variant_qc: Bool to only count variants that pass all variant qc filters (i.e. VQSR, AC0).
+    :param filter_calling_intervals: Bool to filter only to variants in calling intervals.
+    """
+
+    # Read in release data and metadata and VDS
+    ht_release_data = release_sites(data_type).ht()
```
matren395, line 81: currently just locked into the CURRENT_RELEASE - is this ideal ? Should we parametrize this within the script ?

gnomad_qc/v4/create_release/variant_statistics.py
```diff
@@ -0,0 +1,286 @@
...
+
+logging.basicConfig(
+    format="%(asctime)s (%(name)s %(lineno)s): %(message)s",
+    datefmt="%m/%d/%Y %I:%M:%S %p",
+)
+logger = logging.getLogger("release")
+logger.setLevel(logging.INFO)
+
+
+def get_per_sample_variant_counts(
```
jkgoodrich, line 42: Let's move this to gnomad_qc/v4/resources/release.py

gnomad_qc/v4/create_release/variant_statistics.py
```diff
@@ -0,0 +1,286 @@
...
+        GT=hl.vds.lgt_to_gt(vds_data_filtered.LGT, vds_data_filtered.LA)
+    )
+    vds_data_filtered = vds_data_filtered.select_entries("GT")
+    vds_data_filtered = vds_data_filtered.select_rows()
+
+    # Perform Hail's Sample QC module
+    # Resulting table is of most interest for further analysis
+    # Write said table out to determined path
+    sample_qc_ht = hl.sample_qc(vds_data_filtered).cols().key_by("s")
+    sample_qc_ht = sample_qc_ht.checkpoint(
```
jkgoodrich, line 135: I think this function should be split into two so that the first function returns the table, and that can be written out in `main` and the second function performs the below aggregation and logging. Each can also have an argument so if the table exists already the aggregation can be run alone

gnomad_qc/v4/create_release/variant_statistics.py
jkgoodrich: This is good

gnomad_qc/v4/create_release/variant_statistics.py
```diff
@@ -0,0 +1,286 @@
...
+
+    :param data_type: String of either "exomes" or "genomes" for the sample type.
+    :param test: Boolean for if you would like to use a small test set.
+    :param overwrite: Boolean to overwrite checkpoint files if requested.
+    :param filter_variant_qc: Bool to only count variants that pass all variant qc filters (i.e. VQSR, AC0).
+    :param filter_calling_intervals: Bool to filter only to variants in calling intervals.
+    :param singletons_by_pop: Bool to calculate the mean number of singletons by population group.
+    """
+
+    # Read in release data and metadata and VDS
```
jkgoodrich, line 84: Nit pick, add periods after comments.
```suggestion
    # Read in release data and metadata and VDS.
```

gnomad_qc/v4/create_release/variant_statistics.py
```diff
@@ -0,0 +1,286 @@
...
+    ht_release_data = release_sites(data_type).ht()
+    if test:
+        logger.info("Test: filtering to only chr22 sites and 467 test samples in VDS")
+        ht_release_data = ht_release_data.filter(
+            ht_release_data.locus.contig == "chr22"
+        )
+
+    if filter_variant_qc:
```
jkgoodrich, line 92: Instead of having this as an option, lets just add it as a stratification (so I want a number with no filtering and a number with filtering

gnomad_qc/v4/create_release/variant_statistics.py
```diff
@@ -0,0 +1,286 @@
...
+        logger.info("Filter to those which pass all filters")
+        ht_release_data = ht_release_data.filter(hl.len(ht_release_data.filters) == 0)
+
+    if filter_calling_intervals:
```
jkgoodrich, line 96: Same thought as the variant QC, make this a stratification (we would want a number for no calling interval filter, in broad calling intervals, in UKB calling intervals, and in both UKB and Broad intervals), and let's use the 50 bp padding. Or actually, I think the release HT has info about this already that used the 50bp for the annotation so you can probably just use that instead of loading the interval HT.

gnomad_qc/v4/create_release/variant_statistics.py
```diff
@@ -0,0 +1,286 @@
...
+    vds_data_filtered = vds_data_filtered.annotate_entries(
+        GT=hl.vds.lgt_to_gt(vds_data_filtered.LGT, vds_data_filtered.LA)
+    )
+    vds_data_filtered = vds_data_filtered.select_entries("GT")
+    vds_data_filtered = vds_data_filtered.select_rows()
+
+    # Perform Hail's Sample QC module
+    # Resulting table is of most interest for further analysis
+    # Write said table out to determined path
+    sample_qc_ht = hl.sample_qc(vds_data_filtered).cols().key_by("s")
```
jkgoodrich, line 134: I think we probably want to just use the Hail VDS sample QC function instead of the HT one since it was specifically built for the VDS. It will also handle the GT to LGT change

gnomad_qc/v4/create_release/variant_statistics.py
```diff
@@ -0,0 +1,286 @@
...
+    meta_ht = meta(data_type=data_type).ht()
+    vds_data = get_gnomad_v4_vds(test=test, release_only=True)
+    vds_data_mt = vds_data.variant_data
+
+    # Filter to only variants and samples in release data
+    vds_data_filtered = vds_data_mt.filter_rows(
+        ~hl.is_missing(ht_release_data[vds_data_mt.row_key])
+    )
+
+    vds_data_filtered = vds_data_filtered.filter_cols(
```
jkgoodrich, line 115: This is already handled by `release_only=True` above

gnomad_qc/v4/create_release/variant_statistics.py
```diff
@@ -0,0 +1,286 @@
...
+        meta_ht[vds_data_filtered.s].release
+    )
+
+    vds_data_filtered = vds_data_filtered.annotate_cols(
```
jkgoodrich, line 119: With the suggestion below for splitting this function up, this annotation can just be added to the HT in the aggregation function since it isn't needed for computing the per sample stats

jkgoodrich (changes requested):
Some initial larger change requests. I will review the comments and documentation in detail after the code is looking done.

gnomad_qc/v4/create_release/variant_statistics.py
matren395: I believe this would require running hl.vds.sample_qc() twice , once on a VDS filtered to this and once without. Is that something we'd want to support within the script, or just have users run twice ? We can put a pin in this for our chat Wednesday if you'd like

gnomad_qc/v4/create_release/variant_statistics.py
matren395: done with two separate methods - let me know what you think of the approach

gnomad_qc/v4/create_release/variant_statistics.py
matren395: NOTE: not sure this is redundant? At least with the Test VDS, running this line filters the testset from ~550 samples to 467 samples. Same with the new way of filtering it with VDS-native methods. I wonder why that is?

gnomad_qc/v4/create_release/calculate_variant_statistics.py
```diff
@@ -0,0 +1,535 @@
...
+    # Add extra Allele Count and A-Type, according to Hail standards, to help their computation
+    ac_and_atype = vmt_sample_qc_variant_annotations(
+        global_gt=vmt.GT, alleles=vmt.alleles
+    )
+    vmt = vmt.annotate_rows(**ac_and_atype)
+
+    # NOTE: not size efficient, but we are never checkpointing THIS I don't believe
+    vmt = vmt.annotate_rows(**ht_release_data[vmt.row_key])
+
+    if agg_lof:
```
KoalaQin, line 149: This made me think: do you need to get stats for missense, synonymous variants or any specified CSQs?

KoalaQin (changes requested):
I have some high-level comments: 
1. do you mind making smaller functions? Right now each function is a bit long. 
2. all the arguments you pass on variants, I think they should be manipulated on release HT, instead of VDS right away. Maybe it's better to add these annotations to VDS at last. 
3. there are some confusing names that need to be modified. 

I haven't finished the whole thing, could you send me you test run results location and the job link?
(I want to submit the comments for now and will come back to it tomorrow).

gnomad_qc/v4/create_release/variant_statistics.py
KoalaQin: I think it's because the test meta is not updated as the exome final meta. I saw the test meta was created on 2022-04-20 but the exome meta was 2023-07-13, I think it's good enough to use `release_only=True` to filter samples, but we may also update the test meta in another ticket?

KoalaQin: Now you get 581 release_only samples in the test VDS based on the test meta.

jkgoodrich (changes requested):
See [PR](https://github.com/broadinstitute/gnomad_qc/pull/581) for suggested edits

gnomad_qc/v4/create_release/calculate_variant_statistics.py
```diff
@@ -0,0 +1,416 @@
...
+                )
+                for ann, expr in filter_expr.items()
+            }
+        )
+    ).cols()
+
+    return ht.select(**ht._sample_qc)
+
+
+def compute_agg_sample_stats(
```
jkgoodrich, line 166: @KoalaQin This is what I was thinking about when I told you I had a plan for what I want this second part of the script to look like, but I haven't incorporated it in yet 

```
all_strats = [s for s in ht.row_value if isinstance(ht[s], hl.expr.StructExpression)]
ht = ht.transmute(
    gen_anc=["all_samples", meta_ht[ht.s].population_inference.pop],
    _stats_array=[(s, ht[s]) for s in all_strats]
)
ht = ht.explode("_stats_array").explode("gen_anc")
ht = ht.group_by("gen_anc", variant_filter=ht._stats_array[0]).aggregate(
    **{
        m: hl.struct(
            mean=hl.agg.mean(ht._stats_array[1][m]),
            quantiles=hl.agg.approx_quantiles(
                ht._stats_array[1][m] , [0.0, 0.25, 0.5, 0.75, 1.0]
            ),
        )
        for m in ht._stats_array[1] 
        if isinstance(ht._stats_array[1][m], hl.expr.NumericExpression)
    }
)
```

it still needs to support the by_ancestry and change the return type, and be written out to a new HT instead of added as globals

KoalaQin (commented):
@matren395 I think you need to merge from main first and only show the difference on those 4 files we mades changes on. Then we can tag Julia for the final pass of review.

KoalaQin:
@jkgoodrich Here is our final version of code, this branch was too old and my merge got too many commits there, sorry, you can look at #586 if you want to see the difference compared to your last review.

matren395:
Creating a really simple branch to return SET (not variable) per-callset calculations . I mean WE can change them NOW, but they won't be args

PR #544, opened 2024-01-17
Calculate VEP, SNP, and other counts for v4.1

matren395:
Make further changes to my variant_statistics file to calculate the VEP & LOF numbers and SNP-Indel numbers for v4.1 . This was already done for v4, but we would like it as a script in a repo for v4.1 . With this, it should be very easy to do!


matren395:
Hmm so this is very outdated - easier to nest inside of PR #543 with the separate branched PR #611 for what we'd want to calculate. May be the place for that and close this one out!

jkgoodrich:
Closing this out since it is very outdated

PR #547, opened 2024-01-19
Change to use v3 test VDS instead of sparse MT

jkgoodrich:


matren395 (approved):
makes sense enough and all the code looks right - could you walk me through why this change is being made again before we approve it though ?

jkgoodrich:
Yup, this is just to use the new VDS of the test dataset instead of the sparse MT since we are moving completely to VDS it's more helpful to have the genomes test dataset available as a VDS

PR #548, opened 2024-01-19
Update faf and popmax pops to remove to account for change in gnomad_methods

mike-w-wilson:
This is to update the code to work with: https://github.com/broadinstitute/gnomad_methods/pull/658/files.

I had to change the VCF export code for v4 beyond just referencing the version since the code works for both genomes and exomes but genomes uses v3 faf pops while exomes will use v4. This is not pretty but its the least amount of code to add without major restructuring. We should probably rethink these global variables.

### DIFF
--- a/gnomad_qc/v4/create_release/validate_and_export_vcf.py
+++ b/gnomad_qc/v4/create_release/validate_and_export_vcf.py
@@ -552,6 +552,7 @@ def prepare_ht_for_validation(
 def populate_subset_info_dict(
     subset: str,
     description_text: str,
+    data_type: str = "exomes",
     pops: Dict[str, str] = POPS,
     faf_pops: Dict[str, str] = FAF_POPS,
     sexes: List[str] = SEXES,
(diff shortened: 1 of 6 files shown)


gnomad_qc/v4/create_release/validate_and_export_vcf.py
```diff
@@ -666,7 +663,7 @@ def populate_info_dict(
     :param subset_pops: Dict of sample global genetic ancestry group names to use for all subsets in `subset_list` unless the subset
         is 'gnomad', in that case `gnomad_pops` is used. Default is POPS.
     :param gnomad_pops: Dict of sample global genetic ancestry group names for gnomAD data type. Default is POPS.
-    :param faf_pops: Dict with faf genentic ancestry group names (keys) and descriptions (values).  Default is FAF_POPS.
+    :param faf_pops: Dict with gnomAD version (keys) and faf genentic ancestry group names(values).  Default is FAF_POPS.
```
klaricch, line 666: won't make a functional difference here, but for "populate_subset_info_dict", should the default  for 'faf_pops: Dict[str, str] = FAF_POPS' be changed to FAF_POPS["v4"] so that the types match?

gnomad_qc/v4/create_release/validate_and_export_vcf.py
```diff
@@ -1038,16 +1033,18 @@ def main(args):  # noqa: D103
...
+                # logger.info("Filtering to PCSK9 region...")
+                # # Keep only PCSK9.
+                # ht = hl.filter_intervals(
+                #     ht,
+                #     [
+                #         hl.parse_locus_interval(
+                #             "chr1:55039447-55064852", reference_genome="GRCh38"
+                #         )
+                #     ],
+                # )
```
klaricch, line 1047:
```suggestion
```

gnomad_qc/v4/create_release/validate_and_export_vcf.py
```diff
@@ -1022,13 +1024,6 @@ def main(args):  # noqa: D103
 
         if args.export_vcf:
             contig = f"chr{contig}" if contig else None
-
-            if contig and test:
```
klaricch, line 1026: why drop this if statement?

klaricch (changes requested):

gnomad_qc/v4/create_release/validate_and_export_vcf.py
mike-w-wilson: Added this back and the contig is now below the error

gnomad_qc/v4/create_release/validate_and_export_vcf.py
mike-w-wilson: Goodp oint, I'm going to change where the FAF_POPS get processed, doesnt make sense to do it here since only subset uses it.

mike-w-wilson:
Back to you @klaricch ! I've updated where FAF_POPS gets processed based on your comment and I think its better there since it was just being processed and then passed to another function. This way is a bit cleaner so thank you!

gnomad_qc/v4/create_release/validate_and_export_vcf.py
```diff
@@ -552,6 +552,7 @@ def prepare_ht_for_validation(
 def populate_subset_info_dict(
     subset: str,
     description_text: str,
+    data_type: str = "exomes",
     pops: Dict[str, str] = POPS,
     faf_pops: Dict[str, str] = FAF_POPS,
```
klaricch, line 557:
```suggestion
    faf_pops: Dict[str, List[str]] = FAF_POPS,
```

klaricch (changes requested):
just one small edit

klaricch (approved):
lgtm!

PR #552, opened 2024-02-05
Adding my clean-up and changes for the AN and hist fixes

jkgoodrich:


gnomad_qc/v4/annotations/compute_coverage.py
```diff
@@ -126,6 +126,7 @@ def adjust_interval_padding(ht: hl.Table, padding: int) -> hl.Table:
             ht.interval.start.position - padding,
             ht.interval.end.position + padding,
             reference_genome=ht.interval.start.dtype.reference_genome,
+            includes_end=True,
```
mike-w-wilson, line 129:
```suggestion
            # Include the end of the intervals to capture all variants 
            reference_genome=ht.interval.start.dtype.reference_genome,
            includes_end=True,
```

PR #554, opened 2024-02-09
Add support for including mermaid flowcharts at the top of the sphinx documentation

jkgoodrich:
As of now I have only added one very simple flowchart, but others can be added to the docs/flowcharts directory to be included at the top of each sphinx documentation page.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -21,7 +21,8 @@ extensions = [
     "sphinx.ext.viewcode",
     "sphinx_autodoc_typehints",
     "myst_parser",
-    'sphinxarg.ext',
+    "sphinxarg.ext",
+    "sphinxcontrib.mermaid",
 ]
 
 # README.md contains developer documentation for building docs
--- a/docs/flowcharts/v4/index.mmd
+++ b/docs/flowcharts/v4/index.mmd
@@ -1,0 +1,8 @@
+flowchart LR;
+  sample_qc["<a class="reference internal" href='sample_qc/index.html'>sample_qc</a>"];
+  annotations["<a class="reference internal" href='annotations/index.html'>annotations</a>"];
+  variant_qc["<a class="reference internal" href='variant_qc/index.html'>variant_qc</a>"];
+  create_release["<a class="reference internal" href='create_release/index.html'>create_release</a>"];
+  sample_qc --> annotations;
+  annotations --> variant_qc;
+  variant_qc --> create_release;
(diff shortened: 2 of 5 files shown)


docs/generate_api_reference.py
```diff
@@ -35,6 +77,27 @@ def module_doc_path(module):
...
 
+def module_flowchart_path(module, local=False):
+    """
+    Get the path for a module's mermaid flowchart file.
+
+    If local is True, returns the absolute path to the file.
+    If local is False, returns the path relative to the module's documentation file.
+
+    :param module: Module to get flowchart path for.
+    :param local: Whether to return the absolute path to the file.
```
ch-kr, line 88:
```suggestion
    :param local: Whether to return the absolute path to the file.
        Default is False.
```

ch-kr: basic question: why do you need both options?

docs/generate_api_reference.py
```diff
@@ -45,6 +108,28 @@ def package_doc_path(package):
...
 
+def package_flowchart_path(package, local=False):
+    """
+    Get the path for a package's mermaid flowchart file.
+
+    If local is True, returns the absolute path to the file.
+    If local is False, returns the path relative to the package's documentation file.
+
+    :param package: Package to get flowchart path for.
+    :param local: Whether to return the absolute path to the file.
```
ch-kr, line 119:
```suggestion
    :param local: Whether to return the absolute path to the file.
        Default is False.
```

ch-kr (commented):
two minor suggestions and questions

docs/conf.py
```diff
@@ -21,7 +21,8 @@
     "sphinx.ext.viewcode",
     "sphinx_autodoc_typehints",
     "myst_parser",
-    'sphinxarg.ext',
+    "sphinxarg.ext",
+    "sphinxcontrib.mermaid",
```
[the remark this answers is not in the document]
jkgoodrich: me either! But I decided to look it up and realized it will be better to just add them there instead of in the READMEs

docs/flowcharts/v4/index.mmd
```diff
@@ -0,0 +1,8 @@
+flowchart LR;
```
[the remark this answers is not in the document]
jkgoodrich: It's technically wrapped in one, but within the docs instead of within a markdown file

docs/generate_api_reference.py
jkgoodrich: I use the absolute path to check for existence of the .mmd file and the relative path within the .rst files that create the HTMLs. If a package/module doesn't have an mmd file then no flowchart is added to the docs

ch-kr (approved):
LGTM -- can't wait to see this go live!!

PR #555, opened 2024-02-09
First step for building the mermaid flow chart is to parse all the resources and add them as nodes to the flow chart

jkgoodrich:
This is just the first step for building the flow chart. It adds all of the nodes that refer to resources. It also adds helpful node attributes like a link to the gnomad_qc resource function in the sphinx documentation, node type, node color, node display name, node id to put in the flowchart code.

There is a very basic outline for the QCFlowchart class that will be filled in more as I add things to this

### DIFF
--- a/docs/generate_mermaid_flowcharts.py
+++ b/docs/generate_mermaid_flowcharts.py
@@ -1,0 +1,417 @@
+import networkx as nx
+import pkgutil
+import inspect
+import itertools
+import os
+import re
+import sys
+import pathlib
+import importlib
+import setuptools
+
+import hail as hl
+
+from gnomad.resources.resource_utils import (
+    BaseResource,
+    DataException,
+    BaseVersionedResource,
+)
+from gnomad.utils.file_utils import check_file_exists_raise_error
+
+# Need to add the repository root to the path before importing any gnomad_qc modules.
+REPOSITORY_ROOT_PATH = str(pathlib.Path(os.path.abspath(__file__)).parent.parent)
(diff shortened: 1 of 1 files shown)


docs/generate_mermaid_flowcharts.py
```diff
@@ -0,0 +1,417 @@
...
+        elif isinstance(resource, str):
+            paths = [resource]
+        else:
+            err_msg = "" if bind_kwargs is None else f"with {bind_kwargs}"
+            raise ValueError(
+                f"Resource {self.res_obj} {err_msg}did not return a valid resource."
+            )
+
+        for p in paths:
+            if check_file_exists_raise_error(p, error_if_exists=False):
```
ch-kr, line 285: should this have `error_if_not_exists=True`?

docs/generate_mermaid_flowcharts.py
```diff
@@ -0,0 +1,417 @@
...
+                res_bind = res_sig.bind(**dict(bind_kwargs))
+
+                try:
+                    self.add_resource_to_flowchart(
+                        flowchart,
+                        self.res_obj(*res_bind.args, **res_bind.kwargs),
+                        bind_kwargs=bind_kwargs,
+                    )
+                except (ValueError, DataException, KeyError):
+                    continue
```
ch-kr, line 314: should this have a warning message (since it's currently a silent failure)? or are you expecting that some resources will fail

ch-kr (commented):
a couple minor questions

PR #561, opened 2024-02-26
Add checkpoints to ref HT for all sites AN and use v4.0 genomes metadata

jkgoodrich:


mike-w-wilson (approved):
I think I agree with splitting this function. The exome VDS one is pretty substantial already and we'd need to have a bunch of parameter checks if we merged them so separating here makes sense.

gnomad_qc/v4/resources/basics.py
```diff
@@ -67,6 +68,7 @@ def get_gnomad_v4_vds(
         removing withdrawn UKB samples. Default is True.
     :param annotate_meta: Whether to annotate the VDS with the sample QC metadata.
         Default is False.
+    :param data_type: One of 'exomes' or 'genomes'. Default is 'exomes'.
```
jkgoodrich, line 71:
```suggestion
```

PR #563, opened 2024-02-28
Add all sites AN export to TSV

jkgoodrich:
You can see the genomes TSV output here: gs://gnomad/release/4.1/tsv/genomes/gnomad.genomes.v4.1.allele_number.tsv.bgz

### DIFF
--- a/gnomad_qc/v4/annotations/compute_coverage.py
+++ b/gnomad_qc/v4/annotations/compute_coverage.py
@@ -303,7 +303,7 @@ def get_coverage_resources(
         },
     )
     export_coverage_files = PipelineStepResourceCollection(
-        "--export-release-files",
+        "--export-coverage-release-files",
         output_resources={
             "coverage_tsv": release_coverage_tsv_path(data_type, test=test),
             "release_ht": release_coverage(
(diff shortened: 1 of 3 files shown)


gnomad_qc/v4/resources/release.py
```diff
@@ -329,19 +329,36 @@ def release_coverage_tsv_path(
...
+    """
+    Fetch path to all sites AN TSV file.
+
+    :param data_type: 'exomes' or 'genomes'. Default is 'exomes'.
+    :param release_version: Release version. Default is
+        CURRENT_ALL_SITES_AN_RELEASE["exomes"].
+    :param test: Whether to use a tmp path for testing. Default is False.
+    :return: All sites AN TSV path.
+    """
+    return (
```
mike-w-wilson, line 360: MAybe assign release version in the function? Default to None on L348, then sometting like
```suggestion
    release_version = release_version if release_version is not None else CURRENT_ALL_SITES_AN_RELEASE[data_type]
    return (
```

gnomad_qc/v4/annotations/compute_coverage.py
```diff
@@ -312,13 +313,21 @@ def get_coverage_resources(
         },
         pipeline_input_steps=[compute_coverage_ht],
     )
+    export_an_tsv = PipelineStepResourceCollection(
+        "--export-all-sites-an-release-tsv",
+        output_resources={
+            "allele_number_tsv": release_all_sites_an_tsv_path(data_type, test=test)
```
mike-w-wilson, line 319: Wont the release version always be the `CURRENT_ALL_SITES_AN_VERSION["exomes"]`. I know they are in sync now but should probably adjust the function call or function itself?

mike-w-wilson (changes requested):
Just one thing on the tsv path call/function

gnomad_qc/v4/annotations/compute_coverage.py
jkgoodrich: yup, thanks for catching. I fixed this for the other functions in my combined freq PR, but forgot to change it here

mike-w-wilson (approved):

PR #566, opened 2024-03-06
v3 subset script

mike-w-wilson:
This is the most updated subset script. I went ahead and removed subset specific callstats. We've deprioritized the old subset script for years and eventually moved it into a draft after it became outdated but we should get this in. This is the script, with the exception of the subset callstats removal, that was last run on v3 subset requests.

### DIFF
--- a/gnomad_qc/v3/subset.py
+++ b/gnomad_qc/v3/subset.py
@@ -216,6 +216,11 @@ if __name__ == "__main__":
         action="store_true",
     )
     parser.add_argument(
+        "--test",
+        help="Filter to the first 2 partitions for testing.",
+        action="store_true",
+    )
+    parser.add_argument(
         "--dense-mt", help="Whether to make a dense MT.", action="store_true"
     )
     parser.add_argument(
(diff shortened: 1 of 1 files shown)


gnomad_qc/v3/subset.py
```diff
@@ -0,0 +1,291 @@
...
+            if add_variant_qc or pass_only:
+                vd = vds.variant_data
+                ht = release_sites().ht()
+                if pass_only:
+                    logger.info("Filtering to variants that passed variant QC.")
+                    vd = vd.filter_rows(ht[vd.row_key].filters.length() == 0)
+                if add_variant_qc:
+                    logger.info("Adding variant QC annotations.")
+                    vd = vd.annotate_rows(
+                        **make_variant_qc_annotations_dict(
```
KoalaQin, line 146: This function from v4 is using v4 release_sites HT, but `raw_qual_hists` is now under `histograms`, maybe copy that function here, and use v3 release_sites HT?
It will run into error otherwise.

KoalaQin (changes requested):
Is the subsetting only working on the subsets already exist in freq_meta or any list of samples? It seems missing to me if you are extracting call_stats for a subset in the code.

KoalaQin (commented):
I'm checking a subsetting code for Cal and Bob, I found that if they are subsetting some non-releasable samples, I don't think we can still use release HT to get variant QC annotations, because some variants won't be in release HT>

PR #567, opened 2024-03-06
Add final release schema check

mike-w-wilson:
We're removing joint freq and associated field from the v4.1 release. This doesnt need too much of a code change as the script was built for adding or removing tables in the join but there was one spot where I missed flexibility beforehand, setting the final schema order. This update checks if the joined HT fields exist in the defined FINALIZED_SCHEMA and removes them from the ordering if they do not, while also throwing a warning that they are not present.


mike-w-wilson:
I tested these updates with 
```
hailctl dataproc submit mw1 gnomad_qc/v4/create_release/create_release_sites_ht.py --test --data-type exomes --tables-for-join dbsnp filters freq info region_flags in_silico vep 
```

klaricch (approved):
lgtm

PR #569, opened 2024-03-07
Small script to generate v4.1 genomes release HT

mike-w-wilson:
This is a very small script that simply drops the joint annotations from the genomes release HT and reannotates the version and date.


mike-w-wilson:
Closing in favor of recreating the genomes release HT using the create release HT script after we realized frequencies would need to be fixed as well after Julia discovered an AF bug

PR #570, opened 2024-03-07
False Duplication VCF Export 

matren395:
Code to take False Duplication (of three chr21 genes) Hail Table and convert it to a VCF, verify, and export. 

03-07-24: still needs verification and header added, but just wanted to get the PR opened

### DIFF
--- a/gnomad_qc/v4/create_release/false_dup_validate_and_export_vcf.py
+++ b/gnomad_qc/v4/create_release/false_dup_validate_and_export_vcf.py
@@ -1,0 +1,391 @@
+"""Script to validate and export only the false duplication Hail Table of 3 genes to a VCF."""
+import argparse
+import logging
+
+import hail as hl
+from gnomad.utils.annotations import (
+    faf_expr,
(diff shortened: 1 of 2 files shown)


matren395:
still some to do, but unfurling and the existing header code could use a lookover now

gnomad_qc/v4/create_release/false_dup_validate_and_export_vcf.py
```diff
@@ -0,0 +1,671 @@
...
+    #     # Add VRS annotations to info_dict.
+    #     vcf_info_dict.update(vrs_fields_dict)
+
+    return vcf_info_dict
+
+
+def main(args):
+    ht = hl.read_table(get_false_dup_genes_path(release_version="4.0"))
+    ht = prepare_false_dup_ht_for_validation(ht)
+    header_dict = {
```
klaricch, line 650: where does header_dict get used? why call prepare_vcf_filter_header twice?

klaricch (changes requested):

PR #573, opened 2024-03-08
Temp PR for fixing for genome freq HT

jkgoodrich:
Set the allele frequency to missing for any call stats entries with AN == 0

Also added a PR that will be merged to main and performs this modification where it should have been done


mike-w-wilson:
This makes sense!

PR #582, opened 2024-03-18
Add region flags for joint exome genome v4.1

jkgoodrich:


mike-w-wilson (commented):
One thing on the data_type annotations. Also this is pure preference, but I find code easier to read/follow when all internal private methods are stacked at the top of the methods instead of imbedded in the methods code. If you like it embedded, I think its fine as is, I get its closer to the code that calls it but  my brain has a little more difficulty following the flow through main function with it this way. I also recognize we havent established any style/consistency with how we do this so not a major sticking point for me.

mike-w-wilson (approved):
:shipit:

PR #585, opened 2024-03-21
Add region flags to exomes all sites AN

jkgoodrich:
Job: 26b81963ebd84a9498a55bb4c46debc6

### DIFF
--- a/gnomad_qc/v4/annotations/compute_coverage.py
+++ b/gnomad_qc/v4/annotations/compute_coverage.py
@@ -636,8 +636,8 @@ def get_script_argument_parser() -> argparse.ArgumentParser:
         action="store_true",
     )
     parser.add_argument(
-        "--export-all-sites-an-release-tsv",
-        help="Export all sites AN release file.",
+        "--export-all-sites-an-release-files",
+        help="Export all sites AN release HT and TSV file.",
         action="store_true",
     )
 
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/annotations/compute_coverage.py
```diff
@@ -515,12 +533,22 @@ def main(args):
...
-            logger.info("Exporting all sites AN tsv...")
-            res = coverage_resources.export_all_sites_an_release_tsv
+        if args.export_all_sites_an_release_files:
+            logger.info("Exporting all sites AN HT...")
+            res = coverage_resources.export_all_sites_an_release_files
             res.check_resource_existence()
-            ht = res.allele_number_ht.ht()
-            ht = ht.select(AN=ht.AN[0])
+
+            ht = res.an_qual_hist_ht.ht()
```
KoalaQin, line 541: One basic question, I was looking for input resources, it seems that you used AN_qual_hist HT to get a new AN HT with regions flags now, why did we output a AN HT before separately?

KoalaQin (commented):
LGTM, I just double checked if the number of variants outside capture/calling region makes sense, I saw it's consistent trend as in the exome release HT, in the all sites AN HT, there are fewer variants outside Broad calling regions than outside UKB calling regions.

gnomad_qc/v4/annotations/compute_coverage.py
jkgoodrich: I need the qual hists for the combined freq creation, but we only want to release the all sites AN, we don't need to release the qual hists

KoalaQin (approved):
LGTM!

PR #586, opened 2024-03-21
Add LOF-specific consequences and desired unions to code

matren395:
Calculate: 
- Consequences (Mis, Syn, and Lof-consequences) with passing_vqc
- Consequences (Mis, Syn, and Lof-consequences) with passing_vqc and broad_ukb union
- passing_vqc and broad_ukb union
- DO NOT CALCULATE CONSEQUENCES on their own


KoalaQin (approved):
LGTM! @jkgoodrich Could you go over and then we merge it to #543?

PR #593, opened 2024-04-04
Add validate check to joint HT before vcf export

KoalaQin:
This is for now only checking the missingness.

### DIFF
--- a/gnomad_qc/v4/create_release/export_joint_vcf.py
+++ b/gnomad_qc/v4/create_release/export_joint_vcf.py
@@ -38,6 +38,7 @@ from gnomad_qc.v4.resources.release import (
     release_header_path,
     release_sites,
     release_vcf_path,
+    validated_release_ht,
 )
 
 logging.basicConfig(format="%(levelname)s (%(name)s %(lineno)s): %(message)s")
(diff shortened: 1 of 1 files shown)


KoalaQin:
The code is working if either of you could have a look. Test run: [0b51afc812b64aee81096fa94bec4f38](https://console.cloud.google.com/dataproc/jobs/0b51afc812b64aee81096fa94bec4f38?region=us-central1&project=broad-mpg-gnomad)
With cluster setup: 
```
hailctl dataproc start qh1 \
    --requester-pays-allow-all \
    --region=us-central1 \
    --zone="" \
    --tags=dataproc-node,ssh-broad \
    --master-boot-disk-size=300 \
    --packages="ipython<8.22,git+https://github.com/broadinstitute/gnomad_methods.git@qh/validata_joint,git+https://github.com/broadinstitute/gnomad_qc@main" \
    --no-off-heap-memory \
    --max-idle 60m
```

KoalaQin:
@jkgoodrich in my test, I found that comparison of `ami` is all missing, I'm looking into your code, to see if it's right for me to unfurl it with the joint freq index, even though it seems to have been excluded from the cmh test. Maybe you have a quick answer for me?

mike-w-wilson:
@KoalaQin, the Amish genetic ancestry group is only present in genomes so there cannot be a comparison.

KoalaQin:
@mike-w-wilson Then do we need to delete them from the struct or just keep them missing?

PR #600, opened 2024-04-12
Create a unionized p-value for freq comparison

jkgoodrich:
Test run: b08870cb9260410596f703ce880784d3


KoalaQin:
LGTM, in the test, 122 variants had p_value<10e-4 compared to 234 before using cmh p-value.

KoalaQin (approved):
LGTM, in the test, 122 variants had p_value<10e-4 compared to 234 before using cmh p-value, 16 of them from 1-pop.

PR #602, opened 2024-04-12
Add filters to combined frequency HT

jkgoodrich:
Test run job: 33f33c3dda7d40418d86e081361688f3


mike-w-wilson (approved):
One small docstring change then this looks good

PR #605, opened 2024-04-16
Fixes to validate joint release HT

jkgoodrich:
- Fix POPS to use correct pops for genomes
- Change filters handling for exomes and genomes when data_type==joint

### DIFF
--- a/gnomad_qc/v4/create_release/validate_and_export_vcf.py
+++ b/gnomad_qc/v4/create_release/validate_and_export_vcf.py
@@ -1141,8 +1141,7 @@ def main(args):
                         "monoallelic": dt_ht.info.monoallelic,
                         "only_het": dt_ht.info.only_het,
                     }
-                # TODO: Should we be adding filtering info to the joint HT?
-                dt_ht.describe()
+
                 validate_release_t(
                     dt_ht,
                     subsets=SUBSETS[data_type],
--- a/gnomad_qc/v4/create_release/validate_and_export_vcf.py
+++ b/gnomad_qc/v4/create_release/validate_and_export_vcf.py
@@ -643,7 +643,7 @@ def prepare_ht_for_validation(
         )
 
     # Select relevant fields for VCF export
-    if for_joint:
+    if data_type == "joint":
         ht = ht.select("info", filters=hl.missing(hl.tset(hl.tstr)))
     else:
         ht = ht.select("info", "filters", "rsid")
@@ -1149,7 +1149,6 @@ def main(args):
                     variant_filter_field="AS_VQSR",
                     problematic_regions=REGION_FLAG_FIELDS[data_type],
                     single_filter_count=True,
-                    filters_check=not for_joint,
                 )
                 if for_joint:
                     dt_ht = dt_ht.annotate(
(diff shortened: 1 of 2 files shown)


KoalaQin (commented):
I'm running validate test on `exomes` and `genomes` too, somehow it's breaking, I'm identifying the problem. Give me a moment.

mike-w-wilson (commented):
One thing on the select for non joint data types

gnomad_qc/v4/create_release/validate_and_export_vcf.py
```diff
@@ -1136,34 +1140,41 @@ def main(args):
                         "monoallelic": dt_ht.info.monoallelic,
                         "only_het": dt_ht.info.only_het,
                     }
-                # TODO: Should we be adding filtering info to the joint HT?
+
```
KoalaQin, line 1143: After L1133, you need to add: 
```commandline
                    for_joint_validation=(dt == "joint"),
```

KoalaQin (changes requested):

gnomad_qc/v4/create_release/validate_and_export_vcf.py
KoalaQin: Because it was set to `True` by default, or maybe we should set default to `False`.

gnomad_qc/v4/create_release/validate_and_export_vcf.py
```diff
@@ -642,7 +643,7 @@ def prepare_ht_for_validation(
         )
 
     # Select relevant fields for VCF export
-    if for_joint:
+    if data_type == "joint":
         ht = ht.select("info", filters=hl.missing(hl.tset(hl.tstr)))
     else:
         ht = ht.select("info", "filters", "rsid")
```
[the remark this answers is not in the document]
jkgoodrich: sorry, this is fixed now

gnomad_qc/v4/create_release/validate_and_export_vcf.py
jkgoodrich: I think `for_joint_validation` is supposed to be True for all of these. I was a bit confused by it at first, which is why I added the data_type change originally, but I think I have it right now. @mike-w-wilson might know best though

KoalaQin: No, this is needed when we're validate `exomes` and `genomes` release HT, because it's set to True, we need it to be false, so we have other INFO fields included.

jkgoodrich: oh, I see what you are saying, but what we need is `for_joint_validation=for_joint,`

jkgoodrich: if `dt` is "exomes" or "genomes", but `for_joint` (`data_type=="joint") we want `for_joint_validation` to still be True

mike-w-wilson: Yeah, this was how we can differentiate if were validating exomes and genomes for release or validating them within the joint HT.

KoalaQin: Yes, thanks, my brain is not so sharp today...

mike-w-wilson (approved):
This looks good to me, I tested exomes and am considering joint fully tested since we've reviewed output.

KoalaQin:
I'm just starting a cluster with your gnomad_methods PR for joint test.

KoalaQin (approved):
LGTM!

PR #606, opened 2024-04-16
Change the renaming of joint validate HT fields for VCF

jkgoodrich:
Test: eff2565dc9634e70a7166948cf1ca5c7

It would need a fix for ordering, but just a proposal

### DIFF
--- a/gnomad_qc/v4/create_release/validate_and_export_vcf.py
+++ b/gnomad_qc/v4/create_release/validate_and_export_vcf.py
@@ -1215,12 +1215,7 @@ def main(args):
                         *[f for f in info_expr if f not in in_joint_ht]
                     )
                     ht = ht.annotate(info=ht.info.annotate(**info_expr))
-                    ht = ht.annotate_globals(
-                        **{
-                            f"{dt}_{f}": validate_hts[dt].globals[f]
-                            for f in validate_hts[dt].globals
-                        }
-                    )
+                    ht = ht.annotate_globals(**validate_hts[dt].index_globals())
 
             ht.describe()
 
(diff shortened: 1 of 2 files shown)


KoalaQin:
I like your renaming idea, which will the header and export easier. I have some suggestions in PR #607

mike-w-wilson (changes requested):
One thing with adding the exome and genome globals to the validated HT output but once that is in, LGTM! @KoalaQin , I presume you'll be taking this work over from Julia?

KoalaQin:
Yep, happy to take over! Give me a moment.

gnomad_qc/v4/create_release/validate_and_export_vcf.py
```diff
@@ -1170,12 +1208,19 @@ def main(args):
...
+                    info_expr = validate_hts[dt][ht.key].info
+                    info_expr = info_expr.select(
+                        *[f for f in info_expr if f not in in_joint_ht]
+                    )
+                    ht = ht.annotate(info=ht.info.annotate(**info_expr))
+                    ht = ht.annotate_globals(
+                        **{
+                            f"{dt}_{f}": validate_hts[dt].globals[f]
+                            for f in validate_hts[dt].globals
+                        }
```
mike-w-wilson, line 1222: This is already inside a datatype for loop and globals have the prefix
```suggestion
                    ht = ht.annotate_globals(**validate_hts[dt].index_globals())
```

mike-w-wilson (changes requested):

gnomad_qc/v4/create_release/validate_and_export_vcf.py
KoalaQin: It was under testing, now should work. Keep forgetting `index_globals()`.

mike-w-wilson:
I've tested this on all datatypes and the output validated HT looks good!

mike-w-wilson (approved):
LGTM

mike-w-wilson:
Let's not run this until all code updates are in for the joint VCF export in case we find something we need to update when writing the header and export code.

PR #607, opened 2024-04-18
Move quality histogram and keep order while renaming

KoalaQin:
Changes are: 
1. Move histograms to unfurl and add data_type to renaming; 
2. Keep order in `info` while renaming.

### DIFF
--- a/gnomad_qc/v4/create_release/validate_and_export_vcf.py
+++ b/gnomad_qc/v4/create_release/validate_and_export_vcf.py
@@ -294,6 +294,7 @@ def unfurl_nested_annotations(
     entries_to_remove: Set[str] = None,
     data_type: str = "exomes",
     joint_included: bool = False,
+    hist_prefix: str = "",
     freq_comparison_included: bool = False,
     for_joint_validation: bool = False,
 ) -> [hl.expr.StructExpression, Set[str], Dict[str, str]]:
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/create_release/validate_and_export_vcf.py
```diff
@@ -484,7 +518,6 @@ def unfurl_nested_annotations(
 
 def make_info_expr(
     t: hl.Table,
-    hist_prefix: str = "",
```
mike-w-wilson, line 487: 495 should be dropped as well

mike-w-wilson (approved):
One thing, then this looks good to me! Where you put the hists makes more sense to me, not sure why they existed outside of the unfurl method in the first place 🤷 .  I wanted to review before Julia's because I found the hists werent being handled there so lets get this in to her branch and then I'll continue my review. One note, well need to reorder again in the VCF as we want certain fields to start in the info field. Not sure if that changes your approach here but I'm fine updating this part in the other PR.

gnomad_qc/v4/create_release/validate_and_export_vcf.py
KoalaQin: L495? The CTT?

KoalaQin: I got what you mean, old L495, the docstring.

KoalaQin:
Could you merge if this LGTY?

PR #611, opened 2024-05-01
Calculate per callset stats on gnomAD v4.1 per sample variant counts

matren395:
IF WE WANT, calculate a set number of per callset stats from the whole of the gnomAD v4.1 per-sample variant count tables. Simple hl.agg.sum(). 

Would be very easy to also calculate it on all top-level stratifications or for all stats with type 'int64' . However, just counting what I know we want to count here. 

A lot of the work from prior PRs are already included in this PR, so something small like this should be all we need. LIkely!

We still need a scientific high-level chat if we want to calculate these though. This was very easy to put together!

### DIFF
--- a/gnomad_qc/v4/assessment/calculate_per_sample_stats.py
+++ b/gnomad_qc/v4/assessment/calculate_per_sample_stats.py
@@ -258,7 +258,7 @@ def create_per_sample_counts_ht(
     return ht
 
 
-def compute_per_callset_stats(ht: hl.Table) -> hl.Struct:
+def compute_per_callset_stats(ht: hl.Table, all_stats: bool = False) -> hl.Struct:
     """
     Compute a set number of per-callset variant statistics.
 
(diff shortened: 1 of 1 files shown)


matren395:
Previously, I'd attempted to .select() to relevant strats and then .flatten() , then take the sum of every row, but Hail really wasn't cooperating (while keyed by 's' ?) , and the code to calculate the sum of every row would have been as much as the present aggregations. 

Let me know if you see an obvious way to make this more efficient, or if something more efficient is needed here

matren395:
I think I'd know the logic for this if it were a matrix table or if it was keyed by the stratification (not the sample_id) but Hail isn't cooperating in the alloted time I've got for looking into this

matren395:
Ugh pumping the brakes on this implementation - calculating from the 'per sample' table will count the total INSTANCES of these different counts in gnomAD v4.1 (on the entry-level), not the amount of these VARIANTS in gnomAD v4.1 (on the row level). 

It's not "double counting" if we WANT to count entries and not variants or rows. 

These counts could be cool to have, and I think they'd be informative (while we're breaking it down here anyways) , but it's not the traditional summary stats.

PR #616, opened 2024-05-22
Code to run logistic regression on v4 exomes and genomes with ancesty pcs

KoalaQin:


gnomad_qc/v4/assessment/logistic_regression.py
```diff
@@ -0,0 +1,179 @@
...
+
+    ht = release_sites(data_type="joint").ht()
+    # Maybe only filter to variants with a union p-value < 10e-4 later
+
+    if test:
+        logger.info("Filtering to test intervals...")
+        ht = get_test_intervals(ht)
+        ht = ht.checkpoint(hl.utils.new_temp_file("test_intervals", "ht"))
+        exomes_vds = hl.vds.filter_intervals(
+            exomes_vds, ht, split_reference_blocks=True
```
jkgoodrich, line 134: Is this needed? Did the Hail team say this is faster than just filtering the variant matrix table and doing a densify like we do in this script? https://github.com/broadinstitute/gnomad_qc/blob/main/gnomad_qc/v4/sample_qc/generate_qc_mt.py

gnomad_qc/v4/assessment/logistic_regression.py
```diff
@@ -0,0 +1,179 @@
...
+    Union exomes and genomes MatrixTables.
+
+    :param exomes_vds: VDS of exomes
+    :param genomes_vds: VDS of genomes
+    :return: Unioned MatrixTable
+    """
+    logger.info("Densifying exomes...")
+    exomes_mt = hl.vds.to_dense_mt(exomes_vds)
+    exomes_mt = exomes_mt.annotate_cols(is_genome=False)
+    exomes_mt = exomes_mt.select_entries("GT").select_rows().select_cols("is_genome")
```
jkgoodrich, line 94: If this is the only entry you need then you should filter to it before the densify. I think for this you probably also want to filter to only adj genotypes though, so you probably need more

gnomad_qc/v4/assessment/logistic_regression.py
```diff
@@ -0,0 +1,179 @@
...
+            ht.locus.contig, ht.locus.position, ht.locus.position + 1
+        )
+    )
+    ht = ht.key_by(ht.interval)
+    ht = ht.select()
+
+    return ht
+
+
+def densify_union_exomes_genomes(
```
jkgoodrich, line 80: I would split this up. I would run the exomes and genomes filter and densify in parallel, checkpoint each, and then union after those are done and checkpoint

gnomad_qc/v4/assessment/logistic_regression.py
KoalaQin: I didn't talk to Hail team, I tried filter_variants, I found it took very long to finish that step, then I remembered that once the vds has this [store_max_ref_length](https://app.zenhub.com/workspaces/gnomad-5f4d127ea61afc001d6be50b/issues/gh/broadinstitute/gnomad_production/1213), filter_intervals was much faster.

gnomad_qc/v4/assessment/logistic_regression.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def get_test_intervals(ht: hl.Table) -> hl.Table:
+    """
+    Get the test lists for the exomes and genomes MatrixTables.
+
+    :param joint_ht: Joint HT of v4 exomes and genomes.
+    :return: Test Table
+    """
+    # Filter to chr22
```
jkgoodrich, line 33: Before running the chr22 test, make an actual test that is only the first few partitions of chr22

gnomad_qc/v4/assessment/logistic_regression.py
```diff
@@ -0,0 +1,179 @@
...
+
+    pc_scores = ancestry_pca_scores().ht()
+    mt = mt.annotate_cols(pc=pc_scores[mt.col_key].scores)
+
+    # Perform logistic regression
+    ht = hl.logistic_regression_rows(
+        "firth",
+        y=mt.is_genome,
+        x=mt.GT.n_alt_alleles(),
+        covariates=[1] + [mt.pc[i] for i in range(10)],
```
jkgoodrich, line 153: I don't remember how many PCs we used for ancestry assignment off the top of my head, but I would use that number

gnomad_qc/v4/assessment/logistic_regression.py
KoalaQin: I agree, adj seems to make more sense. I will get that.

gnomad_qc/v4/assessment/logistic_regression.py
```diff
@@ -0,0 +1,179 @@
...
+    genomes_mt = hl.vds.to_dense_mt(genomes_vds)
+    genomes_mt = genomes_mt.annotate_cols(is_genome=True)
+    genomes_mt = genomes_mt.select_entries("GT").select_rows().select_cols("is_genome")
+
+    logger.info("Unioning exomes and genomes...")
+    mt = exomes_mt.union_cols(genomes_mt)
+    return mt
+
+
+def main(args):
```
jkgoodrich, line 106: add a try catch for copying the logger in case it fails with an error

gnomad_qc/v4/assessment/logistic_regression.py
KoalaQin: Mike told me it was 10, but I do remember you're exploring until 18,19, I will double check the code.

gnomad_qc/v4/assessment/logistic_regression.py
jkgoodrich: OK, sounds good, if it's faster then go for it

gnomad_qc/v4/assessment/logistic_regression.py
mike-w-wilson: Sorry if I wasnt clear, I wasnt certain it was 10, just thought it may be. Its 20: https://app.zenhub.com/workspaces/gnomad-5f4d127ea61afc001d6be50b/issues/gh/broadinstitute/gnomad_production/496

gnomad_qc/v4/assessment/logistic_regression.py
KoalaQin: I found the entries are not the same as what we used in generate freq for exomes or genomes, could you check the new steps for me?

gnomad_qc/v4/assessment/logistic_regression.py
KoalaQin: Only a few partitions were slower when I tested, when I get the set of intervals on chr22, they are more partitioned and they were densified faster.

gnomad_qc/v4/assessment/logistic_regression.py
KoalaQin: I added this.

gnomad_qc/v4/assessment/logistic_regression.py
KoalaQin: No worries, I put that 10 temporarily because it was just 10 in Julia's code. I changed it in my test.

gnomad_qc/v4/assessment/logistic_regression.py
```diff
@@ -0,0 +1,246 @@
...
+    vmt = vmt.select_entries("AD", "DP", "GT", "GQ")
+    vmt = vmt.annotate_cols(
+        is_genome=True if data_type == "genomes" else False,
+    )
+    vds = hl.vds.VariantDataset(rmt, vmt)
+    vds = vds.checkpoint(
+        hl.utils.new_temp_file(f"temp_{data_type}_vds_filtered", "vds")
+    )
+    mt = hl.vds.to_dense_mt(vds)
+    mt = annotate_adj(mt)
```
KoalaQin, line 121: After looking at your code, I think I could use `filter_to_adj` directly.

PR #619, opened 2024-05-31
Fixes to per sample summary stats

jkgoodrich:
- Adds adj filtering
- Changes to use frequency instead of recomputing allele counts
- Change VDS loading functions to support variant filtering prior to splitting the VDS
- Split out the filtering group HT generation into it's own step so it doesn't need to be done repeatedly

### DIFF
--- a/gnomad_qc/v3/resources/basics.py
+++ b/gnomad_qc/v3/resources/basics.py
@@ -92,7 +92,7 @@ def get_gnomad_v3_vds(
             )
 
     if split:
-        vds = v4_basics._split_and_filter_vds_for_loading(
+        vds = v4_basics._split_and_filter_variant_data_for_loading(
             vds, filter_variant_ht, entries_to_keep
         )
 
--- a/gnomad_qc/v3/resources/basics.py
+++ b/gnomad_qc/v3/resources/basics.py
@@ -28,6 +28,8 @@ def get_gnomad_v3_vds(
     filter_partitions: Optional[List[int]] = None,
     chrom: Optional[Union[str, List[str], Set[str]]] = None,
     filter_variant_ht: Optional[hl.Table] = None,
+    filter_intervals: Optional[List[str]] = None,
+    split_reference_blocks: bool = True,
     entries_to_keep: Optional[List[str]] = None,
     annotate_het_non_ref: bool = False,
 ) -> hl.vds.VariantDataset:
@@ -47,6 +49,9 @@ def get_gnomad_v3_vds(
     :param chrom: Optional argument to filter the VDS to specific chromosomes.
     :param filter_variant_ht: Optional argument to filter the VDS to a specific set of
         variants. Only supported when splitting the VDS.
+    :param filter_intervals: Optional argument to filter the VDS to specific intervals.
+    :param split_reference_blocks: Whether to split the reference data at the edges of
+        the intervals defined by `filter_intervals`. Default is True.
     :param entries_to_keep: Optional argument to keep only specific entries in the
         returned VDS. If splitting the VDS, use the global entries (e.g. 'GT') instead
         of the local entries (e.g. 'LGT') to keep.
@@ -71,6 +76,16 @@ def get_gnomad_v3_vds(
         vds = hl.vds.VariantDataset(
             vds.reference_data._filter_partitions(filter_partitions),
             vds.variant_data._filter_partitions(filter_partitions),
+        )
+
+    if filter_intervals:
+        logger.info("Filtering to %s intervals...", len(filter_intervals))
+        filter_intervals = [
+            hl.parse_locus_interval(x, reference_genome="GRCh38")
+            for x in filter_intervals
+        ]
+        vds = hl.vds.filter_intervals(
+            vds, filter_intervals, split_reference_blocks=split_reference_blocks
         )
 
     if remove_hard_filtered_samples:
--- a/gnomad_qc/v3/resources/basics.py
+++ b/gnomad_qc/v3/resources/basics.py
@@ -50,8 +50,8 @@ def get_gnomad_v3_vds(
     :param entries_to_keep: Optional argument to keep only specific entries in the
         returned VDS. If splitting the VDS, use the global entries (e.g. 'GT') instead
         of the local entries (e.g. 'LGT') to keep.
-    :param annotate_het_non_ref: Whether to annotate non_ref hets to unsplit variant
-        data. Default is False.
+    :param annotate_het_non_ref: Whether to annotate non reference heterozygotes (as
+        '_het_non_ref') to the variant data. Default is False.
     :return: gnomAD v3 dataset with chosen annotations and filters.
     """
     if test:
--- a/gnomad_qc/v3/resources/basics.py
+++ b/gnomad_qc/v3/resources/basics.py
@@ -29,6 +29,7 @@ def get_gnomad_v3_vds(
     chrom: Optional[Union[str, List[str], Set[str]]] = None,
     filter_variant_ht: Optional[hl.Table] = None,
     entries_to_keep: Optional[List[str]] = None,
+    annotate_het_non_ref: bool = False,
 ) -> hl.vds.VariantDataset:
     """
     Get gnomAD VariantDataset with desired filtering and metadata annotations.
@@ -49,6 +50,8 @@ def get_gnomad_v3_vds(
     :param entries_to_keep: Optional argument to keep only specific entries in the
         returned VDS. If splitting the VDS, use the global entries (e.g. 'GT') instead
         of the local entries (e.g. 'LGT') to keep.
+    :param annotate_het_non_ref: Whether to annotate non_ref hets to unsplit variant
+        data. Default is False.
     :return: gnomAD v3 dataset with chosen annotations and filters.
     """
     if test:
@@ -92,6 +95,13 @@ def get_gnomad_v3_vds(
             )
 
     vd = vds.variant_data
+    if annotate_het_non_ref:
+        logger.info("Annotating non_ref hets to unsplit variant data...")
+        vd = vd.annotate_entries(_het_non_ref=vd.LGT.is_het_non_ref())
+        entries_to_keep = (
+            None if entries_to_keep is None else entries_to_keep + ["_het_non_ref"]
+        )
+
     if split:
         vd = v4_basics._split_and_filter_variant_data_for_loading(
             vd, filter_variant_ht, entries_to_keep
--- a/gnomad_qc/v3/resources/basics.py
+++ b/gnomad_qc/v3/resources/basics.py
@@ -91,16 +91,16 @@ def get_gnomad_v3_vds(
                 vds.reference_data, vd.annotate_cols(meta=meta_ht[vd.col_key])
             )
 
+    vd = vds.variant_data
     if split:
-        vds = v4_basics._split_and_filter_variant_data_for_loading(
-            vds, filter_variant_ht, entries_to_keep
+        vd = v4_basics._split_and_filter_variant_data_for_loading(
+            vd, filter_variant_ht, entries_to_keep
         )
 
     if entries_to_keep is not None:
-        vds = hl.vds.VariantDataset(
-            vds.reference_data,
-            vds.variant_data.select_entries(*entries_to_keep),
-        )
+        vd = vd.select_entries(*entries_to_keep)
+
+    vds = hl.vds.VariantDataset(vds.reference_data, vd)
 
     return vds
 
--- a/gnomad_qc/v3/resources/basics.py
+++ b/gnomad_qc/v3/resources/basics.py
@@ -10,6 +10,7 @@ from gnomad.resources.resource_utils import (
     VersionedVariantDatasetResource,
 )
 
+import gnomad_qc.v4.resources.basics as v4_basics
 from gnomad_qc.v3.resources.constants import CURRENT_RELEASE, CURRENT_VERSION
 from gnomad_qc.v3.resources.meta import meta
 from gnomad_qc.v3.resources.sample_qc import hard_filtered_samples
@@ -26,6 +27,8 @@ def get_gnomad_v3_vds(
     test: bool = False,
     filter_partitions: Optional[List[int]] = None,
     chrom: Optional[Union[str, List[str], Set[str]]] = None,
+    filter_variant_ht: Optional[hl.Table] = None,
+    entries_to_keep: Optional[List[str]] = None,
 ) -> hl.vds.VariantDataset:
     """
     Get gnomAD VariantDataset with desired filtering and metadata annotations.
@@ -41,6 +44,11 @@ def get_gnomad_v3_vds(
     :param filter_partitions: Optional argument to filter the VDS to specific partitions
         in the provided list.
     :param chrom: Optional argument to filter the VDS to specific chromosomes.
+    :param filter_variant_ht: Optional argument to filter the VDS to a specific set of
+        variants. Only supported when splitting the VDS.
+    :param entries_to_keep: Optional argument to keep only specific entries in the
+        returned VDS. If splitting the VDS, use the global entries (e.g. 'GT') instead
+        of the local entries (e.g. 'LGT') to keep.
     :return: gnomAD v3 dataset with chosen annotations and filters.
     """
     if test:
@@ -84,17 +92,14 @@ def get_gnomad_v3_vds(
             )
 
     if split:
-        vd = vds.variant_data
-        vd = vd.annotate_rows(
-            n_unsplit_alleles=hl.len(vd.alleles),
-            mixed_site=(hl.len(vd.alleles) > 2)
-            & hl.any(lambda a: hl.is_indel(vd.alleles[0], a), vd.alleles[1:])
-            & hl.any(lambda a: hl.is_snp(vd.alleles[0], a), vd.alleles[1:]),
+        vds = v4_basics._split_and_filter_vds_for_loading(
+            vds, filter_variant_ht, entries_to_keep
         )
-        vd = hl.experimental.sparse_split_multi(vd, filter_changed_loci=True)
+
+    if entries_to_keep is not None:
         vds = hl.vds.VariantDataset(
             vds.reference_data,
-            vd,
+            vds.variant_data.select_entries(*entries_to_keep),
         )
 
     return vds
--- a/gnomad_qc/v4/assessment/calculate_per_sample_stats.py
+++ b/gnomad_qc/v4/assessment/calculate_per_sample_stats.py
@@ -445,10 +445,15 @@ def main(args):
 
     data_type = args.data_type
     test_dataset = args.test_dataset
-    test_n_partitions = (
+    test_partitions = (
         list(range(args.test_n_partitions)) if args.test_n_partitions else None
     )
-    test = args.test_n_partitions or args.test_dataset
+    test_partitions = (
+        [20180, 41229, 46085, 40916]
+        if args.test_difficult_partitions and data_type == "exomes"
+        else test_partitions
+    )
+    test = test_partitions or test_dataset
     overwrite = args.overwrite
     ukb_capture_intervals = (
         False if data_type == "genomes" else not args.skip_filter_ukb_capture_intervals
@@ -478,9 +483,10 @@ def main(args):
 
     try:
         if args.create_filter_group_ht:
+            logger.info("Creating Table of filter groups for summary stats...")
             release_ht = release_sites(data_type=data_type).ht()
-            if test_n_partitions:
-                release_ht = release_ht._filter_partitions(test_n_partitions)
+            if args.test_n_partitions:
+                release_ht = release_ht._filter_partitions(test_partitions)
 
             get_summary_stats_filter_groups_ht(
                 release_ht,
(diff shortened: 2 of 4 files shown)


gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -449,28 +479,16 @@ def main(args):
...
-                broad_capture_intervals = False
-
+        if args.create_filter_group_ht:
+            logger.info("Creating Table of filter groups for summary stats...")
             release_ht = release_sites(data_type=data_type).ht()
-            if test:
-                release_ht = hl.filter_intervals(
-                    release_ht, [hl.parse_locus_interval("chr22")]
-                )
+            filtering_groups_res = get_summary_stats_filtering_groups(
```
KoalaQin, line 485: See my comment in assessment.py, will this Table be overwritten by per_sample_res()?

KoalaQin (changes requested):
I'm sending you back my first round of review, I will take a closer look next week.

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -324,15 +337,17 @@ def create_per_sample_counts_ht(
     :param filter_group_ht: Table containing filter groups for summary stats.
     :return: Table containing per-sample variant counts.
     """
-    # Add extra Allele Count and Allele Type annotations to variant MatrixTable,
-    # according to Hail standards, to help their computation.
-    variant_ac, variant_types = vmt_sample_qc_variant_annotations(
+    # Add Allele Type annotations to variant MatrixTable.
+    _, variant_types = vmt_sample_qc_variant_annotations(
```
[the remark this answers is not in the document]
jkgoodrich: yeah, I thought about that too, but was hoping it happens small enough times per sample that we will still get a good per sample variant count @ch-kr and @mike-w-wilson do you think it's worth the effort to add this in now vs adding a note that it wasn't done, and will have a small impact on per sample numbers?

gnomad_qc/v4/resources/basics.py
```diff
@@ -353,6 +371,55 @@ def get_gnomad_v4_genomes_vds(
     return vds
 
 
+def _split_and_filter_vds_for_loading(
```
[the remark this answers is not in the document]
jkgoodrich: I don't need to, I just felt like it's not really a resource, but rather just useful within the loading functions, I really don't care either way on it, so let me know what you prefer

gnomad_qc/v4/resources/basics.py
```diff
@@ -353,6 +371,55 @@ def get_gnomad_v4_genomes_vds(
...
+        ]
+        vmt = vmt.select_entries(*split_entries_to_keep)
+
+    if filter_variant_ht is not None:
+        filter_locus_ht = (
+            filter_variant_ht.select()
+            .key_by("locus")
+            .checkpoint(hl.utils.new_temp_file("vmt_locus_filter", ".ht"))
+        )
+        vmt = vmt.filter_rows(hl.is_defined(filter_locus_ht[vmt.locus]))
```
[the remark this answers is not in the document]
jkgoodrich: nothing, but this is the way we typically do it if we already have a HT with all the loci rather than making an interval for each variant

gnomad_qc/v4/resources/basics.py
```diff
@@ -353,6 +371,55 @@ def get_gnomad_v4_genomes_vds(
...
+
+    logger.info("Splitting multiallelics...")
+    vmt = vmt.annotate_rows(
+        n_unsplit_alleles=hl.len(vmt.alleles),
```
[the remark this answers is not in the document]
jkgoodrich: We use them in other places, I just moved all of the split stuff out of the loading functions

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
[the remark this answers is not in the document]
jkgoodrich: For frequency this is how we do it, but we very much shouldn't densify for these stats, so it would have to be some sort of modification. 

Also, how to handle sex chromosomes...

This just gets more and more complicated!

```
def densify_and_prep_vds_for_freq(
    vds: hl.vds.VariantDataset,
    ab_cutoff: float = 0.9,
) -> hl.MatrixTable:
    """
    Densify VDS and select necessary annotations for frequency and histogram calculations.

    Select entry annotations required for downstream work. 'DP', 'GQ', and '_het_ab' are
    required for histograms. 'GT' and 'adj' are required for frequency calculations.
    '_high_ab_het_ref' is required for high AB call corrections in frequency and
    histogram annotations.

    Assumes all necessary annotations are present:
        - adj
        - _het_non_ref
        - GT
        - GQ
        - AD
        - DP
        - sex_karyotype

    :param vds: Input VDS.
    :param ab_cutoff: Allele balance cutoff to use for high AB het annotation.
    :return: Dense MatrixTable with only necessary entry annotations.
    """
    logger.info("Densifying VDS...")
    mt = hl.vds.to_dense_mt(vds)

    logger.info("Computing sex adjusted genotypes...")
    ab_expr = mt.AD[1] / mt.DP
    gt_expr = adjusted_sex_ploidy_expr(mt.locus, mt.GT, mt.sex_karyotype)

    mt = mt.select_entries(
        "DP",
        "GQ",
        GT=gt_expr,
        adj=get_adj_expr(gt_expr, mt.GQ, mt.DP, mt.AD),
        _het_ab=ab_expr,
        _high_ab_het_ref=(ab_expr > ab_cutoff) & ~mt._het_non_ref,
    )
    return mt
```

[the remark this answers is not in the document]
ch-kr: I vote we just add a note -- I think the effort to payoff ratio is unbalanced for rerunning after the fixing het nonref issue. what do you mean about the sex chromosomes?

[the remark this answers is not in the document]
jkgoodrich: Just that as is, we are not adjusting for ploidy either, so I probably need to add that in also.
I switched to have an option for only autosomes when I was still trying to get some numbers for Kaitlin, but I should probably add support for sex chromosomes. It's on this thread because I thought about it when I was looking over the code for the high allele balance het issue.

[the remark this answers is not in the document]
mike-w-wilson: Sorry I missed this Friday. Only 198772 sites (0.06% of all sites) have any high AB hets so I agree with a note.

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -322,17 +336,35 @@ def create_per_sample_counts_ht(
...
+        mt = mt.annotate_entries(
+            GT=hl.if_else(
+                (mt.variant_af > 0.01)
+                & ((mt.AD[1] / mt.DP) > ab_cutoff)
+                & mt.GT.is_het()
+                & ~mt._het_non_ref,
+                hl.call(1, 1),
+                mt.GT,
+            )
+        )
```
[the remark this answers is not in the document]
KoalaQin: Or if we all agree on adding a note instead of fixing this, maybe just remove this part?

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -322,17 +336,35 @@ def create_per_sample_counts_ht(
...
 
     # Annotate the MT with the needed annotations.
     mt = annotate_with_ht(mt, filter_group_ht, filter_missing=True)
```
KoalaQin, line 363: If we're still fixing high ab het, this has to be before the autosome_only argument, because you won't have `af`.

KoalaQin (changes requested):
I want to send it back for now, before I ran all the tests, since we agreed on not fixing high ab het.

gnomad_qc/v3/resources/basics.py
```diff
@@ -83,20 +94,24 @@ def get_gnomad_v3_vds(
...
 
-    if split:
-        vd = vds.variant_data
-        vd = vd.annotate_rows(
-            n_unsplit_alleles=hl.len(vd.alleles),
-            mixed_site=(hl.len(vd.alleles) > 2)
-            & hl.any(lambda a: hl.is_indel(vd.alleles[0], a), vd.alleles[1:])
-            & hl.any(lambda a: hl.is_snp(vd.alleles[0], a), vd.alleles[1:]),
+    vd = vds.variant_data
+    if annotate_het_non_ref:
```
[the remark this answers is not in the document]
jkgoodrich: It's completely fine to have both of those as True, you just need to make sure it is annotated before the split

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -322,17 +336,35 @@ def create_per_sample_counts_ht(
...
     :param filter_group_ht: Table containing filter groups for summary stats.
+    :param autosomes_only: Whether to restrict analysis to autosomes only. Default is
+        False.
     :return: Table containing per-sample variant counts.
     """
-    # Add extra Allele Count and Allele Type annotations to variant MatrixTable,
-    # according to Hail standards, to help their computation.
-    variant_ac, variant_types = vmt_sample_qc_variant_annotations(
+    # Add Allele Type annotations to variant MatrixTable.
+    _, variant_types = vmt_sample_qc_variant_annotations(
```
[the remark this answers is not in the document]
jkgoodrich: _ is commonly used to indicate that the returned variant isn't used

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -322,17 +336,35 @@ def create_per_sample_counts_ht(
...
+                & ~mt._het_non_ref,
+                hl.call(1, 1),
+                mt.GT,
+            )
+        )
 
     # Annotate the MT with the needed annotations.
     mt = annotate_with_ht(mt, filter_group_ht, filter_missing=True)
+    mt = mt.filter_entries(mt.GT.is_non_ref())
+    mt = filter_to_adj(mt)
```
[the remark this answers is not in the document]
jkgoodrich: I'm doing it because we only want adj variants

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -449,28 +510,18 @@ def main(args):
...
-                release_ht = hl.filter_intervals(
-                    release_ht, [hl.parse_locus_interval("chr22")]
-                )
+            filtering_groups_res = get_summary_stats_filtering_groups(
+                data_type=data_type, test=test, autosomes_only=autosomes_only
+            )
+            if args.test_n_partitions:
+                release_ht = release_ht._filter_partitions(test_partitions)
+            if autosomes_only:
+                release_ht = filter_to_autosomes(release_ht)
```
KoalaQin, line 522: I think it's still good to filter release_ht for test dataset, so the the test would go faster: 
```suggestion
            if args.test_dataset:
                release_ht = hl.filter_intervals(
                    release_ht, [hl.parse_locus_interval("chr22")]
                )
```

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -426,21 +458,50 @@ def main(args):
...
+
+    # The following four exome partitions have given us trouble with out-of-memory
+    # errors in the past. We are testing on these partitions to ensure that the
+    # per-sample stats script can handle them before running the entire dataset.
+    if data_type == "genomes" and args.test_difficult_partitions:
+        raise ValueError(
+            "Difficult partitions for testing have only been chosen for exomes data."
+        )
+
+    test_partitions = (
```
KoalaQin, line 478: Does this test require a different cluster setup?

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -523,6 +603,16 @@ def main(args):
         choices=["exomes", "genomes"],
         help="Data type (exomes or genomes) to produce summary stats for.",
     )
+    parser.add_argument(
+        "--autosomes-only",
+        help="Whether to restrict analysis to autosomes only.",
```
KoalaQin, line 608: This argument only applies to the part you want to change 
the high ab het non-ref on autosomes, right? I'm not sure how it would be useful to filter release_ht.

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -513,7 +583,17 @@ def main(args):
...
+    )
+    parser.add_argument(
+        "--test-difficult-partitions",
+        help="Whether to test on a set of 4 difficult exome partitions.",
+        action="store_true",
+    )
+    parser.add_argument(
+        "--test-dataset",
         help="Test on a small number of variants",
         action="store_true",
```
KoalaQin, line 598: I think we have to specify this: VDS will be using VDS of a small number of samples, and release_ht only chr22.

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -479,26 +530,45 @@ def main(args):
...
+                data_type, autosomes_only=autosomes_only
+            ).ht()
+            vds_load_func = (
+                get_gnomad_v4_vds
+                if data_type == "exomes"
+                else get_gnomad_v4_genomes_vds
+            )
+            mt = vds_load_func(
+                test=test_dataset,
+                split=True,
```
KoalaQin, line 549: Wouldn't here be: `test=test`?

KoalaQin (changes requested):
A few comments and questions.

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
jkgoodrich: It requires the changes coming in the next PR

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
jkgoodrich: I want to be able to test the full test dataset, for the faster test, use `--test-dataset` and `--test-n-partitions n`

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
jkgoodrich: I only want to test on the test dataset if `--test-dataset` is passed

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -479,26 +529,49 @@ def main(args):
...
+            ).write(filtering_groups_res.path, overwrite=overwrite)
+
+        if args.create_per_sample_counts_ht:
+            logger.info(
+                "Calculating per-sample variant statistics for %s...", data_type
             )
-            create_per_sample_counts_ht(mt, filter_groups_ht).write(
+            if test:
+                logger.warning(
+                    "This test requires that the full filtering groups Table has been "
```
KoalaQin, line 540: If it's a full filtering groups Table, then in assessment and above output path, do you still need to put "test"?

KoalaQin: Because if I want to run the next steps, with the written table, I don't have variant_ac in it.

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -449,28 +511,16 @@ def main(args):
...
-                broad_capture_intervals = False
-
+        if args.create_filter_group_ht:
+            logger.info("Creating Table of filter groups for summary stats...")
             release_ht = release_sites(data_type=data_type).ht()
-            if test:
-                release_ht = hl.filter_intervals(
-                    release_ht, [hl.parse_locus_interval("chr22")]
-                )
+            filtering_groups_res = get_summary_stats_filtering_groups(
```
KoalaQin, line 517: I think it still makes more sense to get test filtering group HT too, by filtering to test-dataset and test-partitions, because the full filtering group Table will take very long to run, especially you're adding new annotations `variant_ac`, `variant_af`, it will be cost efficient to test and see if that's all we want.

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -449,28 +511,16 @@ def main(args):
...
-            else:
-                logger.info("Calculating per-sample variant statistics for genomes...")
-                mt = get_gnomad_v4_genomes_vds(
-                    test=test, release_only=True, split=True, chrom=chrom
-                ).variant_data
-                ukb_capture_intervals = False
-                broad_capture_intervals = False
-
+        if args.create_filter_group_ht:
+            logger.info("Creating Table of filter groups for summary stats...")
```
KoalaQin, line 515: How about this for this chunk? 
```commandline
            logger.info("Creating Table of filter groups for %s for summary stats...", data_type)
            release_ht = release_sites(data_type=data_type).ht()
            filtering_groups_res = get_summary_stats_filtering_groups(
                data_type=data_type, test=test
            )
            if args.test_n_partitions:
                release_ht = release_ht._filter_partitions(test_partitions)
            if args.test_dataset:
                release_ht = hl.filter_intervals(
                    release_ht, [hl.parse_locus_interval("chr22")]
                )

            get_summary_stats_filter_groups_ht(
                release_ht,
                pass_filters=not args.skip_pass_filters,
                ukb_capture=ukb_capture_intervals,
                broad_capture=broad_capture_intervals,
                by_csqs=not args.skip_by_csqs,
                vep_canonical=args.vep_canonical,
                vep_mane=args.vep_mane,
                rare_variants_afs=rare_variants_afs,
            ).write(filtering_groups_res.path, overwrite=overwrite)
```

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -479,26 +529,49 @@ def main(args):
...
             )
-            create_per_sample_counts_ht(mt, filter_groups_ht).write(
+            if test:
+                logger.warning(
+                    "This test requires that the full filtering groups Table has been "
+                    "created. Please run --create-filter-group-ht without "
+                    "--test-dataset or --test-n-partitions if it doesn't already exist."
+                )
+            filter_groups_ht = get_summary_stats_filtering_groups(data_type).ht()
+            vds_load_func = (
```
KoalaQin, line 545:
```suggestion
            filter_groups_ht = get_summary_stats_filtering_groups(data_type, test=test).ht()

```

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -479,26 +529,49 @@ def main(args):
...
+            logger.info(
+                "Calculating per-sample variant statistics for %s...", data_type
             )
-            create_per_sample_counts_ht(mt, filter_groups_ht).write(
+            if test:
+                logger.warning(
+                    "This test requires that the full filtering groups Table has been "
+                    "created. Please run --create-filter-group-ht without "
+                    "--test-dataset or --test-n-partitions if it doesn't already exist."
+                )
```
KoalaQin, line 543: This needs to change if you agree with my suggestions.

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -322,17 +336,36 @@ def create_per_sample_counts_ht(
...
+    _, variant_types = vmt_sample_qc_variant_annotations(
         global_gt=mt.GT, alleles=mt.alleles
     )
-    mt = mt.annotate_rows(variant_ac=variant_ac, variant_atypes=variant_types)
+    mt = mt.annotate_rows(variant_atypes=variant_types)
 
     # Annotate the MT with the needed annotations.
     mt = annotate_with_ht(mt, filter_group_ht, filter_missing=True)
+    if autosomes_only:
+        ab_cutoff = 0.9
```
KoalaQin, line 352: Add a logger here?

KoalaQin (changes requested):
Hopefully this is the last round of this PR, I still want the test on filtering group HT creation, so this will be feasible for a test on both exomes and genomes.

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -479,26 +529,49 @@ def main(args):
...
+                    "--test-dataset or --test-n-partitions if it doesn't already exist."
+                )
+            filter_groups_ht = get_summary_stats_filtering_groups(data_type).ht()
+            vds_load_func = (
+                get_gnomad_v4_vds
+                if data_type == "exomes"
+                else get_gnomad_v4_genomes_vds
+            )
+            mt = vds_load_func(
+                test=test_dataset,
```
KoalaQin, line 551: Probably can't be solved in this PR, but genomes test VDS has the same number of partitions as the whole VDS, 115376 for (31236097, 80)  variants x samples, this over-partition makes the test run super slow.

jkgoodrich (commented):
It's clear from all of the comments around the testing of the script that you don't understand how I'm using each of the tests that I have added, and when each should be used. I have attempted to clarify when each of these tests is actually useful.

Additionally, once the `--create-filter-group-ht` has been tested, it's actually fairly quick and costs very very little to make the full filter group HT, and it can then just be used as is for the memory tests. I'm sorry that the current table doesn't have the values you were looking for, those values are the values that it will be changed to for the memory test in the PR that I plan to submit when we can get this one in. However, in the mean time feel free to update the HT, its very simple and since we haven't recorded the final run, it's not a problem at all to change. 

Since you seem concerned though about the way I am testing, I have also added a test for PCSK9 that can be used with the test dataset to have a very fast run of the full pipeline in the cases where we plan to make many changes and runs of a test with both --create-filter-group-ht and --create-per-sample-counts-ht.

If you think that the tests are now documented well enough for someone to come back and understand under what situations you would want to use each test, then I suggest we get this PR in so that we can switch our focus to a PR with changes that can get around the many memory issues the current code has.

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
jkgoodrich: Yes, as discussed previously, we should only be focusing on the exomes right now, we can handle these finer details when we get to working with the genomes.

KoalaQin (approved):
Thank you so much for adding in all the changes, and it's very clear now, my intention was to get a successful quick test for the full pipeline, without rewriting big files. I have a small comment but LGTM!

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -479,23 +596,56 @@ def main(args):
...
+                split=True,
+                release_only=True,
+                filter_partitions=test_partitions,
+                filter_variant_ht=filter_groups_ht,
```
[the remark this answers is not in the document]
jkgoodrich: Thanks for pointing this out, I added a PR to fix `annotate_with_ht` in gnomad_methods, and removed the `filter_missing=True` from the call to it here

PR #621, opened 2024-06-06
Small changes in agg function to make it work for per-sample stats on genomes

KoalaQin:
Add data_type parameter so it will work on genomes. 
Test: [51fc7def2cee490f84bf060ed1018fcb](https://console.cloud.google.com/dataproc/jobs/51fc7def2cee490f84bf060ed1018fcb?region=us-central1&project=broad-mpg-gnomad)

### DIFF
--- a/gnomad_qc/v4/assessment/calculate_per_sample_stats.py
+++ b/gnomad_qc/v4/assessment/calculate_per_sample_stats.py
@@ -404,7 +404,6 @@ def compute_agg_sample_stats(
     meta_ht: Optional[hl.Table] = None,
     by_ancestry: bool = False,
     by_subset: bool = False,
-    data_type: str = "exomes",
 ) -> hl.Table:
     """
     Compute aggregate statistics for per-sample QC metrics.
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/assessment/calculate_per_sample_stats.py
```diff
@@ -414,6 +415,7 @@ def compute_agg_sample_stats(
     :param by_ancestry: Boolean indicating whether to stratify by ancestry.
     :param by_subset: Boolean indicating whether to stratify by subset. This is only
          working on "exomes" data.
+    :param data_type: Data type for sample QC metrics, e.g. "exomes" or "genomes".
```
jkgoodrich, line 418: I don't think we need to add data_type, we can just check if by_subset is True, and then in main, if args.data_type and args.by_subset are both set, raise an error

jkgoodrich (changes requested):

gnomad_qc/v4/assessment/calculate_per_sample_stats.py
KoalaQin: You're right, I did that.

jkgoodrich (approved):
LGTM

PR #630, opened 2024-07-30
Add script to compute pext

jkgoodrich:
Tested with v7

Command: 
```
hailctl dataproc submit jg1 gnomad_qc/analyses/compute_pext.py \
    --test \
    --gtex-version v7 \
    --vep-version 85 \
    --annotation-level \
    --base-level \
    --browser-ht \
    --overwrite \
    --pyfiles ../gnomad_methods/gnomad,gnomad_qc
```

Job ID: ee21aac09a784eaa81a955eedfe1374f

Not tested on v10 because the resource isn't able to be public yet

Relies on gnomad_methods [PR](https://github.com/broadinstitute/gnomad_methods/pull/721)


gnomad_qc/analyses/compute_pext.py
```diff
@@ -0,0 +1,279 @@
...
+    overwrite = args.overwrite
+
+    hl.init(
+        log="/pext.log",
+        default_reference="GRCh37" if gtex_version == "v7" else "GRCh38",
+        tmp_dir="gs://gnomad-tmp-4day",
+    )
+
+    pext_res = get_pipeline_resources(test, gtex_version, vep_version, overwrite)
+    ht = pext_res.context_ht.ht().naive_coalesce(5000)
```
ch-kr, line 179: will this ever need to change? (will this table always have >5k partitions?)

gnomad_qc/analyses/compute_pext.py
```diff
@@ -0,0 +1,279 @@
...
+        create_tx_annotation_by_region(base_pext_ht).write(
+            res.browser_pext_ht.path, overwrite=overwrite
+        )
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser(
+        description="This script creates the pext annotation files for release."
+    )
+    parser.add_argument("--overwrite", help="Overwrite data", action="store_true")
```
ch-kr, line 239:
```suggestion
    parser.add_argument("--overwrite", help="Overwrite data.", action="store_true")
```
nit for consistency with the other descriptions

gnomad_qc/analyses/compute_pext.py
```diff
@@ -0,0 +1,279 @@
...
+        "--gtex-version",
+        help="GTEx version to use for pext scores.",
+        default="v10",
+        choices=["v7", "v10"],
+    )
+    parser.add_argument(
+        "--vep-version",
+        help="VEP version to use for pext scores.",
+        default="105",
+        choices=["85", "95", "101", "105"],
```
ch-kr, line 257:
```suggestion
        choices=["85", "105"],
```
v7 has to be GENCODE v19 and v10 has to be v39, right?

gnomad_qc/analyses/resources.py
```diff
@@ -0,0 +1,56 @@
...
+    return (
+        f"gs://gnomad-tmp/gnomad_{name}_testing/{version}"
+        if test
+        else f"gs://gnomad/analyses/{name}/{version}"
+    )
+
+
+def get_pext(
+    name: str = "base_level",
+    gtex_version: str = "v7",
```
ch-kr, line 37:
```suggestion
    gtex_version: str = "v10",
```
changing for consistency with the vep version

gnomad_qc/analyses/resources.py
```diff
@@ -0,0 +1,56 @@
...
+    vep_version: str = "105",
+    test: bool = False,
+) -> TableResource:
+    """
+    Get pext annotation file.
+
+    :param name: Name of the pext annotation file. Default is "base_level".
```
ch-kr, line 44:
```suggestion
    :param name: Name of the pext annotation file. Default is "base_level". 
        Must be one of ["annotation_level", "base_level", "browser"].
```
though not sure this is helpful, since resource existence is checked elsewhere

gnomad_qc/analyses/resources.py
```diff
@@ -0,0 +1,56 @@
...
+    :param gtex_version: GTEx version. Default is "v7".
```
ch-kr, line 45:
```suggestion
    :param gtex_version: GTEx version. Default is "v10".
```
(if you change the default above)

ch-kr (changes requested):
a few minor requests. will get the gnomad_methods PR back to you soon!

gnomad_qc/analyses/compute_pext.py
jkgoodrich: This is just the context HT, I was struggling to use it with the large number of partitions, and this made it much better, happy to change if you think there is a better way to handle it, or add a comment about the number of partitions in the context HTs

gnomad_qc/analyses/compute_pext.py
```diff
@@ -0,0 +1,279 @@
...
+    Get pipeline resources for pext annotation pipeline.
+
+    :param test: Whether to get resources for a test run.
+    :param gtex_version: GTEx version to get resources for.
+    :param vep_version: VEP version to get resources for.
+    :param overwrite: Whether to overwrite existing data.
+    :return: Pipeline resource collection for pext annotation pipeline.
+    """
+    if gtex_version == "v7":
+        gnomad_res = gnomad_grch37
```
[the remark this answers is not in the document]
jkgoodrich: Actually, I will just change it overall to use the VEP that corresponds to the version rather than making vep an option

gnomad_qc/analyses/compute_pext.py
```diff
@@ -0,0 +1,279 @@
...
+    :param ht: VEP context HT to filter.
+    :param tx_mt: Transcript MT to filter.
+    :return: Filtered HT and MT.
+    """
+    test_intervals = [
+        hl.parse_locus_interval("19:45409011-45412650"),  # ENSG00000130203, APOE
+        hl.parse_locus_interval("13:47405685-47471169"),  # ENSG00000102468, HTR2A
+        hl.parse_locus_interval("1:910579-917497"),  # ENSG00000187642, C1orf170/PERM1
+        hl.parse_locus_interval("1:6521211-6526255"),  # ENSG00000215788, TNFRSF25
+        hl.parse_locus_interval("1:1138888-1142071"),  # ENSG00000186891, TNFRSF18
```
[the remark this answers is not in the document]
jkgoodrich: Yup, thanks for catching. Copy paste error from Qin's v2 script

gnomad_qc/analyses/compute_pext.py
```diff
@@ -0,0 +1,279 @@
...
+
+
+def filter_to_test(
+    ht: hl.Table,
+    tx_mt: hl.MatrixTable,
+) -> Tuple[hl.Table, hl.MatrixTable]:
+    """
+    Filter `ht` and `tx_mt` to a test set of genes.
+
+    Genes to filter to:
```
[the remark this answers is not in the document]
jkgoodrich: I'm not sure, I just copied Qin's gene list over for testing. To be honest, I didn't really use the tests, since I ran this in a notebook initially, and it doesn't take that long on the full dataset. If you have some genes that you think would be good to add for testing, I'm happy to add them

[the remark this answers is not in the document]
ch-kr: ah that's helpful to know, thanks!

I don't have any genes that I regularly check pext for, so no suggestions here. not high priority, but maybe we could ask someone on the analysis or curation teams for genes they regularly check for future reference?

gnomad_qc/analyses/compute_pext.py
ch-kr: I typically like to add an argument with a default (just in case the number of partitions you want to use isn't 5k in a rerun), but if that is too much for pext (which we shouldn't have to run often), just a comment sounds good to me

ch-kr (approved):
just one comment about partitions but LGTM

gnomad_qc/analyses/compute_pext.py
jkgoodrich: I can add a default (I usually do for requested output partitions), I was just thinking that the context HT will always be about the same size, and this partitioning was really just to make it not take forever, so might not need to change the number of partitions that make it run better

ch-kr: sounds good, maybe just adding a comment is the easiest then?

gnomad_qc/analyses/compute_pext.py
```diff
@@ -109,7 +137,7 @@ def get_pipeline_resources(
         pipeline_name="compute_pext",
         pipeline_resources={
             "context HT": {"context_ht": gnomad_res.vep_context.versions[vep_version]},
-            "GTEx RSEM MT": {"tx_mt": gnomad_res.gtex_rsem.versions[gtex_version]},
+            "GTEx RSEM MT": {"tx_mt": tx_mt},
```
ch-kr, line 140: maybe add a NOTE (or TODO) here that you can flip this line back to `"GTEx RSEM MT": {"tx_mt": gnomad_res.gtex_rsem.versions[gtex_version]},` once the public version is available? just since that is obvious now but may be forgotten later

gnomad_qc/analyses/compute_pext.py
```diff
@@ -118,35 +146,40 @@ def get_pipeline_resources(
...
+            "base_pext_ht": get_pext("base_level", gtex_version, test),
+        },
+    )
+    gnomad_exomes = PipelineStepResourceCollection(
+        "--gnomad-exomes-ht",
+        input_resources={
+            "gnomad exomes HT": {"gnomad_exomes_ht": gnomad_exomes},
+        },
+        output_resources={
+            "gnomad_exomes_pext_ht": get_pext("gnomad_exomes", gtex_version, test),
```
ch-kr, line 166: `get_pext` already returns `gnomad` in the name, so this will become `f"{_analysis_root('pext', version=version, test=test)}/gnomad.pext.gtex_v10.gnomad_exomes.ht"` -- maybe `name` here could just be `exomes`? and same comment below for genomes

ch-kr (commented):
a couple minor things

PR #631, opened 2024-08-12
Small modifications to v4 genomes loading to work in summary stats code

jkgoodrich:
These minor changes were made during runtime of the genomes summary stats


mike-w-wilson (approved):
Just want to update this call here to match the arg change: https://github.com/broadinstitute/gnomad_qc/blob/152c5568afd6ee77ebde04b7737ea2b08e2a7705/gnomad_qc/v4/annotations/compute_coverage.py#L430 
Then LGTM

PR #633, opened 2024-09-09
Add --min-samples and --keep-tissues-under-min-samples args to allow for some samples to also be fully excluded from the pext files

jkgoodrich:
Depends on gnomad_methods [PR](https://github.com/broadinstitute/gnomad_methods/pull/729)

### DIFF
--- a/gnomad_qc/analyses/compute_pext.py
+++ b/gnomad_qc/analyses/compute_pext.py
@@ -12,9 +12,9 @@ from gnomad.resources.grch38.gnomad import public_release as gnomad_release_grch
 from gnomad.utils.reference_genome import get_reference_genome
 from gnomad.utils.slack import slack_notifications
 from gnomad.utils.transcript_annotation import (
-    TISSUES_TO_EXCLUDE,
     clean_tissue_name_for_browser,
     create_tx_annotation_by_region,
+    get_tissues_to_exclude,
     perform_tx_annotation_pipeline,
     summarize_transcript_expression,
 )
@@ -206,6 +206,7 @@ def main(args):
     test = args.test
     gtex_version = args.gtex_version
     overwrite = args.overwrite
+    min_samples = args.min_samples
     annotation_level = ["annotation_level"] if args.annotation_level else []
     annotation_level += ["gnomad_exomes"] if args.gnomad_exomes_ht else []
     annotation_level += ["gnomad_genomes"] if args.gnomad_genomes_ht else []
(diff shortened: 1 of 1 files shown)


gnomad_qc/analyses/compute_pext.py
```diff
@@ -224,6 +225,20 @@ def main(args):
     ht = pext_res.context_ht.ht().naive_coalesce(5000)
     tx_mt = pext_res.tx_mt.mt()
 
+    # Any tissue below 50 samples will NOT get shown on website in overall pext or as
```
ch-kr, line 228: should 50 be 100 here based on what you sent?
> We observed that for low number of samples the variation in the number of expressed genes is
higher. The number of expressed genes reaches saturation for sample numbers over 100. This
suggests that for tissues with low number of available samples: kidney cortex, kidney medulla,
bladder, endocervix, ectocervix, and fallopian tube the analyses (correlation of gene expression
between tissues, number of expressed genes, number of expressed transcripts) may display high
variation.

https://bmcgenomics.biomedcentral.com/articles/10.1186/s12864-021-08125-9

gnomad_qc/analyses/compute_pext.py
```diff
@@ -224,6 +225,20 @@ def main(args):
...
+    # individual tissue.
+    # Exclude testes and cell lines from overall pext but keep in display as
+    # individual tissues (agreed to take a look at the per tissue pext values to make
+    # sure this makes sense).
```
ch-kr, line 232:
```suggestion
    # Exclude testes and cell lines from overall pext but keep in display as
    # individual tissues (checked per tissue pext values to make
    # sure this made sense).
```
changing tense since this was done, right?

gnomad_qc/analyses/compute_pext.py
```diff
@@ -224,6 +225,20 @@ def main(args):
...
+    tissues_to_exclude = get_tissues_to_exclude(
+        tx_mt, reproductive=False, cell_lines=False, min_samples=min_samples
```
ch-kr, line 234: should these be True now?

gnomad_qc/analyses/compute_pext.py
```diff
@@ -305,6 +322,24 @@ def main(args):
...
     )
+    parser.add_argument(
+        "--min-samples",
+        help=(
+            "Minimum number of samples to include a tissue in the overall mean pext. "
+            "If keep-tissues-under-min-samples is not set, these tissues will also be "
+            "excluded from the individual tissue pext values."
+        ),
+        type=int,
+        default=50,
```
ch-kr, line 333: 100?

ch-kr (commented):
a couple questions to double check

gnomad_qc/analyses/compute_pext.py
jkgoodrich: Yeah, it's 100 based on what I sent, but Konrad wanted to lower it to 50 to rescue Bladder in v10.

gnomad_qc/analyses/compute_pext.py
jkgoodrich: Sorry, this comment was supposed to get deleted. I copied what I wrote in Slack so that I knew what I was supposed to change things to. I will fix

gnomad_qc/analyses/compute_pext.py
jkgoodrich: No, I think we want these still displayed on the browser and in the downloaded file, we just don't want them in the mean. I will add comments

jkgoodrich: Cleaned up and added comments, let me know if it makes sense

gnomad_qc/analyses/compute_pext.py
jkgoodrich: Same comment about Bladder

ch-kr: does there need to be additional logic based on today's ATGU/DSP meeting since we decided to keep bladder in the downloads but not display on the browser (because n samples < 100)?

ch-kr (commented):
one more question about bladder

gnomad_qc/analyses/compute_pext.py
jkgoodrich: yup, just added

mike-w-wilson (approved):
Review only the final commit as requested:  :shipit:

PR #640, opened 2024-10-07
Add federated validity check

klaricch:


gnomad_qc/v5/data_ingestion/federated_validity_checks.py
```diff
@@ -0,0 +1,50 @@
...
+        ht = ht._filter_partitions(range(2))
+
+    expected_contigs = [
+        i
+        for i in hl.get_reference("GRCh38").contigs
+        if i in [f"chr{i}" for i in range(1, 23)] + ["chrX", "chrY"]
+    ]
+    logger.info("Summarizing variants and checking contigs.")
+    summarize_variants(ht, expected_contigs=expected_contigs)
+
```
matren395, line 38: odd note: "summarize_variants" is always going to throw Missing_Contigs if a test is passed in. Is that cool/expected behavior, or would we want to account for that there (and in error reporting) as well ?

matren395 (commented):
Some Qs about --test

gnomad_qc/v5/data_ingestion/federated_validity_checks.py
klaricch: yeah at the moment i prefer it to throw the error for the test. it's mostly to make sure all the functions are actually running right now

gnomad_qc/v5/data_ingestion/federated_validity_checks.py
```diff
@@ -0,0 +1,50 @@
...
+logging.basicConfig(format="%(levelname)s (%(name)s %(lineno)s): %(message)s")
+logger = logging.getLogger("federated_validity_checks")
+logger.setLevel(logging.INFO)
+
+
+def main(args):
+    """Perform validity checks for federated data."""
+    hl.init(
+        log="/federated_validity_checks.log",
```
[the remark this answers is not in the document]
klaricch: hmm, it does work and is the typical syntax we've used in the past
am also going to add a try/finally statement for copying the log

gnomad_qc/v5/data_ingestion/federated_validity_checks.py
```diff
@@ -0,0 +1,50 @@
...
+        ht = public_release(data_type="exomes").ht()
+        ht = ht._filter_partitions(range(2))
+
+    expected_contigs = [
+        i
+        for i in hl.get_reference("GRCh38").contigs
+        if i in [f"chr{i}" for i in range(1, 23)] + ["chrX", "chrY"]
+    ]
+    logger.info("Summarizing variants and checking contigs.")
+    summarize_variants(ht, expected_contigs=expected_contigs)
```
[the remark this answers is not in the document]
klaricch: Changed this to default load in the full gnomad ht and only filter to the first two partitions if test is passed. Eventually will need to change this but am not sure where the data will be stored or how the paths will be configured, so it's more of a placeholder for now

gnomad_qc/v5/data_ingestion/federated_validity_checks.py
```diff
@@ -0,0 +1,50 @@
...
+        log="/federated_validity_checks.log",
+        tmp_dir="gs://gnomad-tmp-4day",
+    )
+    hl.default_reference("GRCh38")
+    test = args.test
+
+    # TODO: Add resources to intake federated data once obtained.
+    if test:
+        ht = public_release(data_type="exomes").ht()
+        ht = ht._filter_partitions(range(2))
```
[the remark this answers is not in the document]
klaricch: right now it's just to make sure things run (will be editing some functions to work better with Table format rather than a flattened info field), and we don't have any federated data to use. In the future the test option may be useful for federated data just to make sure nothing fails from some formatting issue. I can change the test function to allow a variable number of partitions to make it more flexible.

gnomad_qc/v5/resources/basics.py
```diff
@@ -6,3 +6,26 @@
...
+    """
+    Return path to temporary QC bucket.
+
+    :param version: Version of annotation path to return.
+    :return: Path to bucket with temporary QC data.
+    """
+    return f"gs://gnomad-tmp/gnomad.genomes.v{version}.qc_data/"
+
+
+# TODO: Change default to CURRENT_VERSION.
```
matren395, line 22: why not add CURRENT_VERSION now ? or is that poor practice if there's not a version out

matren395 (commented):
one last comment - once the linting stuff is done, should be gtg

gnomad_qc/v5/resources/basics.py
klaricch: Just remembered that CURRENT_VERSION is already defined for v5 in another script, will update

matren395 (approved):
lgtm!

PR #644, opened 2024-11-21
Add argument to export pext tsv files

jkgoodrich:


gnomad_qc/analyses/compute_pext.py
```diff
@@ -338,6 +352,17 @@ def main(args):
...
+    if args.export_tsv:
+        logger.info("Exporting pext Hail Tables to TSVs...")
+        res = pext_res.export_tsv
+        res.check_resource_existence()
+
+        for ht in [res.base_pext_ht.ht(), res.annotation_level_pext_ht.ht()]:
+            ht = ht.annotate(
+                **{k: ht[k].expression_proportion for k in hl.eval(ht.tissues)}
+            )
+            ht.export(res.pext_base_tsv)
```
mike-w-wilson, line 364: Update this line to work for annotation tsv too

mike-w-wilson (changes requested):
One fix then LGTM

PR #645, opened 2024-12-02
Write a new final filter Table with all variants for de novo release

KoalaQin:
Our original final filter table filtered out variants that are not in the release by freq_ht, we want to have non-missing filters for de novo variants (which weren't in the release because some probands were excluded due to relatedness). 
Test: [45ff85cc7538413ba0ea020cdb96f3ad](https://console.cloud.google.com/dataproc/jobs/45ff85cc7538413ba0ea020cdb96f3ad?region=us-central1&inv=1&invt=AbjExw&project=broad-mpg-gnomad)

### DIFF
--- a/gnomad_qc/v4/resources/variant_qc.py
+++ b/gnomad_qc/v4/resources/variant_qc.py
@@ -1,6 +1,6 @@
 """Script containing variant QC related resources."""
 
-from typing import Union
+from typing import Optional, Union
 
 from gnomad.resources.grch38 import (
     na12878_giab,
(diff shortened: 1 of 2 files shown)


gnomad_qc/v4/resources/variant_qc.py
```diff
@@ -1,6 +1,6 @@
 """Script containing variant QC related resources."""
 
-from typing import Union
+from typing import Optional, Union
```
jkgoodrich, line 3: Optional isn't needed in this case, it's only needed when you want to support `None`
```suggestion
from typing import Union
```

gnomad_qc/v4/resources/variant_qc.py
```diff
@@ -296,20 +296,24 @@ def get_variant_qc_result(
 
 
 def final_filter(
-    data_type: str = "exomes", test: bool = False
+    data_type: str = "exomes", test: bool = False, simplified: Optional[bool] = False
```
jkgoodrich, line 299: Changing to make it a little more obvious than simplified 
```suggestion
    data_type: str = "exomes",
    all_variants: bool = False,
    only_filters: bool = False,
    test: bool = False,
```

gnomad_qc/v4/resources/variant_qc.py
```diff
@@ -296,20 +296,24 @@ def get_variant_qc_result(
...
 ) -> VersionedTableResource:
     """
     Get finalized variant QC filtering Table.
 
     :param data_type: Whether to return 'exomes' or 'genomes' data. Default is exomes.
     :param test: Whether to use a tmp path for variant QC tests.
+    :param simplified: Whether to get the simplified version of the Table. The simplified version
+                       includes variants not in the release, with only "filters" fields.
+                       Default is False.
```
jkgoodrich, line 308:
```suggestion
    :param all_variants: Whether to return the Table with all variants. Default is 
        False.
    :param only_filters: Whether to return the Table with only the 'filters' field. 
        Default is False.
    :param test: Whether to use a tmp path for variant QC tests. Default is False.
```

gnomad_qc/v4/resources/variant_qc.py
```diff
@@ -296,20 +296,24 @@ def get_variant_qc_result(
...
     :return: VersionedTableResource for final variant QC data.
     """
+    suffix = ".final_filter.simplified" if simplified else ".final_filter"
```
jkgoodrich, line 311:
```suggestion
    suffix = (
        f".final_filter{'.all_variants' if all_variants else ''}"
        f"{'.only_filters' if only_filters else ''}"
    )
```

jkgoodrich (changes requested):

PR #664, opened 2025-03-28
Update release to bool and aou qc metric file, release 3/27

mike-w-wilson:
A few updates after I realized release was a string and AoU released QC metric data for all files. The file change doesnt matter for this module but I figured better to grab the all sample file here rather than flagged samples. I also moved to a left join as the AoU SQL query was pulling age data for 600k participants but we only want srWGS sample data.

### DIFF
--- a/gnomad_qc/v5/sample_qc/merge_project_meta.py
+++ b/gnomad_qc/v5/sample_qc/merge_project_meta.py
@@ -17,7 +17,7 @@ pm.execute_notebook(  # noqa
     "utils_restart_kernel_with_gnomad_packages.ipynb",
     "utils_restart_notebook_output.ipynb",
     parameters={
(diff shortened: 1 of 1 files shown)


gnomad_qc/v5/sample_qc/merge_project_meta.py
```diff
@@ -17,7 +17,7 @@
     "utils_restart_kernel_with_gnomad_packages.ipynb",
     "utils_restart_notebook_output.ipynb",
     parameters={
-        "GNOMAD_QC_BRANCH": "main",
+        "GNOMAD_QC_BRANCH": "mw/v5_project_meta_update",
```
klaricch, line 20:
```suggestion
        "GNOMAD_QC_BRANCH": "main",
```

klaricch (approved):

PR #668, opened 2025-05-07
Update v3 sample_qc and variant_qc resources to pull from gs://gnomad-autoclass instead of gs://gnomad

jkgoodrich:


ch-kr (approved):
LGTM! only one small question, should we add a note that accessing data in archive is associated with additional fees?

PR #670, opened 2025-05-16
Remove loci with >100 alt alleles from vds sample QC

KoalaQin:
This PR makes two main changes to improve test coverage and performance for `compute_sample_qc`:
1. Excludes a locus with extreme allele count from loading to avoid memory or runtime issues.
2. Updates test mode logic in `compute_sample_qc` to load a small genomic interval known to contain 2 loci with >100 alt alleles. This enables:
   - Verifying correct behavior using `count_rows()`.
   - Switching from loading all loci for 10 samples → to loading a small interval for all samples (excluding pre-filtered ones), which:
     - Preserves correct sample count filtering.
     - Reduces runtime in test mode (now it takes ~14 mins).
    
Test run in this [notebook](https://workbench.researchallofus.org/workspaces/aou-rw-4b0bd63f/gnomadproduction/analysis/preview/hard_filter_02_vds_sample_qc.ipynb). 
     
BTW, I tried to load first partitions or filter_partitions, they all seemed to be slow, I really prefer `hl.vds.filter_intervals()`

### DIFF
--- a/gnomad_qc/v5/resources/basics.py
+++ b/gnomad_qc/v5/resources/basics.py
@@ -313,7 +313,7 @@ def get_aou_failing_genomic_metrics_samples() -> hl.Table:
 
 
 def add_project_prefix_to_sample_collisions(
-    ht: hl.Table,
+    t: Union[hl.Table, hl.MatrixTable],
     sample_collisions: hl.Table,
     project: Optional[str] = None,
     sample_id_field: str = "s",
@@ -321,7 +321,7 @@ def add_project_prefix_to_sample_collisions(
     """
     Add project prefix to sample IDs that exist in multiple projects.
 
-    :param ht: Table to add project prefix to sample IDs.
+    :param t: Table or MatrixTable to add project prefix to.
     :param sample_collisions: Table of sample IDs that exist in multiple projects.
     :param project: Optional project name to prepend to sample collisions. If not set, will use 'ht.project' annotation. Default is None.
     :param sample_id_field: Field name for sample IDs in the table.
@@ -330,26 +330,39 @@ def add_project_prefix_to_sample_collisions(
     logger.info(
         "Adding project prefix to sample IDs that exists in multiple projects..."
     )
-    collisions = sample_collisions.aggregate(hl.agg.collect_as_set(sample_collisions.s))
-    ht = ht.key_by()
+    # Get set of sample IDs that are in collisions
+    collision_ids = hl.literal(
+        sample_collisions.aggregate(hl.agg.collect_as_set(sample_collisions.s))
+    )
 
+    # Determine if input is a Table or MatrixTable
+    is_matrix = isinstance(t, hl.MatrixTable)
+
+    # Get the appropriate expression to use as prefix
     if project is not None:
         prefix_expr = hl.literal(project)
     else:
         try:
-            prefix_expr = ht["project"]
-        except KeyError:
+            prefix_expr = t.project
+        except Exception:
             raise ValueError(
-                "No project name provided and 'ht' does not contain a 'project' field."
+                "No project provided and 'project' field not found in the input table or MatrixTable columns."
             )
 
-    ht = ht.annotate(
-        **{
-            f"{sample_id_field}": hl.if_else(
-                hl.literal(collisions).contains(ht["s"]),
-                hl.delimit([prefix_expr, ht[sample_id_field]], "_"),
-                ht[sample_id_field],
-            )
-        }
+    # Build new sample ID expression
+    sample_id_expr = hl.if_else(
+        collision_ids.contains(t[sample_id_field]),
+        hl.delimit([prefix_expr, t[sample_id_field]], "_"),
+        t[sample_id_field],
     )
-    return ht.key_by(sample_id_field)
+
+    if is_matrix:
+        t = t.key_cols_by()
+        t = t.annotate_cols(**{sample_id_field: sample_id_expr})
+        t = t.key_cols_by(sample_id_field)
+    else:
+        t = t.key_by()
+        t = t.annotate(**{sample_id_field: sample_id_expr})
+        t = t.key_by(sample_id_field)
+
+    return t
--- a/gnomad_qc/v5/resources/basics.py
+++ b/gnomad_qc/v5/resources/basics.py
@@ -142,24 +142,21 @@ def get_aou_vds(
     :return: AoU v8 VDS.
     """
     aou_v8_resource = aou_test_dataset if test else aou_genotypes
-    vds = aou_v8_resource.vds()
 
     if isinstance(chrom, str):
         chrom = [chrom]
 
-    if autosomes_only and sex_chr_only:
+    if autosomes_only or sex_chr_only:
+        rg = aou_v8_resource.vds().reference_genome
+        sex_chrom = set(rg.x_contigs + rg.y_contigs)
+        if sex_chr_only:
+            chrom = list(sex_chrom)
+        else:
+            chrom = list(set(rg.contigs) - (sex_chrom | set(rg.mt_contigs)))
+    elif autosomes_only and sex_chr_only:
         raise ValueError(
             "Only one of 'autosomes_only' or 'sex_chr_only' can be set to True."
         )
-
-    # Apply chromosome filtering.
-    if sex_chr_only:
-        vds = hl.vds.filter_chromosomes(vds, keep=["chrX", "chrY"])
-    elif autosomes_only:
-        vds = hl.vds.filter_chromosomes(vds, keep_autosomes=True)
-    elif chrom and len(chrom) > 0:
-        logger.info("Filtering to chromosome(s) %s...", chrom)
-        vds = hl.vds.filter_chromosomes(vds, keep=chrom)
 
     if n_partitions and chrom:
         logger.info(
@@ -183,6 +180,7 @@ def get_aou_vds(
         )
         vds = hl.vds.VariantDataset(reference_data, variant_data)
     elif n_partitions:
+        logger.info("Loading VDS with %s partitions...", n_partitions)
         vds = hl.vds.read_vds(aou_v8_resource.path, n_partitions=n_partitions)
     else:
         vds = aou_v8_resource.vds()
--- a/gnomad_qc/v5/resources/basics.py
+++ b/gnomad_qc/v5/resources/basics.py
@@ -107,6 +107,7 @@ def get_aou_vds(
     split: bool = False,
     filter_samples: Optional[Union[List[str], hl.Table]] = None,
     test: bool = False,
+    n_partitions: Optional[int] = None,
     filter_partitions: Optional[List[int]] = None,
     chrom: Optional[Union[str, List[str], Set[str]]] = None,
     autosomes_only: bool = False,
@@ -126,6 +127,7 @@ def get_aou_vds(
         rather than grab an already split VDS. Default is False.
     :param filter_samples: Optional samples to filter the VDS to. Can be a list of sample IDs or a Table with sample IDs.
     :param test: Whether to load the test VDS instead of the full VDS. The test VDS includes 10 samples selected from the full dataset for testing purposes. Default is False.
+    :param n_partitions: Optional argument to read the VDS with a specific number of partitions.
     :param filter_partitions: Optional argument to filter the VDS to a list of specific partitions.
     :param chrom: Optional argument to filter the VDS to a specific chromosome(s).
     :param autosomes_only: Whether to include only autosomes. Default is False.
@@ -158,6 +160,35 @@ def get_aou_vds(
     elif chrom and len(chrom) > 0:
         logger.info("Filtering to chromosome(s) %s...", chrom)
         vds = hl.vds.filter_chromosomes(vds, keep=chrom)
+
+    if n_partitions and chrom:
+        logger.info(
+            "Filtering to chromosome(s) %s with %s partitions...", chrom, n_partitions
+        )
+        reference_data = hl.read_matrix_table(
+            hl.vds.VariantDataset._reference_path(aou_v8_resource.path)
+        )
+        reference_data = hl.filter_intervals(
+            reference_data,
+            [hl.parse_locus_interval(x, reference_genome="GRCh38") for x in chrom],
+        )
+        intervals = reference_data._calculate_new_partitions(n_partitions)
+        reference_data = hl.read_matrix_table(
+            hl.vds.VariantDataset._reference_path(aou_v8_resource.path),
+            _intervals=intervals,
+        )
+        variant_data = hl.read_matrix_table(
+            hl.vds.VariantDataset._variants_path(aou_v8_resource.path),
+            _intervals=intervals,
+        )
+        vds = hl.vds.VariantDataset(reference_data, variant_data)
+    elif n_partitions:
+        vds = hl.vds.read_vds(aou_v8_resource.path, n_partitions=n_partitions)
+    else:
+        vds = aou_v8_resource.vds()
+        if chrom:
+            logger.info("Filtering to chromosome %s...", chrom)
+            vds = hl.vds.filter_chromosomes(vds, keep=chrom)
 
     # Count initial number of samples.
     n_samples_before = vds.variant_data.count_cols()
--- a/gnomad_qc/v5/resources/basics.py
+++ b/gnomad_qc/v5/resources/basics.py
@@ -351,22 +351,19 @@ def add_project_prefix_to_sample_collisions(
 
     if is_matrix:
         t = t.key_cols_by()
-        sample_id_expr = hl.if_else(
-            collision_ids.contains(t[sample_id_field]),
-            hl.delimit([prefix_expr, t[sample_id_field]], "_"),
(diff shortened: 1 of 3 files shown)


gnomad_qc/v5/resources/basics.py
```diff
@@ -187,6 +187,15 @@ def get_aou_vds(
     n_samples_after = vds.variant_data.count_cols()
     logger.info("Removed %d samples from VDS.", n_samples_before - n_samples_after)
 
+    # Remove the chr4 site with excessive numbers of alleles (n=22233) which
+    # might cause memory issue later on
```
ch-kr, line 191:
```suggestion
    # Remove the chr4 site with excessive numbers of alleles (n=22233) to
    # avoid memory issues with `split_multi`.
```
copying the format of https://github.com/broadinstitute/gnomad_qc/blob/d4159eba1bc3e0ba99f38953925329ebe411f810/gnomad_qc/v4/resources/basics.py#L188

gnomad_qc/v5/sample_qc/hard_filters.py
```diff
@@ -59,11 +61,27 @@ def compute_aou_sample_qc(
...
 
+    logger.info("Excluding loci with more than 100 alternative alleles...")
+    vmt = vds.variant_data
+
+    if test:
+        logger.info(
+            "Rows before filtering loci with >100 alt alleles: %s...", vmt.count_rows()
+        )
+
+    vmt = vmt.filter_rows(vmt.n_unsplit_alleles < 101)
```
ch-kr, line 72: where is this `n_unsplit_alleles` annotation added?

gnomad_qc/v5/sample_qc/hard_filters.py
```diff
@@ -84,10 +103,11 @@ def main(args):
             n_partitions=args.sample_qc_n_partitions,
             test=test,
         )
-        logger.info("")
         ht = add_project_prefix_to_sample_collisions(
             ht, project="aou", sample_collisions=sample_id_collisions.ht()
         )
+        # Note: The prefix "aou" is only added to the 'all' Table, but not to the
```
ch-kr, line 109: should "Note" here be capitalized ("NOTE")? also, why do we want the collisions renamed on only the "all" table and not the other two tables?

gnomad_qc/v5/sample_qc/hard_filters.py
```diff
@@ -38,17 +39,18 @@ def compute_aou_sample_qc(
     :param test: If true, test the function on a smaller subset of the data.
     :return: Table containing sample QC metrics
     """
-    logger.info("Computing sample QC")
-    logger.info("Loading test VDS..." if test else "Loading VDS...")
-
     if test:
```
ch-kr, line 42: thanks for testing this filtering. however, I'm not sure this is the testing behavior we want to keep in the code. this test dataset code is really testing the behavior of `filter_rows`, which we shouldn't need to test, and `n_unsplit_alleles`, which also seems like something that shouldn't require testing -- the behavior of these is good to verify, but I'm not sure we need this in the repo long term

ch-kr (changes requested):
thanks for adding! a few questions

gnomad_qc/v5/sample_qc/hard_filters.py
ch-kr: I also didn't realize until just now, but this test overwrote the test referenced in the earlier PR. maybe PR tests should be run in a notebook like z_PR_(number)_test.ipynb (or another naming)

gnomad_qc/v5/sample_qc/hard_filters.py
KoalaQin: It's added by the `split` parameter in the loading function.

gnomad_qc/v5/sample_qc/hard_filters.py
KoalaQin: Yeah, I could rename the test notebook. 
The problem of using the test VDS as before is that we won't see any difference when we use that n_unsplit_alleles, because 10 sample wouldn't have >100 alt alleles, the results before and after are the same. What do you think it's a better strategy? I'm open to suggestions.

gnomad_qc/v5/sample_qc/hard_filters.py
KoalaQin: I think I will have to change more code to rename those temp files too.

gnomad_qc/v5/sample_qc/hard_filters.py
KoalaQin: Do you think it's better just to switch back to the old test?

ch-kr: yeah, I think it's better to keep the older test but add some documentation that the >100 n alt alleles filtering was tested and will not impact the original test dataset

gnomad_qc/v5/sample_qc/hard_filters.py
ch-kr: thanks!

this update makes sense to me (setting split=False and removing these loci prior to splitting multiallelics)

gnomad_qc/v5/sample_qc/hard_filters.py
```diff
@@ -59,11 +60,21 @@ def compute_aou_sample_qc(
         vds, intervals=telomeres_and_centromeres.ht(), keep=False
     )
 
+    logger.info("Excluding loci with more than 100 alternative alleles...")
+    vmt = vds.variant_data
+    vmt = vmt.annotate_rows(n_unsplit_alleles=hl.len(vmt.alleles))
+    vmt = vmt.filter_rows(vmt.n_unsplit_alleles < 101)
```
ch-kr, line 66:
```suggestion
    vmt = vmt.filter_rows(vmt.n_unsplit_alleles < 100)
```
I don't think we need to keep n_unsplit_alleles ==  100 (though using 101 or 100 here shouldn't make a huge difference)

gnomad_qc/v5/sample_qc/hard_filters.py
```diff
@@ -72,31 +83,60 @@ def compute_aou_sample_qc(
     return sample_qc_ht.naive_coalesce(n_partitions)
 
 
+def rename_sample_collisions_in_outputs(
```
ch-kr, line 86: thank you for working on this code to rename the sample IDs!

looking at this again, I wonder if it's better to update `add_project_prefix_to_sample_collisions` to be more flexible and operate on either a Table or MatrixTable. that way, you could add the project prefix to the sample collisions prior to calling `compute_stratified_sample_qc`, and you wouldn't need to adjust each of the three output tables separately.

while we're on this topic -- I know I'm the one that flagged the project collisions because it was listed under hard filtering on Mike's flowchart. however, just to double check: is it required to have the project collisions annotated on the sample QC output HTs?

ch-kr (commented):
a couple conceptual questions about the sample collision renaming

gnomad_qc/v5/sample_qc/hard_filters.py
KoalaQin: Let's align with AoU? Because they used alt alleles >100 and I got the same count.

gnomad_qc/v5/sample_qc/hard_filters.py
KoalaQin: I think it's nice to have them renamed, make it more convenient for later merging to project meta etc...

gnomad_qc/v5/sample_qc/hard_filters.py
KoalaQin: I found that `hl.vds.sample_qc` and `hl.sample_qc` (since now I'm using vmt instead of vds) will get a bit different results: 
Screenshot 2025-05-16 at 1 45 04 PM
They don't have exactly the same fields, but the major counts are the same, do I need to put it back to a VDS?

gnomad_qc/v5/sample_qc/hard_filters.py
KoalaQin: I could try to change the other function

KoalaQin: But I'm thinking twice `key_cols_by()` might make it slow?

ch-kr: good point. maybe if the input is a MT, the function shouldn't do the rekey. what do you think?

gnomad_qc/v5/sample_qc/hard_filters.py
ch-kr: yeah, let's convert the MT back into a VDS to get the set of sample QC metrics we used for v4

gnomad_qc/v5/sample_qc/hard_filters.py
KoalaQin: I got this error if I don't rekey it: 
```
ValueError: VDS.validate: expect variant data to have a single col key of type string, found struct{}
```
I can drop the changes in the prefix function and add the prefix later on when we're merging it to meta?

KoalaQin:
No prefix at this step, I don't want to make it slow with `key_cols_by()`, could you review while I'm running the test?

gnomad_qc/v5/sample_qc/hard_filters.py
```diff
@@ -28,7 +28,7 @@ def compute_aou_sample_qc(
     """
     Perform sample QC on AoU VDS.
 
-    .. note::
+    ..note::
```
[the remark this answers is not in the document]
ch-kr: bumping this

gnomad_qc/v5/sample_qc/hard_filters.py
ch-kr: it looks like that error is because of this block:
image
it's a bit uglier, but wouldn't it work if you annotated the updated sample IDs as a new column name? in the case that the input is a HT, you can rekey by renaming that new field back to the desired sample ID field, and in the case that the input is a MT, you can return the annotation as-is so that you can re-key after extracting only the columns from the MT.

your last suggestion is also fine -- please add documentation that the prefix needs to be added later on if you choose that route

gnomad_qc/v5/sample_qc/hard_filters.py
```diff
@@ -59,6 +60,21 @@ def compute_aou_sample_qc(
         vds, intervals=telomeres_and_centromeres.ht(), keep=False
     )
 
+    logger.info("Excluding loci with more than 100 alternative alleles...")
+    # NOTE: Filtering loci with >100 alleles has no effect on the test dataset,
+    # since it contains no such loci. However, this filtering is still necessary
+    # on the full dataset because the number of alleles can change after removing
+    # samples from the raw VDS. As a result, it may not always match the 'EXCESS_ALLELES'
```
ch-kr, line 67: can you expand on what `EXCESS_ALLELES` is here or add a link? it's good to assume whoever is reading this script in the future doesn't know or doesn't remember what that field is

gnomad_qc/v5/sample_qc/hard_filters.py
```diff
@@ -96,7 +107,7 @@ def get_script_argument_parser() -> argparse.ArgumentParser:
     parser = argparse.ArgumentParser()
     parser.add_argument(
         "--overwrite",
-        help="Overwrite all Matrixtables/Tables. (default: False).",
+        help="Overwrite all Matrix Tables/Tables. (default: False).",
```
ch-kr, line 110:
```suggestion
        help="Overwrite all MatrixTables/Tables. (default: False).",
```

gnomad_qc/v5/sample_qc/hard_filters.py
```diff
@@ -59,6 +56,23 @@ def compute_aou_sample_qc(
...
     )
 
+    logger.info("Excluding loci with more than 100 alternative alleles...")
+    # NOTE: Filtering loci with >100 alleles has no effect on the test dataset,
+    # since it contains no such loci. However, this filtering is still necessary
+    # on the full dataset because the number of alleles can change after removing
+    # samples from the raw VDS. As a result, it may not always match the ``EXCESS_ALLELES`` flag
+    # in the ``filters`` field of the raw VDS. (AoU v8 defined this as a site with >100 alternate alleles
+    # in their quality report: `All of Us Genomic Quality Report
+    # <https://support.researchallofus.org/hc/en-us/articles/29390274413716-All-of-Us-Genomic-Quality-Report>`_)
```
ch-kr, line 66: I can't build the docs right now because I'm getting this error
```
Traceback (most recent call last):
  File "/Users/kchao/code/gnomad_qc/docs/./generate_api_reference.py", line 221, in <module>
    write_package_doc(pkg)
  File "/Users/kchao/code/gnomad_qc/docs/./generate_api_reference.py", line 190, in write_package_doc
    write_package_doc(full_module_name)
  File "/Users/kchao/code/gnomad_qc/docs/./generate_api_reference.py", line 193, in write_package_doc
    write_module_doc(full_module_name)
  File "/Users/kchao/code/gnomad_qc/docs/./generate_api_reference.py", line 154, in write_module_doc
    module = importlib.import_module(module_name)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/kchao/miniconda3/envs/gnomad-qc/lib/python3.11/importlib/__init__.py", line 126, in import_module
    return _bootstrap._gcd_import(name[level:], package, level)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<frozen importlib._bootstrap>", line 1204, in _gcd_import
  File "<frozen importlib._bootstrap>", line 1176, in _find_and_load
  File "<frozen importlib._bootstrap>", line 1147, in _find_and_load_unlocked
  File "<frozen importlib._bootstrap>", line 690, in _load_unlocked
  File "<frozen importlib._bootstrap_external>", line 940, in exec_module
  File "<frozen importlib._bootstrap>", line 241, in _call_with_frames_removed
  File "/Users/kchao/code/gnomad_qc/gnomad_qc/v4/sample_qc/relatedness.py", line 37, in <module>
    from gnomad_qc.v4.sample_qc.cuKING.cuking_outputs_to_ht import cuking_outputs_to_ht
ModuleNotFoundError: No module named 'gnomad_qc.v4.sample_qc.cuKING.cuking_outputs_to_ht'
Running Sphinx v6.2.1
making output directory... done
[autosummary] generating autosummary for: api_reference/v4/annotations/compute_coverage.rst, api_reference/v4/annotations/fix_freq_an.rst, api_reference/v4/annotations/generate_freq.rst, api_reference/v4/annotations/generate_freq_genomes.rst, api_reference/v4/annotations/generate_variant_qc_annotations.rst, api_reference/v4/annotations/index.rst, api_reference/v4/annotations/insilico_predictors.rst, api_reference/v4/annotations/vep_context_ht.rst, api_reference/v4/annotations/vrs_annotation_batch.rst, api_reference/v4/assessment/calculate_per_sample_stats.rst, ..., api_reference/v5/data_ingestion/field_md_to_html.rst, api_reference/v5/data_ingestion/index.rst, api_reference/v5/index.rst, api_reference/v5/resources/basics.rst, api_reference/v5/resources/constants.rst, api_reference/v5/resources/index.rst, api_reference/v5/resources/meta.rst, api_reference/v5/resources/rwb_utils.rst, api_reference/v5/resources/sample_qc.rst, index.rst
INFO (gnomad.utils.filtering 659): hgdp missing from ['non_v2', 'non_topmed', 'non_cancer', 'controls_and_biobanks', 'non_neuro']
INFO (gnomad.utils.filtering 659): tgp missing from ['non_v2', 'non_topmed', 'non_cancer', 'controls_and_biobanks', 'non_neuro']
loading intersphinx inventory from https://hail.is/docs/0.2/objects.inv...
myst v2.0.0: MdParserConfig(commonmark_only=False, gfm_only=False, enable_extensions=set(), disable_syntax=[], all_links_external=False, url_schemes=('http', 'https', 'mailto', 'ftp'), ref_domains=None, fence_as_directive=set(), number_code_blocks=[], title_to_header=False, heading_anchors=0, heading_slug_func=None, html_meta={}, footnote_transition=True, words_per_minute=200, substitutions={}, linkify_fuzzy_links=True, dmath_allow_labels=True, dmath_allow_space=True, dmath_allow_digits=True, dmath_double_inline=False, update_mathjax=True, mathjax_classes='tex2jax_process|mathjax_process|math|output_area', enable_checkboxes=False, suppress_warnings=[], highlight_code_blocks=True)
building [mo]: targets for 0 po files that are out of date
writing output...
building [html]: targets for 51 source files that are out of date
updating environment: [new config] 51 added, 0 changed, 0 removed
reading sources... [100%] index

Warning, treated as error:
/Users/kchao/code/gnomad_qc/docs/index.rst:37:toctree contains reference to nonexisting document 'api_reference/index'
```
so I am assuming this formatting looks ok (and we should fix the docs at some point)

ch-kr (approved):
one minor thing, then LGTM

gnomad_qc/v5/sample_qc/hard_filters.py
KoalaQin: Yeah, I'm getting some errors too, seems not the same one, I will make another PR once I find a solution.

PR #676, opened 2025-06-03
Add code to print cuKING command

ch-kr:
Major changes:
- Adds `relatedness.py` script with two sections: one to print cuKING command and another to store code calling `mt_to_cuking_inputs` (though this will be run in a notebook)

Minor changes:
- Adds `get_cuking_input_path` resource to sample QC resources
- Adds `get_cuking_output_path` resource to sample QC resources

### DIFF
--- a/gnomad_qc/v5/sample_qc/relatedness.py
+++ b/gnomad_qc/v5/sample_qc/relatedness.py
@@ -117,7 +117,7 @@ def main(args):
             # We also cannot run `mt_to_cuking_inputs.py` directly because
             # Hail needs to be initialized with a temporary directory
             # to avoid memory errors.
-            joint_qc_mt = get_joint_qc(test=test).mt
+            joint_qc_mt = get_joint_qc(test=test).mt()
             check_resource_existence(
                 output_step_resources={"cuking_input_parquet": get_cuking_input_path()}
             )


gnomad_qc/v5/sample_qc/relatedness.py
```diff
@@ -0,0 +1,180 @@
+"""Script to compute relatedness estimates among pairs of samples in the joint gnomAD v4 (exomes + genomes) + AoU callset."""
+
+import argparse
+import logging
+import textwrap
+
+import hail as hl
+
+# from gnomad_qc.resource_utils import check_resource_existence
```
jkgoodrich, line 9: Is thing commented because it will be used later?

gnomad_qc/v5/sample_qc/relatedness.py
```diff
@@ -0,0 +1,180 @@
...
+    logger.info(
+        "Printing a command that can be used to submit a dsub job to run "
+        "cuKING on the files created by --prepare-cuking-inputs."
+    )
+    logger.warning(
+        "This printed command assumes that the cuKING directory is in the same "
+        "location where the command is being run and that $PROJECT_ID is set!"
+    )
+    SPLIT_FACTOR = cuking_split_factor
+    TOTAL_SHARDS = SPLIT_FACTOR * (SPLIT_FACTOR + 1) / 2
```
jkgoodrich, line 66: Maybe just add a comment for what TOTAL_SHARDS is and why it's computed this way. 

Also, since these are not constants, do they need to be capitalized and does split factor need to be redefined?

gnomad_qc/v5/sample_qc/relatedness.py
```diff
@@ -0,0 +1,180 @@
...
+                --shard-index=SHARD_INDEX" \\
+            done
+            """
+        )
+    )
+
+
+def main(args):
+    """Compute relatedness estimates among pairs of samples in the callset."""
+    # test = args.test
```
jkgoodrich, line 91: Assuming these will be used later? Did you want to add a test?

gnomad_qc/v5/sample_qc/relatedness.py
```diff
@@ -0,0 +1,180 @@
...
+        )
+        return
+
+    hl.init(
+        log="/home/jupyter/workspaces/gnomadproduction/relatedness.log",
+        tmp_dir=f"gs://{WORKSPACE_BUCKET}/tmp/4_day",
+    )
+    hl.default_reference("GRCh38")
+
+    # joint_qc_mt = get_joint_qc().mt()
```
jkgoodrich, line 110: Used later?

gnomad_qc/v5/sample_qc/relatedness.py
```diff
@@ -0,0 +1,180 @@
...
+    )
+
+
+def main(args):
+    """Compute relatedness estimates among pairs of samples in the callset."""
+    # test = args.test
+    # overwrite = args.overwrite
+    min_emission_kinship = args.min_emission_kinship
+
+    if args.print_cuking_command:
```
jkgoodrich, line 95: Where is the QC MT converted to the parquet files?

jkgoodrich (commented):
Just a few questions

gnomad_qc/v5/sample_qc/relatedness.py
ch-kr: I added these initially with code to run the QC MT > parquet conversion but ran into issues with that (will respond in more detail in the other thread)

gnomad_qc/v5/sample_qc/relatedness.py
ch-kr: this will be done in a notebook (https://workbench.researchallofus.org/workspaces/aou-rw-4b0bd63f/gnomadproduction/analysis/preview/relatedness_02_prep_AoU_and_gnomAD_data_for_cuKING.ipynb) -- I initially had the code in this PR and ran into an import error. I also noticed that Mike or Kristen commented in another notebook that the Hail init needs to include a tmp dir to avoid memory issues so was planning to run this conversion in the notebook above following the example set on v7 (see comment for more details)
image

ch-kr: also -- happy to put the code back here if you happen to know how to fix the import error 😆

gnomad_qc/v5/sample_qc/relatedness.py
jkgoodrich: no problem, just making sure

gnomad_qc/v5/sample_qc/relatedness.py
jkgoodrich: Ah, so I think that is all run on the terminal, so they should be included in the print

jkgoodrich: I don't know enough about how the quotes are handled to know if the dsub --command will work as expected, but it looks the same as in the screenshot, so if that worked, then it should be fine

jkgoodrich (commented):
A couple small things

jkgoodrich (approved):
lgtm!

PR #677, opened 2025-06-05
Fix print cuKING command

ch-kr:
PR fixes minor typos missed in printing cuKING command, updates default machine type, and adds notebook used to set up dsub

### DIFF
--- a/gnomad_qc/v5/sample_qc/relatedness.py
+++ b/gnomad_qc/v5/sample_qc/relatedness.py
@@ -21,8 +21,8 @@ def print_cuking_command(
     min_emission_kinship: float = 0.5,
     cuking_split_factor: int = 4,
     location: str = "us-central1",
-    machine_type: str = "a2-highgpu-1g",
-    accelerator_count: int = 1,
+    machine_type: str = "a2-highgpu-2g",
+    accelerator_count: int = 2,
     accelerator_type: str = "nvidia-tesla-a100",
     requester_pays_project: str = "terra-vpc-sc-93ccd8d2",
 ) -> None:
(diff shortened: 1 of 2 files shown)


gnomad_qc/v5/sample_qc/relatedness.py
```diff
@@ -59,29 +60,24 @@ def print_cuking_command(
...
+        for SHARD_INDEX in $(seq 0 $((TOTAL_SHARDS - 1))); do
+        cuKING_dsub \\
+        --location={location} \\
+        --machine-type={machine_type} \\
+        --accelerator-count={accelerator_count} \\
+        --accelerator-type={accelerator_type} \\
+        --command="cuking \\
+        --input_uri="{cuking_input_path}" \\
+        --output_uri=f"{cuking_output_path}/out_split_${{SHARD_INDEX}}.parquet" \\
+        --requester_pays_project={requester_pays_project} \\
```
mike-w-wilson, line 75: FMK, Is there a reason `input_uri` needs quotes but `requester_pays_project` does not?  It looks like this is  exactly what ran in the NB, well what was pasted, so I'm good approving this but I think in hindsight, I'd make the nested  `--command` into a defined variable above the cuking_command defintion to make it a bit more readable (the opposite of this run-on sentence). This is sorta what co-pilot is getting at but I dont think that needs to be done now since we ran this already.

mike-w-wilson (approved):
👍

gnomad_qc/v5/sample_qc/relatedness.py
klaricch: i think it was just that requester_pays_project uses more simple characters/no variables although in this case i don't think it matters

mike-w-wilson (commented):
The script looks good now, shard_index shows up in the command how you'd expect. What is the reasoning behind including one of the relatedness notebooks?

PR #684, opened 2025-06-24
Option to use public sites HT for summary_stats.py

matren395:
The sites release HT has been moved to (only) the public bucket. Add an arg to optionally call that.

### DIFF
--- a/gnomad_qc/v4/assessment/summary_stats.py
+++ b/gnomad_qc/v4/assessment/summary_stats.py
@@ -35,10 +35,11 @@ def main(args):
     test = args.test
     overwrite = args.overwrite
     data_type = args.data_type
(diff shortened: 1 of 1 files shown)


gnomad_qc/v4/assessment/summary_stats.py
```diff
@@ -35,10 +35,11 @@ def main(args):
     test = args.test
     overwrite = args.overwrite
     data_type = args.data_type
+    public = args.get_public_sites
```
ch-kr, line 38: since this is an argument, I don't think public needs to be True by default in the resources (but I also don't see a reason why we shouldn't read from the public bucket rather than our own buckets by default) 😆

ch-kr (approved):
LGTM

PR #693, opened 2025-07-15
add no heap heap memory

klaricch:


ch-kr (approved):
LGTM! one minor comment to rename PR (add no heap heap memory > add no off heap memory)

PR #695, opened 2025-08-01
Subset v4 genomes

mike-w-wilson:


ch-kr (commented):
thank you for updating this script and finding the `dataclass` option! I like that stores all of the options we pass into the script and even does some quick validation that the passed arguments make sense together.

just a few questions

gnomad_qc/v4/subset.py
```diff
@@ -2,268 +2,429 @@
...
-                parallel="header_per_shard",
-                tabix=True,
-            )
+    return mt
+
+
+def process_metadata_export(
+    meta_ht: hl.Table,
+    vmt: hl.MatrixTable,
+    keep_data_paths: bool,
```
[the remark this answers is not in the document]
mike-w-wilson: I went back and forth on this. I settled on if more than one field needed from the config, pass the whole config, otherwise pass the single field however passing the single config may be the better approach. I think I'll change it

gnomad_qc/v4/subset.py
```diff
@@ -2,268 +2,429 @@
...
+    if config.split_multi:
+        logger.info("Splitting multi-allelics...")
+        mtds = apply_split_multi_logic(mtds, config)
+
+    logger.info(
+        "Applying min_rep to the variant data MT because remove_dead_alleles in"
+        " hl.vds.filter_samples may result in variants that do not have the minimum"
+        " representation..."
+    )
+    mtds = apply_min_rep_logic(mtds, config)
```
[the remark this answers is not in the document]
mike-w-wilson: Sorry missed an else statement here, this is only meant for an unsplit VDS

gnomad_qc/v4/subset.py
```diff
@@ -2,268 +2,429 @@
...
+    :param mtds: MatrixTable or VariantDataset to process.
+    :param config: Processing configuration.
+    :return: MatrixTable or VariantDataset with variant QC annotations.
+    """
+    ds = mtds if config.output_vcf else mtds.variant_data
+    ht = release_sites(data_type=config.data_type).ht()
+    selected_variant_qc_annotations = {}
+
+    if config.variant_qc_annotations is None:
+        vqc_annotations = list(ht.row_value.keys())
```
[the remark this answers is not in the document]
ch-kr: nah happy to keep this if this is how we've been handling annotations in the past. maybe we should just add a note?

gnomad_qc/v4/subset.py
```diff
@@ -1,269 +1,471 @@
...
-        meta_ht = meta_ht.annotate(
-            project_meta=meta_ht.project_meta.drop(*data_to_drop)
-        )
-        meta_ht = meta_ht.checkpoint(
-            f"{output_path}/metadata.ht", overwrite=args.overwrite
-        )
-        meta_ht.export(f"{output_path}/metadata.tsv.bgz")
+    finally:
+        logger.info("Copying log to logging bucket...")
+        hl.copy_log(f"{config.output_path}/v4_subset.log")
```
ch-kr, line 453:
```suggestion
        hl.copy_log(get_logging_path("v4_subset"))
```
or maybe remove this import if you prefer writing out to `config.output_path`

gnomad_qc/v4/subset.py
```diff
@@ -1,269 +1,471 @@
...
+    :param meta_ht: The meta HT.
+    :param vds_cols: The VDS columns.
+    :return: The subset HT.
+    """
+    if subset_workspaces and not meta_ht:
+        raise ValueError("A meta HT must be provided if subsetting by workspaces.")
 
-    if args.subset_workspaces:
+    if subset_workspaces:
         terra_workspaces = hl.literal(subset_ht.terra_workspace.lower().collect())
```
ch-kr, line 140:
```suggestion
        terra_workspaces = hl.literal(meta_ht.terra_workspace.lower().collect())
```
I completely missed this earlier

ch-kr (commented):
just a few more minor things (including some I missed last time, sorry!)

gnomad_qc/v4/subset.py
mike-w-wilson: I'll remove the import. `get_logging_path` uses gnomad-tmp and adds  a "exomes_4.1_qc"  to the path and I didnt feel this belonged there.

gnomad_qc/v4/subset.py
mike-w-wilson: Oh its suppose to be subset_workspaces

gnomad_qc/v4/subset.py
```diff
@@ -1,269 +1,468 @@
...
+            tmp_dir=args.tmp_dir,
+        )
+
+
+def get_gnomad_datasets(data_type: str, n_partitions: Optional[int], test: bool):
+    """
+    Get requested data type's v4 VariantDataset.
+
+    :param data_type: Type of data to subset.
+    :param n_partitions: Number of desired partitions for the VDS, repartioned on read.
```
ch-kr, line 95:
```suggestion
    :param n_partitions: Number of desired partitions for the VDS, repartitioned on read.
```

gnomad_qc/v4/subset.py
```diff
@@ -1,269 +1,468 @@
...
+    :return: The subset HT.
     """
-    Make a dictionary of gnomAD release annotation expressions to annotate onto the subsetted data.
+    if subset_workspaces and not meta_ht:
+        raise ValueError("A meta HT must be provided if subsetting by workspaces.")
+
+    if subset_workspaces:
+        terra_workspaces = hl.literal(
+            subset_workspaces.terra_workspace.lower().collect()
+        )
```
ch-kr, line 137: should this be something like:
```
subset_workspaces_ht = hl.import_table(subset_workspaces)
terra_workspaces = hl.literal(
        subset_workspaces_ht.terra_workspace.lower().collect()
    )
```

gnomad_qc/v4/subset.py
```diff
@@ -1,269 +1,468 @@
...
-                **make_variant_qc_annotations_dict(vd.row_key, variant_qc_annotations)
-            )
-        vds = hl.vds.VariantDataset(vds.reference_data, vd)
 
-    if args.vcf or args.dense_mt:
-        logger.info("Densifying VDS")
-        mt = hl.vds.to_dense_mt(vds)
+        if config.output_vcf:
+            logger.info("Densifying VDS for VCF export...")
+            mtds = hl.vds.to_dense_mt(vds)
```
ch-kr, line 378: I completely forgot about this earlier, but does `vds.to_dense_mt` remove the need for an additional filter after to remove only reference rows? we had to remove these manually in the past with something like
```
mtds = mtds.filter_rows(hl.len(mtds.alleles) > 1)
```

mike-w-wilson: Yeah it should remove the ref only itself with this line: https://github.com/hail-is/hail/blob/458d54dbe14a7188720bf706d6acba1114adf77e/hail/python/hail/vds/methods.py#L57

PR #701, opened 2025-09-04
v4 subset fix sparse split multi filter changed loci

mike-w-wilson:
Missed adding in the filter change loci option. I also dropped a leftover print statement and added the ability to name the subset file itself.


klaricch (approved):
lgtm!

PR #702, opened 2025-09-05
ExpressionResource for failing metric aou doesnt exist

mike-w-wilson:
This resource doesn't exist. It looks like it was maybe rolled into the resource futher down in the file, `samples_to_exclude`


mike-w-wilson:
I'm now confused if this is needed or not? The exclude file exists which is generated by the function that is causing the pylint failure, though it doesnt write it out..

mike-w-wilson:
Closing after generating here: https://app.zenhub.com/workspaces/gnomad-5f4d127ea61afc001d6be50b/issues/gh/broadinstitute/gnomad_production/1994

PR #707, opened 2025-09-16
Fix samples_to_exclude when fileter_samples is None

mike-w-wilson:
If filter_samples is none, this function would  fail with

`TypeError: union: parameter 's': expected expression of type set<any>, found set: set()`

This fixes that.


ch-kr (approved):
thanks for fixing!

PR #717, opened 2025-09-25
Update metadata releasable field for gnomad samples that withdrew consent

mike-w-wilson:


gnomad_qc/v5/sample_qc/merge_project_meta.py
```diff
@@ -220,10 +220,18 @@ def select_only_final_fields(
     :return: Filtered Hail Table containing only fields required for the final schema.
     """
     # gnomAD v4 genome metadata has some samples with age_alt, which is the mean age of
-    # the age_bin field. This creates a single age field.
+    # the age_bin field. This creates a single age field. Also, 897 samples were consented
```
ch-kr, line 223: we added this [line](https://github.com/broadinstitute/gnomad_qc/blob/899e37e163bb92f78037f56cfad507e8e156d1b5/gnomad_qc/v5/sample_qc/relatedness.py#L245) to `get_consent_samples_to_drop` in the relatedness script, so should we add a note that we're going back to add this to the project meta now too?

ch-kr (approved):
one small comment but LGTM

PR #719, opened 2025-09-29
Show the GCS migration warning only once

mike-w-wilson:


gnomad_qc/v3/resources/resource_utils.py
```diff
@@ -0,0 +1,20 @@
+"""Centralized migration warnings for v3 resources moved to new GCS locations."""
+
+import functools
+import logging
+from typing import Literal
+
+
+@functools.lru_cache(maxsize=None)
```
ch-kr, line 8: I asked cursor about `maxsize` here, and it seemed to think `maxsize=2` would be slightly more optimal:
image

what do you think? also, I didn't know Literal was a type, that's cool

mike-w-wilson: Oh nice, yes I will update that, thanks for looking into it!

PR #727, opened 2025-11-14
Update AN/coverage code to support running in Batch

ch-kr:


mike-w-wilson:
This is covered in #772

PR #758, opened 2026-02-24
Add environment parameter to v5 resources and scripts for Hail Batch support

mike-w-wilson:
This adds the secure terra bucket as the resource for all v5 batch work.  I had to update a number of variables to functions like `get_*` so we could pass `environment`. I want to maintain support for RWB but any steps that we run or rerun will be through batch for now. Supporting dataproc for sites level work, which is desired,  is OOS for this PR.


gnomad_qc/v5/sample_qc/identify_trios.py
klaricch: nitpick but oder of this is flipped with what is used in the other scripts

klaricch (changes requested):

gnomad_qc/v5/resources/annotations.py
ch-kr: should this be `f"gs://{BATCH_BUCKET}/aou_sites_vcf/v8/echo_full_gnomad_annotated.sites-only.vcf.gz"`?

ch-kr (dismissed):
thank you for working on this!

one other comment: `meta` gets called without an environment arg in both the coverage and frequency scripts (not sure if you want to add to this PR or add to later PRs)

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
ch-kr: maybe this should also be updated since this has been changed in identify_trios.py?

ch-kr (commented):
a few more minor comments

klaricch (changes requested):

klaricch (changes requested):
this could be a good one to run the copilot review on again

klaricch (approved):
lgtm

PR #759, opened 2026-02-25
add release resources

klaricch:


mike-w-wilson (commented):
Mostly just flipping defaults from rwb to batch since we do not plan on running anything in RWB anymore and these are all writes.

mike-w-wilson (approved):
LGTM

PR #765, opened 2026-04-10
Filter rows before samples to avoid reading in the entire VDS and also make sample count logging optional

mike-w-wilson:
As written the VDS is read in full and all calls are rewritten since the sample filter `remove_dead_alleles` is true. This all happens before filter partitions. By filtering rows first, we remove extra unneeded work.  The work actually done here is just a reordering of operations. The only new piece is making sample count logging optional, so we dont run two counts just for the log statement every time we want the VDS.

### DIFF
--- a/gnomad_qc/v5/resources/basics.py
+++ b/gnomad_qc/v5/resources/basics.py
@@ -298,7 +298,7 @@ def get_aou_vds(
     :param naive_coalesce_partitions: Optional number of partitions to coalesce the VDS to. Default is None.
     :param add_project_prefix: Whether to prefix sample IDs (e.g., ``'aou_'``) for samples that exist in multiple projects to avoid ID collisions. Default is False.
     :param environment: Environment to use. Default is "batch". Must be one of "rwb" or "batch".
-    :param log_sample_counts: Whether to log sample counts before/after filtering.
+    :param log_sample_counts: Whether to log sample counts before/after filtering out samples to exclude.
         When False, skips the ``count_cols`` calls used for logging. Default is True.
     :return: AoU v8 VDS.
     """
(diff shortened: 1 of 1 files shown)


gnomad_qc/v5/resources/basics.py
```diff
@@ -297,6 +298,8 @@ def get_aou_vds(
     :param naive_coalesce_partitions: Optional number of partitions to coalesce the VDS to. Default is None.
     :param add_project_prefix: Whether to prefix sample IDs (e.g., ``'aou_'``) for samples that exist in multiple projects to avoid ID collisions. Default is False.
     :param environment: Environment to use. Default is "batch". Must be one of "rwb" or "batch".
+    :param log_sample_counts: Whether to log sample counts before/after filtering.
```
klaricch:
```suggestion
    :param log_sample_counts: Whether to log sample counts before/after filtering out samples to exclude.
```

klaricch (changes requested):
one pylint error

gnomad_qc/v5/resources/basics.py
klaricch: i think this change originally went it but was then dropped

klaricch (changes requested):

gnomad_qc/v5/resources/basics.py
mike-w-wilson:
```suggestion
    :param log_sample_counts: Whether to log sample counts before/after filtering out samples to exclude.
```

mike-w-wilson: Sorry the rebase messed with this, all set now!

klaricch (approved):

PR #767, opened 2026-04-15
Update behavior of file existence checks

ch-kr:
PR adds:
- Wrapper around `check_resource_existence` to avoid checking for file existence if the environment is Batch
- Update to `get_samples_to_exclude` to avoid checking if the samples to exclude resource exists if the environment is Batch

### DIFF
--- a/gnomad_qc/v5/resources/basics.py
+++ b/gnomad_qc/v5/resources/basics.py
@@ -176,6 +176,9 @@ def _check_resource_existence(
     :return: None.
     """
     if environment == "batch":
+        logger.info(
+            "Skipping resource existence check for Batch environment. Please rerun with --overwrite to overwrite existing resources."
+        )
         return
     check_resource_existence(
         input_step_resources=input_step_resources,
--- a/gnomad_qc/v5/resources/basics.py
+++ b/gnomad_qc/v5/resources/basics.py
@@ -1,7 +1,7 @@
 """Script containing generic resources."""
 
 import logging
-from typing import List, Optional, Set, Union
+from typing import Dict, List, Optional, Set, Union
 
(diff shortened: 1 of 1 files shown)


gnomad_qc/v5/resources/basics.py
```diff
@@ -151,6 +152,38 @@ def _init_hail(
...
+        of. Default is None.
+    :param output_step_resources: A dictionary with keys as pipeline step that generate
+        output files and the value as a list of the output files to check the existence
+        of. Default is None.
+    :param overwrite: The overwrite parameter used when writing the output files.
+        Default is False.
+    :return: None.
+    """
+    if environment == "batch":
+        return
```
klaricch, line 182: can add a log message here that it won't be checked

klaricch (changes requested):

gnomad_qc/v5/resources/basics.py
```diff
@@ -151,6 +152,41 @@ def _init_hail(
...
+    :param output_step_resources: A dictionary with keys as pipeline step that generate
+        output files and the value as a list of the output files to check the existence
+        of. Default is None.
+    :param overwrite: The overwrite parameter used when writing the output files.
+        Default is False.
+    :return: None.
+    """
+    if environment == "batch":
+        logger.info(
+            "Skipping resource existence check for Batch environment. Please rerun with --overwrite to overwrite existing resources."
```
klaricch:
```suggestion
            "Skipping resource existence check for Batch environment. To replace any existing outputs, run with --overwrite."
```

klaricch (changes requested):

klaricch (approved):

PR #768, opened 2026-04-28
add code for export_true_positive_vcfs

klaricch:


gnomad_qc/v5/resources/annotations.py
```diff
@@ -364,6 +364,36 @@ def get_variant_qc_annotations(
...
+    adj: bool = False,
+    true_positive_type: str = "transmitted_singleton",
+    environment: str = "batch",
+) -> str:
+    """
+    Provide the path to the true positive VCF used as input to VQSR.
+
+    :param version: Version of true positive VCF path to return.
+    :param test: Whether to use a tmp path for testing.
+    :param adj: Whether to use adj genotypes.
```
mike-w-wilson:
```suggestion
    :param version: Version of true positive VCF path to return. Default is CURRENT_ANNOTATION_VERSION.
    :param test: Whether to use a tmp path for testing. Default is False.
    :param adj: Whether to use adj genotypes. Default is False.
```

mike-w-wilson (approved):
one tiny doc suggestion then LGTM

PR #774, opened 2026-05-29
Add code to create truth sample VDS

ch-kr:
Add code to run gVCF combiner to create truth sample VDS


gnomad_qc/v5/resources/variant_qc.py
```diff
@@ -34,3 +36,49 @@ def _variant_qc_root(
...
+
+Stored for reference but not used directly.
+"""
+
+truth_samples_gvcf_paths = f"{_variant_qc_root(environment='batch')}/aou/truth_samples/truth_samples_gvcf_paths.tsv"
+"""
+Path to a single-column TSV listing the GCS path to each truth-sample gVCF (one per line).
+
+The sample IDs are intentionally not stored in this repo, and the truth-sample bucket cannot be listed
+The combiner therefore reads the gVCF paths from this manifest by known object paths rather than globbing the bucket.å
```
mike-w-wilson:
```suggestion
The combiner therefore reads the gVCF paths from this manifest by known object paths rather than globbing the bucket.
```

gnomad_qc/v5/resources/variant_qc.py
```diff
@@ -34,3 +36,49 @@ def _variant_qc_root(
...
+"""
+
+truth_samples_vds = VariantDatasetResource(
+    f"{_variant_qc_root(test=False)}/control_samples.vds",
+)
+"""
+VDS containing 8 GiaB samples.
+"""
```
mike-w-wilson, line 63: Is this fine to end up in our gnomad bucket? I know its GiaB but just wondering were its technically AoU generated sample level data.

gnomad_qc/v5/variant_qc/create_truth_samples_vds.py
```diff
@@ -0,0 +1,435 @@
...
+        "create_truth_samples_vds",
+        environment=environment,
+        billing_project=args.gcp_billing_project,
+        experimental=args.experimental,
+        **_get_batch_resource_kwargs(args),
+    )
+
+    manifest_path = args.manifest_path
+    output_path = (
+        f"{qc_temp_prefix(environment=environment)}truth_samples.vds"
```
mike-w-wilson, line 277: Nit: Prod uses control_samples.vds

gnomad_qc/v5/resources/variant_qc.py
```diff
@@ -34,3 +36,49 @@ def _variant_qc_root(
...
+
+truth_samples_aou_gvcfs_bucket = (
+    f"gs://{AOU_BUCKET}/wgs/short_read/snpindel/aux/qc/control_samples/"
+)
+"""
+AoU bucket containing gVCFs from 8 Genomes-in-a-Bottle (GiaB) samples sequenced with the same protocol as the AoU v8 data.
+
+Stored for reference but not used directly.
+"""
+
```
mike-w-wilson: > Stored for reference but not used directly.

Could this just be a note rather than declaring a variable?

gnomad_qc/v5/variant_qc/create_truth_samples_vds.py
```diff
@@ -0,0 +1,435 @@
+"""
+Script to create a VDS of the 8 Genomes-in-a-Bottle (GiaB) truth samples from their gVCFs.
+
+The GiaB gVCFs were sequenced with the same protocol as the AoU v8 data and live in
+``truth_samples_aou_gvcfs_bucket``. They are already reblocked, so they are passed
```
mike-w-wilson: Why not just list the bucket? This requires you to open another file to see what that is

gnomad_qc/v5/variant_qc/create_truth_samples_vds.py
```diff
@@ -0,0 +1,435 @@
...
+)
+from gnomad_qc.v5.resources.variant_qc import (
+    get_truth_samples_combiner_plan,
+    truth_samples_gvcf_paths,
+    truth_samples_vds,
+)
+
+logging.basicConfig(
+    format="%(asctime)s (%(name)s %(lineno)s): %(message)s",
+    datefmt="%m/%d/%Y %I:%M:%S %p",
```
mike-w-wilson, line 40: is force=true here  no longer needed to see logs? Realizing I never removed mine.

gnomad_qc/v5/variant_qc/create_truth_samples_vds.py
```diff
@@ -0,0 +1,435 @@
...
+    ``hl.Locus`` resolves the reference genome, which would initialize Hail on import
+    and break the later ``_init_hail`` call.
+
+    :return: Value Interval spanning :data:`TEST_INTERVAL`.
+    """
+    return hl.Interval(
+        hl.Locus(_TEST_CONTIG, _TEST_START, reference_genome="GRCh38"),
+        hl.Locus(_TEST_CONTIG, _TEST_END, reference_genome="GRCh38"),
+        includes_end=True,
+    )
```
mike-w-wilson, line 70: If you asked me this morning if hl.Locus would trigger an init I would say yes....neat.

gnomad_qc/v5/variant_qc/create_truth_samples_vds.py
```diff
@@ -0,0 +1,435 @@
...
+
+    manifest_path = args.manifest_path
+    output_path = (
+        f"{qc_temp_prefix(environment=environment)}truth_samples.vds"
+        if test
+        else truth_samples_vds.path
+    )
+
+    if args.create_truth_samples_vds:
```
mike-w-wilson, line 282: I'm not sure this is worth using but I wrote this years ago: https://github.com/broadinstitute/gnomad_qc/blob/main/gnomad_qc/federated/data_ingestion/create_vds.py which calls this wrapper function https://github.com/broadinstitute/gnomad_methods/blob/8adda7d8a40cc8f6325e26690c074efa827df6aa/gnomad/utils/file_utils.py#L227

gnomad_qc/v5/variant_qc/create_truth_samples_vds.py
```diff
@@ -0,0 +1,435 @@
...
+    """
+    with hfs.open(manifest_path) as f:
+        gvcf_paths = [line.strip() for line in f if line.strip().startswith("gs://")]
+
+    if not gvcf_paths:
+        raise ValueError(
+            f"No gVCF paths (lines starting with 'gs://') found in {manifest_path}."
+        )
+
+    logger.info("Read %d gVCF path(s) from %s.", len(gvcf_paths), manifest_path)
```
mike-w-wilson: I know well historically watch loggers but would it be easier/save compute if we assert the gvcf_paths length is 8? Kill the script with an error if it isnt ?

gnomad_qc/v5/resources/variant_qc.py
mike-w-wilson, line 63: Tiny but we may want to pin this to version=5.0 so that it doesnt break should we end up with multiple variant_qc  versions? I'm guessing  we wouldnt regenerate this

gnomad_qc/v5/resources/variant_qc.py
```diff
@@ -34,3 +36,49 @@ def _variant_qc_root(
...
+    The plan lets a failed or interrupted combiner run be resumed. It is written to the
+    writable batch variant QC tree (or a temp path for tests).
+
+    :param test: Whether to return a temporary test path. Default is False.
+    :return: Path to the combiner plan JSON.
+    """
+    if test:
+        return f"{qc_temp_prefix(environment='batch', days=4)}truth_samples_combiner_plan.json"
+
+    return f"{_variant_qc_root(environment='batch')}/aou/truth_samples/truth_samples_combiner_plan.json"
```
mike-w-wilson: Same logic with pinning a version here.

mike-w-wilson (commented):
Mostly minor things but one main question about data storage.

gnomad_qc/v5/resources/variant_qc.py
ch-kr: that's a good question. I was thinking yes since the GiaB samples are publicly available, and some of them are even consented for commercial redistribution

ch-kr: adding note here that we chatted about using the Batch bucket instead so will update

gnomad_qc/v5/variant_qc/create_truth_samples_vds.py
ch-kr: oops! I'll switch everything to truth samples for consistency with previous versions

gnomad_qc/v5/variant_qc/create_truth_samples_vds.py
ch-kr: oh yeah, I didn't even realize I removed this. it looks like yes based on the outputs from my test runs

gnomad_qc/v5/variant_qc/create_truth_samples_vds.py
ch-kr: oh no, I didn't realize this existed. 🤦‍♀️ I think this wouldn't work here because of the `hl.hadoop_copy` used to get the gVCF paths

gnomad_qc/v5/variant_qc/create_truth_samples_vds.py
ch-kr: although it doesn't show me when it created the batch so I'll add it back

ch-kr: adding the force=True back didn't show the batch ID still 🤔

mike-w-wilson (approved):
LGTM

PR #776, opened 2026-06-10
Update resource existence check

ch-kr:
I previously updated this to skip checks in Batch, but this only needs to skip checks for the read-only buckets (buckets that are not accessible with our user emails)

### DIFF
--- a/gnomad_qc/v5/resources/basics.py
+++ b/gnomad_qc/v5/resources/basics.py
@@ -2,7 +2,7 @@
 
 import logging
 from os import getenv
-from typing import Dict, List, Optional, Set, Union
+from typing import Dict, List, Optional, Set, Tuple, Union
 
 import hail as hl
 from gnomad.resources.resource_utils import (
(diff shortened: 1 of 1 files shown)


gnomad_qc/v5/resources/basics.py
```diff
@@ -228,9 +279,13 @@ def _check_resource_existence(
     """
     if environment == "batch":
         logger.info(
-            "Skipping resource existence check for Batch environment. To replace any existing outputs, run with --overwrite."
+            "Skipping resource existence checks for unstattable bucket resources "
+            "(read-only Batch and controlled-access AoU buckets). To replace any "
+            "existing outputs there, run with --overwrite."
```
mike-w-wilson: This tells us its happening but doesnt really inform us of anything more. I'm a bit on the fence if its worth adding but since its fairly quick to look through the dicts and paths, the logger could inform the script runner exactly which step/bucket is being skipped. We could also trigger the logger only when something is dropped? Curious what your thoughts are here.

mike-w-wilson (commented):
One comment but other than that question, looks good

gnomad_qc/v5/resources/basics.py
ch-kr: I like this idea; added

mike-w-wilson (approved):
LGTM

PR #777, opened 2026-06-12
Add CLAUDE.md

mike-w-wilson:


CLAUDE.md
```diff
@@ -0,0 +1,180 @@
+# CLAUDE.md
+
+Guidance for Claude Code when working in this repository.
+
+## Project Overview
```
ch-kr, line 5: should this also contain something describing that gnomad_qc isn't a library or installable package?

CLAUDE.md
```diff
@@ -0,0 +1,180 @@
...
+./lint --disable=W                                    # pylint, warnings disabled
+pydocstyle --match-dir='(?!v2|cuKING).*' gnomad_qc    # skips v2 and cuKING
+```
+
+### Docs
+```
+./docs/build.sh    # sphinx with -W (warnings are errors); CI publishes to gh-pages
+```
+
+## Architecture
```
ch-kr, line 89: I have this in my current version:
```
### Per-release versioning

Code is split by release: `gnomad_qc/v2/`, `v3/`, `v4/`, `v5/`. Each release directory follows the same structure, applied roughly in pipeline order:

- `resources/` — versioned input/output file paths and resource objects (Hail `TableResource`/`MatrixTableResource`/`VariantDatasetResource` from `gnomad_methods`). Single source of truth for where data lives in GCS; never hard-code paths in pipeline scripts.
- `sample_qc/` — hard filters, sex/ancestry inference, relatedness, outlier filtering, metadata HT.
- `annotations/` — info HT, coverage, frequencies, VEP, in-silico predictors, variant QC features.
- `variant_qc/` — random forest / VQSR training, evaluation, final filter.
- `create_release/` — combine annotations, prepare VCF release, validity checks.

`v3/README.md` documents the v3 pipeline step-by-step. `v4/README.md` is minimal; `v4/sample_qc/README.md` covers cuKING setup. v5 is the active development line.
```
we don't need to add all of this, but maybe we should add a section that says each version has the same structure (resources, sample_qc, annotations, variant_qc, release)?

CLAUDE.md
```diff
@@ -0,0 +1,180 @@
...
+  read/write/checkpoint I/O.
+- Standard flags: `--overwrite`, `--test`, and in v5 `--environment`.
+- A `finally` block that copies the Hail log to a logging path.
+- Parser built in a `get_script_argument_parser()`-style function, invoked
+  under `if __name__ == "__main__":`.
+
+v4 scripts manage inter-step dependencies with `PipelineStepResourceCollection`
+/ `check_resource_existence` from `gnomad_qc/resource_utils.py`, and optionally
+wrap `main` in `slack_notifications(...)` (token from `gnomad_qc/slack_creds.py`,
+channel from `--slack-channel`). v5 favors simpler `main()` bodies.
```
ch-kr, line 151: should we merge something about how pipeline scripts are run here? something like
```
### Script invocation pattern

Pipeline scripts expose argparse step flags (e.g. `--create-info-ht`, `--generate-trio-stats`, `--run-pc-relate`). A single script typically implements many discrete steps; `main(args)` dispatches by flag. When adding a step, follow the existing pattern: add a CLI flag, a corresponding output resource in `resources/<area>.py`, and gate I/O with `check_resource_existence` (`gnomad_qc/resource_utils.py`) so reruns fail fast on missing inputs and refuse to clobber outputs unless `--overwrite` is passed.
```

CLAUDE.md
ch-kr: should we also add a note to this CLAUDE.md that we don't need to over-optimize for style? code can be more ugly in this repo as long as it is correct and efficient

ch-kr (commented):
a few initial thoughts, thank you for adding!

CLAUDE.md
```diff
@@ -0,0 +1,307 @@
...
+
+| Subpackage | Purpose |
+|------------|---------|
+| `resources/` | Versioned input/output paths and resource objects. Single source of truth for where data lives in GCS. |
+| `sample_qc/` | Hard filters, sex/genetic-ancestry inference, relatedness, outlier filtering, metadata HT. |
+| `annotations/` | Info HT, coverage, frequencies, Ensembl VEP, in-silico predictors, variant QC features. |
+| `variant_qc/` | Random forest / VQSR training, evaluation, final filter. |
+| `create_release/` | Combine annotations, prepare the VCF release, validity checks. |
+
+Not every version has every subpackage (v5 has no `variant_qc/` yet; v4 also has
```
ch-kr, line 104: claude `/doctor` pointed out that v5 does have `variant_qc` now

CLAUDE.md
```diff
@@ -0,0 +1,307 @@
...
+```bash
+python gnomad_qc/v5/annotations/generate_variant_qc_annotations.py --environment batch --create-info-ht --test
+```
+
+Exploratory work happens in Jupyter notebooks on a Dataproc cluster
+(`hailctl dataproc connect <cluster> nb`). Small checks can run against Hail's
+local Spark backend, streaming data from GCS via the storage connector.
+
+Cost conventions: read from the public gnomAD buckets where possible to avoid
+requester-pays and egress charges, use **us-central1-b**, prefer autoscaling
```
ch-kr, line 190: claude flagged this for an unrelated reason (it couldn't find guidance for using zone b). should we still encourage claude to use us-central1-b given the resource issues we've had in that zone?

CLAUDE.md
```diff
@@ -0,0 +1,307 @@
...
+clusters, and prefer preemptible workers for initial runs.
+
+## Code Style
+
+- black (`preview = true`, py39–py311), isort (`black` profile, `gnomad` is
+  third-party), autopep8. `cuKING` is excluded from all formatting/linting;
+  `v2` is excluded from pydocstyle.
```
ch-kr, line 197: `/doctor` flagged this is also excluded from pylint

CLAUDE.md
```diff
@@ -0,0 +1,307 @@
...
+  referenced or built by the repo.
+- **Never remove code without asking first.**
+- **Never use `repartition()`** — see the Hail notes above.
+- **Never hardcode a `gs://` path** — go through a `get_*()` resource function.
+
+## Repo Gotchas
+
+- `tmp_*.py` and other `tmp_*` files at the repo root are scratch/debugging
+  artifacts, not pipeline code — don't pattern-match off them or import them.
+- `cuKING/` (in versions that have it) is a vendored CUDA relatedness tool,
```
ch-kr, line 297: from claude:
```
cuKING/ (in versions that have it) is a vendored CUDA relatedness tool

  ▎ It's a git submodule (.gitmodules → populationgenomics/cuKING), not vendored — the directory is empty until git submodule update --init. Worth stating, since a fresh clone won't have it and CI checks out submodules recursively.
```

ch-kr (commented):
a few more comments from running `/doctor`

PR #778, opened 2026-07-07
Fixes for trio stats

ch-kr:
PR adds two `naive_coalece` for trio stats HT creation. One coalesce is added per chromosome, and one is added for the full table; ideally, we'd run the coalesce per-chromosome, but I've already generated trio stats per chromosome, so the best path for v5 is to run a coalesce on the union.

PR also adds some checks on the unioned trio stats HT

### DIFF
--- a/gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
+++ b/gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
@@ -232,8 +232,10 @@ def run_generate_trio_stats(
 
 
 # Per-chromosome trio stats inherit the dense trio MT's fine partitioning
-# (~2k-12k partitions each for a few tens of MB), so each is coalesced to this
-# both at generation and as it is read into the union (~22x this, ~550, combined).
+# (~2k-12k partitions each for a few tens of MB), so each is coalesced down to
+# this value -- both at generation and again as it's read into the union.
+# Across the 22 autosomes that's ~22 * 25 = ~550 partitions combined, versus
+# the ~138k we'd get from their native partitionings.
 TRIO_STATS_N_PARTITIONS_PER_CHROM = 25
 
 # adj/raw paired count fields produced by generate_trio_stats_expr.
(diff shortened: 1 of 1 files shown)


gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -539,13 +681,39 @@ def main(args):
                 output_step_resources={"trio_stats_ht": [trio_stats_ht_path]},
                 overwrite=overwrite,
             )
+            # The per-chrom HTs inherit the dense trio MTs' fine partitioning
+            # (~2k-12k partitions each), so naive_coalesce each one as it is read;
+            # union() then concatenates ~22x TRIO_STATS_N_PARTITIONS_PER_CHROM
+            # partitions rather than ~138k.
```
klaricch:
```suggestion
            # union() then concatenate to get len(chroms) * TRIO_STATS_N_PARTITIONS_PER_CHROM partitions instead of the sum of every chrom's native count (~138k).
```

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -231,6 +231,143 @@ def run_generate_trio_stats(
     return generate_trio_stats(mt)
 
 
+# Per-chromosome trio stats inherit the dense trio MT's fine partitioning
+# (~2k-12k partitions each for a few tens of MB), so each is coalesced to this
+# both at generation and as it is read into the union (~22x this, ~550, combined).
```
klaricch:
```suggestion
# Per-chromosome trio stats inherit the dense trio MT's fine partitioning
# (~2k-12k partitions each for a few tens of MB), so each is coalesced down to
# this value -- both at generation and again as it's read into the union.
# Across the 22 autosomes that's ~22 * 25 = ~550 partitions combined, versus
# the ~138k we'd get from their native partitionings.
```

klaricch (changes requested):

klaricch (approved):

PR #781, opened 2026-08-27
Split v5 info HT generation into stratified, checkpointed steps

klaricch:
Replaces the single-pass create-info-ht flow with four independent steps so the expensive AC aggregation can be stratified by chromosome and allele-count range, checkpointed, and never repeated:

- --generate-ac-info-ht: run the AC info aggregation for one stratum (--chrom and/or --min-alleles/--max-alleles, inclusive bounds), with chunk (--explode-partitions), scout (--scout-alleles), or per-contig read restriction. --use-local-allele-agg computes the AC aggregation and AS_pab_max in local-allele space (identical output, much faster at high-allele loci); --validate-local-allele-agg diffs it against the original before writing. Each output records its creation parameters in an ac_info_ht_parameters global.
- --union-ac-info-hts: assemble the strata, with schema, row-count, and key-disjointness assertions. Inputs can be derived as --union-contigs x --union-allele-strata (each table's recorded parameters are verified against the stratum its path claims) or listed explicitly.
- --create-sites-vcf-ht: import and reformat the AoU annotated sites-only VCF once (adds AS_lowqual).
- --create-final-info-ht: join the unioned AC info HT with the VCF HT (optionally shuffle-free via --repartition-for-join) to write the final info HT.

Because the VCF is never interval-filtered, split/min-repped VCF loci can no longer be dropped at scout/chunk interval boundaries (the bug that produced rows with missing info annotations). Chunk intervals are now derived from variant-locus bounds in genomic order, an empty interval derivation writes an empty provenance-stamped stratum instead of silently reading the full VDS, and checkpoint paths are derived from run parameters (durable annotations bucket by default, matching the production-by-default resource convention; 30-day temp bucket with --use-tmp-info-paths, or automatically on --test/--test-n-partitions runs, with --*-override escape hatches).

Scout-interval derivation stays distributed (only the ~2 chunk-boundary loci per interval are collected to the driver, never the full target set), with a default of 20 target loci per re-read partition calibrated on min10-max100 test runs. The AC-info read restriction reuses the existing --chrom argument (originally a separate --contig flag); its chunk/test mutual-exclusion checks are scoped to --generate-ac-info-ht so trio-stats usage of --chrom is unchanged.

Resource changes: get_ac_info_ht_checkpoint_path derives parameterized names, new get_vcf_ht_checkpoint_path, get_aou_vds gains a sample_collisions parameter (collision set collected once per call) and treats read_intervals=[] as a zero-row read, and add_project_prefix_to_sample_collisions accepts a pre-collected set. The precomputed sample-artifact JSONs (collision / high-quality / release sample IDs, read automatically inside get_aou_vds) moved to a separate PR (kl/aou_vds_sample_artifact_jsons); until that lands, VDS loads here use the original sample-table scans.


gnomad_qc/v5/resources/basics.py
```diff
@@ -457,11 +397,22 @@ def get_aou_vds(
...
     _validate_environment(environment, _SAMPLE_DATA_ENVIRONMENTS)
+    if filter_partitions is not None and read_intervals is not None:
+        raise ValueError(
+            "`filter_partitions` and `read_intervals` are mutually exclusive."
+        )
     aou_v8_resource = aou_test_dataset if test else aou_genotypes
-    vds = aou_v8_resource.vds()
+    # NOTE: explicit None check -- an empty interval list must restrict the read
+    # to zero rows, not silently fall back to an unrestricted full-VDS read.
+    read_args = {"intervals": read_intervals} if read_intervals is not None else None
```
ch-kr, line 482: maybe there should be a note here that this empty VDS read is desired for info HT compute? I initially was going to comment that maybe this line should raise an error rather than read the empty VDS before I started reading the variant QC annotation script

gnomad_qc/v5/resources/annotations.py
```diff
@@ -279,6 +274,107 @@ def get_info_ht(
...
+
+    Uses the 30-day temp bucket and a filename derived from the run's parameters
+    (each component included only when set), so per-stratum and per-chunk runs
+    get distinct paths without manual overrides, e.g.
+    ``ac_info_ht_test_3p_min10_max100.ht`` or ``ac_info_ht_chunk0_2000_max9.ht``.
+    This helper only returns the checkpoint path and does not imply that reruns
+    will automatically reuse existing data.
+
+    :param version: Version of annotation path to return.
+    :param test: Whether the path should use the test suffix.
```
ch-kr: nit, maybe the `stable` parameter in this function and `get_vcf_ht_checkpoint_path` below should be renamed `test`, since that's how we usually use that parameter in resource functions? and if `stable` -> `test`, maybe `test` could be `add_test_suffix` or something similar

gnomad_qc/v5/resources/annotations.py
```diff
@@ -279,6 +274,107 @@ def get_info_ht(
...
+    if stable:
+        prefix = (
+            f"{_annotations_root(version, test=False, environment=environment)}"
+            "/create_info_ht"
+        )
+    else:
+        prefix = (
+            f"{qc_temp_prefix(version=version, environment=environment, days=30)}"
+            "create_info_ht"
+        )
```
ch-kr: nit, maybe we should pass the `days` parameter to the `qc_temp_prefix` call within [`_annotations_root`](https://github.com/broadinstitute/gnomad_qc/blob/8c748ee1590ec573ed6855b20cd564c672dc8d7b/gnomad_qc/v5/resources/annotations.py#L53), and then this section could be
```suggestion
    prefix = (
        f"{_annotations_root(version, test=stable, environment=environment)}"
        "/create_info_ht"
    )
```

ch-kr: this would also shorten the function below

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -548,68 +1031,775 @@ def get_tp_ht_for_vcf_export(
...
+    overwrite = args.overwrite
+    test_n_partitions = args.test_n_partitions
+    test = args.test or test_n_partitions is not None
+
+    _validate_args(args, test)
+
+    # Interval derivation is only needed to read the VDS.
+    need_vds = args.generate_ac_info_ht or args.generate_sibling_stats
+    sub_intervals = _derive_read_intervals(args, need_intervals=need_vds)
+
```
ch-kr: maybe these three lines could be moved to right above line 1631 (`vds = None`), since `sub_intervals` doesn't seem to be called until later and these lines would remove the need for an extra comment describing when the VDS isn't needed

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -58,29 +60,108 @@
 logger.setLevel(logging.INFO)
 
 
-def generate_ac_info_ht(vds: hl.vds.VariantDataset) -> hl.Table:
+def _aou_sample_artifact_path(name: str, environment: str, test: bool = False) -> str:
```
ch-kr: maybe we should replace the current resources with these jsons?

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -548,68 +1031,775 @@ def get_tp_ht_for_vcf_export(
...
+            # a few partitions of the full (not test) VDS).
+            test=args.test,
+            environment=environment,
+        )
+
+    try:
+        if args.write_sample_artifacts:
+            logger.info("Writing run-invariant sample artifacts...")
+            # Import here to avoid circular imports.
+            from gnomad_qc.v5.resources.meta import get_sample_id_collisions, meta
```
ch-kr: nit, these could be moved to the top of the script

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -98,6 +179,158 @@ def generate_ac_info_ht(vds: hl.vds.VariantDataset) -> hl.Table:
         for f in ac_filter_groups
     }
 
+
+def _grp_ac_expr_local(mt: hl.MatrixTable, ac_filter_groups: Dict) -> Dict:
```
ch-kr, line 127: I wonder if we should add this (and the LA version of computing AS_pab_max) to gnomad_methods for other groups with large numbers of alt alleles?

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -136,22 +372,212 @@ def generate_ac_info_ht(vds: hl.vds.VariantDataset) -> hl.Table:
...
+    # Checkpoint both sides so the counts, anti-joins, and field diff below don't
+    # each re-run the (expensive) aggregation.
+    orig = _ac_arrays(candidate=False).checkpoint(
+        hl.utils.new_temp_file("validate_orig", "ht")
+    )
+    cand = _ac_arrays(candidate=True).checkpoint(
+        hl.utils.new_temp_file("validate_cand", "ht")
+    )
+
+    # (2) Key-set parity in both directions; (3) empty-input guard.
```
ch-kr: minor, why do the comments start with (2) and (3) and not (1)?

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -136,22 +372,212 @@ def generate_ac_info_ht(vds: hl.vds.VariantDataset) -> hl.Table:
...
+    # (so divergence to NA is caught, not silently skipped) and never yields a
+    # missing predicate that `hl.agg.all` would skip. Unequal lengths => mismatch.
+    def _elem_int(x, y):
+        return hl.if_else(
+            hl.is_missing(x) | hl.is_missing(y),
+            hl.is_missing(x) & hl.is_missing(y),
+            x == y,
+        )
+
+    def _elem_float(x, y):
```
ch-kr: nit, should this be renamed to indicate it's checking only pab_max, or should the pab_atol be renamed something like float_atol?

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -114,9 +347,12 @@ def generate_ac_info_ht(vds: hl.vds.VariantDataset) -> hl.Table:
...
     )
 
-    ac_info_expr = ac_info_expr.annotate(
-        AS_pab_max=pab_max_expr(mt.LGT, mt.LAD, mt.LA, hl.len(mt.alleles))
-    )
+    if use_local_allele_agg:
+        logger.info("Using local-allele-space AS_pab_max...")
+        pab_expr = _local_pab_max_expr(mt)
+    else:
+        pab_expr = pab_max_expr(mt.LGT, mt.LAD, mt.LA, hl.len(mt.alleles))
```
ch-kr: minor, for readability, maybe this could be moved above and merged with the AC expr?
```
if use_local_allele_agg:
    logger.info("Using local-allele-space for AC aggregation and AS_pab_max...")
    pab_expr = _local_pab_max_expr(mt)
    grp_ac_expr = _grp_ac_expr_local(mt, ac_filter_groups)
else:
    pab_expr = pab_max_expr(mt.LGT, mt.LAD, mt.LA, hl.len(mt.alleles))
    grp_ac_expr = _grp_ac_expr_original(mt, ac_filter_groups)
```

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -548,68 +1031,775 @@ def get_tp_ht_for_vcf_export(
...
+    holds many target loci. This keeps the partition count in a healthy range
+    regardless of how many target loci there are.
+
+    Intervals never span contigs.
+
+    :param target_ht: Table of target loci (e.g. from `scout_target_loci`).
+    :param n_partitions: Desired (approximate) number of intervals/partitions. If
+        set, takes precedence over `rows_per_partition`.
+    :param rows_per_partition: Target number of loci per interval/partition. Used
+        only when `n_partitions` is None. Defaults to 2000 if neither is set.
```
ch-kr: how did you decide on a default of 2000?

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -548,68 +1031,775 @@ def get_tp_ht_for_vcf_export(
...
+    """
+    loci = target_ht.key_by("locus").select().distinct()
+    collected = loci.locus.collect()
+    n_loci = len(collected)
+    if n_loci == 0:
+        return []
+
+    if n_partitions is None:
+        rows_per_partition = rows_per_partition or 2000
+        n_partitions = max(1, math.ceil(n_loci / rows_per_partition))
```
ch-kr: is this extra `max` necessary? there's already a `max` below (and a check for n_loci == 0 above)

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -548,68 +1031,775 @@ def get_tp_ht_for_vcf_export(
...
+                hi=hl.agg.max(vd.locus.position),
+            ),
+        )
     )
+    spans = {contig: max(b.hi + 1 - b.lo, 1) for contig, b in bounds.items()}
+    total_span = sum(spans.values())
+    # Iterate contigs in genomic order: the aggregation dict comes back in
+    # lexicographic key order (chr1, chr10, ..., chr2, ...), which read_vds's
+    # partitioner rejects for multi-contig chunks.
+    contig_order = {
```
ch-kr: this gets used another time below in `group_scout_loci_into_intervals`; maybe it could be saved as a constant?

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -548,68 +1031,775 @@ def get_tp_ht_for_vcf_export(
...
+                "--chunk-start/--chunk-stop) are mutually exclusive: use one "
+                "read-restriction mode."
+            )
+        if args.test_n_partitions is not None:
+            raise ValueError(
+                "--contig cannot be combined with --test-n-partitions; use a "
+                "small contig (e.g. chr21) as the test unit instead."
+            )
+
+    if args.scout_alleles and args.min_alleles is None and args.max_alleles is None:
```
ch-kr, line 1573: should there be a check that `scout_alleles` shouldn't be passed with `explode_partitions`? there doesn't seem to be anything blocking someone passing these two together, but `scout_alleles` would be skipped based on the elif logic in `_derive_read_intervals`

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -548,68 +1031,775 @@ def get_tp_ht_for_vcf_export(
...
+                hl.Locus(contig, sub_lo, reference_genome=reference_genome),
+                hl.Locus(contig, sub_hi, reference_genome=reference_genome),
+                includes_start=True,
+                includes_end=last,
+            )
+        )
+    return intervals
+
+
+def _validate_args(args, test: bool) -> None:
```
ch-kr: claude flagged this:
> --explode-partitions with only one of --chunk-start/--chunk-stop reaches range(5, None) at [:1260-1261](vscode-webview://0c924ss7s684cn4nfvvei22cfk4c0qp8d09is4p7o4ct01lk8j7h/gnomad_qc/v5/annotations/generate_variant_qc_annotations.py#L1260-L1261) and dies with a bare TypeError; every other combination gap in _validate_args is explicitly validated. Fix: In _validate_args, raise a ValueError if --explode-partitions is set and either chunk bound is None.

ch-kr: claude also flagged
> --generate-sibling-stats combined with --contig / --explode-partitions / --scout-alleles writes a partial-genome sib-stats table to the canonical production path. need_vds at [generate_variant_qc_annotations.py:1586](vscode-webview://0c924ss7s684cn4nfvvei22cfk4c0qp8d09is4p7o4ct01lk8j7h/gnomad_qc/v5/annotations/generate_variant_qc_annotations.py#L1586) is generate_ac_info_ht or generate_sibling_stats, so _derive_read_intervals restricts the shared VDS regardless of which step asked; get_sib_stats ([annotations.py:87-107](vscode-webview://0c924ss7s684cn4nfvvei22cfk4c0qp8d09is4p7o4ct01lk8j7h/gnomad_qc/v5/resources/annotations.py#L87-L107)) has no contig/chunk parameterization. Downstream, out-of-region loci join to missing and hl.or_else(..., False) silently marks them not-sibling-singleton.

these extra args weren't designed to be run with sib stats, so maybe there should be a check here that --generate-sibling-stats isn't passed with --contig / --explode-partitions / --scout-alleles

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -548,68 +1031,775 @@ def get_tp_ht_for_vcf_export(
...
+        set, takes precedence over `rows_per_partition`.
+    :param rows_per_partition: Target number of loci per interval/partition. Used
+        only when `n_partitions` is None. Defaults to 2000 if neither is set.
+    :param reference_genome: Reference genome used to order contigs so the returned
+        intervals are in genomic order (required by `read_vds`'s partitioner).
+    :return: List of `hl.Interval` objects, in genomic order, each covering a
+        contiguous chunk of target loci within a single contig.
+    """
+    loci = target_ht.key_by("locus").select().distinct()
+    collected = loci.locus.collect()
```
ch-kr: claude flagged that this collect contradicts a previous docstring:

> Scout mode collects every target locus to the driver, contradicting the "stays distributed (no driver collect)" docstring at [:1112](vscode-webview://0c924ss7s684cn4nfvvei22cfk4c0qp8d09is4p7o4ct01lk8j7h/gnomad_qc/v5/annotations/generate_variant_qc_annotations.py#L1112) — group_scout_loci_into_intervals does loci.locus.collect() at [:1167](vscode-webview://0c924ss7s684cn4nfvvei22cfk4c0qp8d09is4p7o4ct01lk8j7h/gnomad_qc/v5/annotations/generate_variant_qc_annotations.py#L1167). Nothing requires --contig, so a genome-wide --scout-alleles --max-alleles 9 (a documented stratum) collects hundreds of millions of loci onto the driver. Fix: Derive boundaries in Hail (hl.scan.count() on the checkpointed target HT, filter to idx % rows_per_partition == 0, collect only boundaries), or reject --scout-alleles without a --min-alleles lower bound.

I don't think you need hl.scan.count() since add_index should work just fine, but I stopped prompting for more clarification on this point since I'm running out of tokens 😅

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
ch-kr, line 1: minor, github flagged that there's a merge conflict in this script (just with the imports)

Image

ch-kr (changes requested):
a few questions and comments, many of which are minor. one additional question: can you share your plan / example commands for the full AC info HT run? I think you mentioned you're planning to run per contig splitting on number of alt alleles but want to confirm the flow in the script

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
klaricch: addressed in https://github.com/broadinstitute/gnomad_qc/pull/782

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
klaricch: changing to 20, which is what im currently using for a test on 2000 partitions with 10-100 alleles, and which runs faster than 2000

klaricch:
https://github.com/broadinstitute/gnomad_qc/pull/781#pullrequestreview-5045129364


here are the example commands for a 2k partition test (pre changes in round 2 of this PR):
# Generate AC info hts

hailctl batch submit \
--image "$IMAGE" \
--memory highmem \
-v /Users/kristen/code/gnomad_qc/gnomad_qc:/pkg/gnomad_qc \
--env PYTHONPATH=/pkg \
-- \
python3 -m gnomad_qc.v5.annotations.generate_variant_qc_annotations \
--environment batch \
--app-name test_2000p_ac_info_max9 \
--generate-ac-info-ht \
--explode-partitions \
--chunk-start 0 \
--chunk-stop 2000 \
--read-subintervals-scale 15 \
--max-alleles 9 \
--worker-memory highmem \
--driver-memory highmem \
--driver-cores 8 \
--worker-cores 2


hailctl batch submit \
--image "$IMAGE" \
--memory highmem \
-v /Users/kristen/code/gnomad_qc/gnomad_qc:/pkg/gnomad_qc \
--env PYTHONPATH=/pkg \
-- \
python3 -m gnomad_qc.v5.annotations.generate_variant_qc_annotations \
--environment batch \
--app-name test_2000p_ac_info_min10_max100 \
--generate-ac-info-ht \
--test-n-partitions 2000 \
--scout-alleles \
--scout-rows-per-partition 20 \
--min-alleles 10 \
--max-alleles 100 \
--use-local-allele-agg \
--worker-memory highmem \
--driver-memory highmem \
--driver-cores 8 \
--worker-cores 2


Note: would use --chrom <chrom_name> in place of --test-n-partitions 2000 or --explode-partitions --chunk-start 0 --chunk-stop 2000 for the real run


# Union  AC info hts

hailctl batch submit \
--image "$IMAGE" \
-v /Users/kristen/code/gnomad_qc/gnomad_qc:/pkg/gnomad_qc \
--env PYTHONPATH=/pkg \
-- \
python3 -m gnomad_qc.v5.annotations.generate_variant_qc_annotations \
--environment batch \
--app-name test_2000p_ac_info_union \
--union-ac-info-hts \
--test \
--union-input-ac-info-ht-paths \
gs://fc-11093c2b-590e-424a-91ac-0cc040d562fc/batch-tmp-30day/gnomad.genomes.v5.0.qc_data/create_info_ht/ac_info_ht_chunk0_2000_max9.ht \
gs://fc-11093c2b-590e-424a-91ac-0cc040d562fc/batch-tmp-30day/gnomad.genomes.v5.0.qc_data/create_info_ht/ac_info_ht_test_2000p_min10_max100.ht \
--ac-info-ht-checkpoint-path-override gs://fc-11093c2b-590e-424a-91ac-0cc040d562fc/batch-tmp-30day/gnomad.genomes.v5.0.qc_data/create_info_ht/ac_info_ht_test_2000p_union.ht \
--worker-memory highmem \
--driver-memory highmem \
--driver-cores 8 \
--worker-cores 2

# Create sites vcf ht

hailctl batch submit \
--image "$IMAGE" \
--memory highmem \
-v /Users/kristen/code/gnomad_qc/gnomad_qc:/pkg/gnomad_qc \
--env PYTHONPATH=/pkg \
-- \
python3 -m gnomad_qc.v5.annotations.generate_variant_qc_annotations \
--environment batch \
--app-name sites_vcf_ht \
--create-sites-vcf-ht \
--vcf-ht-checkpoint-path gs://fc-11093c2b-590e-424a-91ac-0cc040d562fc/batch-tmp-30day/gnomad.genomes.v5.0.qc_data/create_info_ht/sites_vcf.ht \
--driver-memory highmem \
--driver-cores 4 \
--worker-cores 2

# Create final info ht

hailctl batch submit \
--image "$IMAGE" \
-v /Users/kristen/code/gnomad_qc/gnomad_qc:/pkg/gnomad_qc \
--env PYTHONPATH=/pkg \
-- \
python3 -m gnomad_qc.v5.annotations.generate_variant_qc_annotations \
--environment batch \
--app-name test_2000p_final_info_ht \
--create-final-info-ht \
--test \
--ac-info-ht-checkpoint-path-override gs://fc-11093c2b-590e-424a-91ac-0cc040d562fc/batch-tmp-30day/gnomad.genomes.v5.0.qc_data/create_info_ht/ac_info_ht_test_2000p_union.ht \
--vcf-ht-checkpoint-path-override gs://fc-11093c2b-590e-424a-91ac-0cc040d562fc/batch-tmp-30day/gnomad.genomes.v5.0.qc_data/create_info_ht/sites_vcf.ht \
--info-ht-path-override gs://fc-11093c2b-590e-424a-91ac-0cc040d562fc/batch-tmp-30day/gnomad.genomes.v5.0.qc_data/create_info_ht/info_ht_test_2000p_final.ht \
--driver-memory highmem \
--driver-cores 4 \
--worker-cores 2

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
klaricch: i could add it, although it only came up as an issue with workers being preempted in batch

ch-kr: maybe not worth it right now then 🤔

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -548,6 +1101,604 @@ def get_tp_ht_for_vcf_export(
...
+    environment = args.environment
+
+    scout_input_partitions = args.test_n_partitions
+    vds = get_aou_vds(
+        high_quality_only=True,
+        filter_partitions=(
+            range(scout_input_partitions) if scout_input_partitions else None
+        ),
+        annotate_meta=False,
+        test=args.test,
```
ch-kr, line 1340: claude keeps flagging in review that the VDS reading behavior is inconsistent between here and `compute_chunks`. if I read this function correctly, this function shouldn't need the `test=args.test` line, so maybe we could remove to avoid this false positive reporting? or maybe add a comment that this was useful for testing and is deliberately different behavior

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -548,6 +1101,604 @@ def get_tp_ht_for_vcf_export(
...
+                hl.Locus(contig, sub_lo, reference_genome=reference_genome),
+                hl.Locus(contig, sub_hi, reference_genome=reference_genome),
+                includes_start=True,
+                includes_end=last,
+            )
+        )
+    return intervals
+
+
+def _validate_args(args, test: bool) -> None:
```
ch-kr: `test` doesn't seem to be used anymore

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -548,6 +1101,604 @@ def get_tp_ht_for_vcf_export(
...
+
+    if args.export_true_positive_vcfs and not (
+        args.transmitted_singletons or args.sibling_singletons
+    ):
+        raise ValueError(
+            "--export-true-positive-vcfs requires at least one of"
+            " --transmitted-singletons or --sibling-singletons"
+        )
+
+    if args.union_ac_info_hts and args.generate_ac_info_ht:
```
ch-kr, line 1586: minor, claude flagged:
> X2 [Agreed: Opus agree, Sonnet agree] _validate_args blocks only --union-ac-info-hts + --generate-ac-info-ht; nothing rejects --generate-ac-info-ht + --create-final-info-ht with --min-alleles/--max-alleles, so one invocation writes a single stratum to the _union path and joins it straight into the production info HT. (generate_variant_qc_annotations.py:1560-1566)
Fix: In _validate_args, raise when args.create_final_info_ht and args.generate_ac_info_ht and (args.min_alleles is not None or args.max_alleles is not None), mirroring the existing mutual-exclusion check.

gnomad_qc/v5/annotations/generate_variant_qc_annotations.py
```diff
@@ -548,6 +1101,604 @@ def get_tp_ht_for_vcf_export(
...
+
+    Rather than one interval (and therefore one partition) per target locus, the
+    sorted target loci are grouped into chunks so that each resulting interval -- and
+    thus each partition produced when the VDS is re-read with these intervals --
+    holds many target loci. This keeps the partition count in a healthy range
+    regardless of how many target loci there are.
+
+    The derivation stays distributed: only the chunk-boundary loci (~2 per
+    interval) are collected to the driver, never the full target set, so a
+    genome-wide scout with hundreds of millions of target loci is safe.
```
ch-kr: minor, claude flagged
> X4 [Agreed: Opus agree, Sonnet agree] The docstring at 1220-1222 claims a genome-wide scout is safe because only ~2 boundary loci per interval reach the driver, but rows_per_partition or 20 (line 1253) makes the interval count ceil(n_loci / 20), so the collect() at 1276 scales as n_loci / 10 — ~30M rows for 300M loci. ([generate_variant_qc_annotations.py:1220-1222,1253,1276](vscode-webview://0c924ss7s684cn4nfvvei22cfk4c0qp8d09is4p7o4ct01lk8j7h/gnomad_qc/v5/annotations/generate_variant_qc_annotations.py#L1220-L1222)) Fix: Correct the docstring to state driver collect scales as 2 * n_loci / rows_per_partition and that the 20-row default suits only small high-allele strata; optionally clamp the derived n_partitions so the default can't blow up the driver.

ch-kr (commented):
just a few more minor things

ch-kr (approved):
LGTM!

PR #782, opened 2026-09-01
Kl/aou vds sample artifact jsons

klaricch:
Loading the AoU VDS rescans the ~330-partition sample metadata tables on every call; this PR snapshots those sample sets once as small permanent JSONs.

meta.py: aou_sample_artifact_path (permanent bucket for full scope, 4-day temp for test), load_aou_sample_artifact_json, and write_aou_vds_sample_jsons (sample_id_collisions.json, high_quality_samples.json, release_samples.json; reruns need --overwrite).
basics.py: get_aou_vds prefers the JSONs for collision prefixing and the release/high-quality filters, falling back to the original scans when absent. 
create_sample_qc_metadata_ht.py: standalone --write-vds-sample-artifact-jsons step (plus --test) — lives with the meta HT builder; rerun whenever the meta HT is rewritten.
compute_coverage.py: get_group_membership_ht returns the ~365k-row lookup coalesced to 10 partitions, fixing the ~330-task fan-out at the source.

### DIFF
--- a/gnomad_qc/v5/resources/meta.py
+++ b/gnomad_qc/v5/resources/meta.py
@@ -1,20 +1,32 @@
 """Script containing metadata related resources."""
 
+import json
+import logging
+from typing import Optional
+
+import hail as hl
+import hailtop.fs as hfs
 from gnomad.resources.resource_utils import (
     ExpressionResource,
     TableResource,
     VersionedTableResource,
 )
+from gnomad.utils.file_utils import file_exists
 
 from gnomad_qc.v5.resources.basics import (
     _SAMPLE_DATA_ENVIRONMENTS,
+    _check_resource_existence,
     _get_base_bucket,
     _validate_environment,
+    qc_temp_prefix,
 )
 from gnomad_qc.v5.resources.constants import (
     CURRENT_PROJECT_META_VERSION,
     CURRENT_SAMPLE_QC_VERSION,
 )
+
+logger = logging.getLogger("meta_resources")
+logger.setLevel(logging.INFO)
 
 
 def get_project_meta(environment: str = "batch") -> VersionedTableResource:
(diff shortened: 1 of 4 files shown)


gnomad_qc/v5/resources/meta.py
```diff
@@ -56,6 +68,144 @@ def get_sample_id_collisions(environment: str = "batch") -> TableResource:
...
+) -> str:
+    """
+    Return the permanent path to a run-invariant AoU sample-level artifact.
+
+    Artifacts are small precomputed files read at VDS load (or by pipeline
+    workers) instead of rescanning the ~330-partition sample metadata tables:
+    the VDS-loading JSONs written by `write_aou_vds_sample_jsons`
+    (``sample_id_collisions.json``, ``high_quality_samples.json``,
+    ``release_samples.json``) and pipeline-specific Hail Tables such as the
+    frequency pipeline's ``meta_small.ht`` and ``gm_coalesced[_reduce].ht``.
```
mike-w-wilson:
```suggestion
    ``release_samples.json``).
```
I would drop the pipeline specific references for now since these are not using this function at the moment. We can add to the docstring if we do.

gnomad_qc/v5/sample_qc/create_sample_qc_metadata_ht.py
```diff
@@ -591,6 +605,27 @@ def get_script_argument_parser() -> argparse.ArgumentParser:
...
+        action="store_true",
+    )
+    parser.add_argument(
+        "--test",
+        help=(
+            "With --write-vds-sample-artifact-jsons, write the test-scoped "
+            "artifact paths (used when loading the 10-sample test VDS)."
+        ),
+        action="store_true",
+    )
```
mike-w-wilson: --test is only used inside --write-vds-sample-artifact-jsons  line 501. On a normal meta-HT run it parses fine but is silently ignored, and lines 579–581 still checkpoint to the production meta() path. Since --test means "write to test paths" in other scripts, someone running create_sample_qc_metadata_ht.py --test --overwrite out of habit would overwrite the production meta HT instead of a test one. Lets require the artifact flag with test or just drop it since they are cheap to run and overwrite

mike-w-wilson (changes requested):
Two small things!

mike-w-wilson (approved):
lgtm

PR #783, opened 2026-09-02
Kl/aou vds sample artifact jsons

klaricch:
remove remaining "test" references in artifacts


klaricch:
addressed in https://github.com/broadinstitute/gnomad_qc/pull/784

PR #785, opened 2026-09-08
Update adj for AoU data to use GQ20

mike-w-wilson:
After reviewing the AoU bam DP and GQ, we determined there was no reason to up the adj GQ threshold to 30 for AoU. This uses GQ>20 for ref calls (All will pass) and uses gnomAD's threshold for variant data.

### DIFF
--- a/gnomad_qc/v5/annotations/annotation_utils.py
+++ b/gnomad_qc/v5/annotations/annotation_utils.py
@@ -4,6 +4,7 @@ import logging
 from typing import Union
 
 import hail as hl
+from gnomad.utils.annotations import get_adj_expr as get_gnomad_adj_expr
 
 logging.basicConfig(format="%(levelname)s (%(name)s %(lineno)s): %(message)s")
 logger = logging.getLogger("annotation_utils")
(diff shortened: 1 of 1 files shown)


gnomad_qc/v5/annotations/annotation_utils.py
```diff
@@ -38,21 +50,27 @@ def annotate_adj_no_dp(
         gt_expr = mt.GT
         ad_expr = mt.AD
```
ch-kr: as you mentioned in the comment above, the ref blocks in the AoU VDS don't have LGT or LAD, so I think we should add an extra clause to this if statement

```
----------------------------------------
Global fields:
    'ref_block_max_length': int32
----------------------------------------
Column fields:
    's': str
----------------------------------------
Row fields:
    'locus': locus<GRCh38>
----------------------------------------
Entry fields:
    'GQ': int32
    'END': int32
    'GT': call
    'LEN': int32
----------------------------------------
Column key: ['s']
Row key: ['locus']
----------------------------------------
```

ch-kr (commented):
thanks for adding! one comment

klaricch (commented):
other than the one change mentioned by kc lgtm

gnomad_qc/v5/annotations/annotation_utils.py
```diff
@@ -15,8 +14,10 @@
 # Adapted from https://github.com/broadinstitute/gatk/pull/8772/files.
 #
 # The AoU VDS has no DP entry field, so DP is approximated as sum(LAD). Hom-ref calls
-# (reference blocks, which carry GQ but no LAD) are filtered on GQ only. All other
-# calls use the usual gnomAD cutoffs: GQ, DP (as sum(LAD)), and AB for het calls.
+# with no allele depths (reference blocks, which carry GQ but no LAD) are filtered on
+# GQ only. All other calls, including hom-ref calls that do carry allele depths (e.g.
```
ch-kr, line 19: are there other homref calls that have AD aside from the ones that are downcoded after splitting multiallelics?

ch-kr (approved):
LGTM

klaricch (approved):

gnomad_qc/v5/annotations/annotation_utils.py
```diff
@@ -12,47 +12,100 @@
...
+    return mt.annotate_entries(adj=adj_expr)
 
 
 def get_adj_expr(
     gt_expr: hl.expr.CallExpression,
-    gq_expr: Union[hl.expr.Int32Expression, hl.expr.Int64Expression],
+    gq_expr: hl.expr.Int32Expression | hl.expr.Int64Expression,
     ad_expr: hl.expr.ArrayNumericExpression,
-    adj_gq: int = 30,
+    adj_gq: int = 20,
```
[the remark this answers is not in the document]
mike-w-wilson: This is a solid find but the existing frequency script is dated and I am not concerned about this as I have a new branch that will go in once AN is in

gnomad_qc/v5/annotations/annotation_utils.py
mike-w-wilson: No, that phrasing is misleading here in AoU context but if anyone used this for some reason and their hom refs had AD, they would be routed through the normal cutoff

